PR #173, opened 2015-04-14
SimplifyingMinisat: return value from bool function

jirislaby:
compiler complains: control reaches end of non-void function


msoos:
Hey! Thanks for noticing this, it's pretty bad. I fixed it by returning `s->okay()` which is the more correct value to return :) Thanks again!

PR #175, opened 2015-04-16
few fixes for bulding easily in OBS

jirislaby:


msoos:
I'm sorry to say again, but you are a magician. I need to learn all these tricks and put them my other repos. Really good stuff, thank you! I just read up on you on linux.com, congrats! And SuSE sounds great too, it was the first Linux distro I ever used (and I paid for it, handbook came along, was fun!). Thanks again for all and send a note if you need help!

Mate

PR #176, opened 2015-04-17
fix CMAKE_INSTALL_LIBDIR bug I introduced

jirislaby:
Commit ffa79942584bc61919f93a97c2394809c9291ff9 (CMakeLists: support
different lib dirs) added possibility to override the directory where
libraries are installed. But when using the PATH type, the path is
being evaluated absolute to the current working directory. The
libraries then end up in a weird local directory suffixed by the lib
string.

In fact we want CMAKE_INSTALL_LIBDIR to be a string like lib or lib64,
so switch PATH to STRING.


msoos:
Thanks!

PR #185, opened 2015-05-19
Attempt to hide all symbols apart from those we want to be public in

delcypher:
the STP shared library. This unfortunately completely breaks compilation
of stp (and several other executables) because they use STP's internal
API. This is gross!

We need to come up with a way of fixing this.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -221,6 +221,18 @@ include(CheckCxxHashMap)
 check_cxx_hashmap()
 
 # -----------------------------------------------------------------------------
+# Setup symbol visibility flags
+# -----------------------------------------------------------------------------
+include(GenerateExportHeader)
+# FIXME: YUCK! I know this is deprecated but *_VISIBILITY_PRESET and VISIBILITY_INLINES_HIDDEN
+# properties don't work on shared libraries built from OBJECT libraries.
+add_compiler_export_flags(CXX_SYMBOL_VISIBILITY_FLAGS)
+set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} ${CXX_SYMBOL_VISIBILITY_FLAGS}")
+# Note -fvisibility-inlines-hidden is not a valid option when compiling C
+string(REPLACE -fvisibility-inlines-hidden "" C_SYMBOL_VISIBILITY_FLAGS ${CXX_SYMBOL_VISIBILITY_FLAGS})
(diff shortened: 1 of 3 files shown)


delcypher:
@msoos Do not commit this! I did this so you could try it out and we could have a discussion.

I tried hiding STP's symbols (you can check using ``nm -C -D lib/libstp.so``). This works but it breaks compilation of the ``stp`` executable entirely because it tries to use symbols I've now hidden. If there wasn't enough evidence already of poor design this is a **strong reminder** that STP's internals are a mess. A well designed front end to a library should be able to work only using the public API.

Right now I see a few short term solutions

* We keep making more and more symbols public until the linking succeeds. This is gross. I'd advise against doing.
* We move the ``main_common`` code into the STP library itself so that it can access all of STP's internal symbols. This means we'd have to expose a few more symbols but probably far less than the above approach
* We build the ``stp`` executable against the static library always. This would avoid the issue because symbols are never hidden in a static library.

I'm also a little unsure about exposing the ``cpp_interface``. I'm not sure if it's used outside the library. If not we should not expose its symbols.

Long term what really needs to happen is we need to create a brand new C (or C++) API and have the ``stp`` executable use that.

There is a minor issue with all this which is hiding the symbols prevents us from writing unit tests against STP's internal which is something we should be doing. I think the only solution for that would be to always build the unit tests against the static library or disable symbol hiding when running tests.

msoos:
> A well designed front end to a library should be able to work only using the public API.

Indeed. We need a new API. I suggest we create a v2 version fo the C library API, make the parser separate only using the library API, and make the main use the parser API (which could be a superset of the library API). I don't mind doing this in a hack-until-it-works way rather then trying to design things best and then after lengthy discussion realizing that some other call would be needed. Once something reasonable is done, we can rearrange cuts (i.e. functions exposed, unexposed, etc) and fix up function parameters, calling conventions. I don't want it to be hectic, but I don't want to design an API that in the end cannot be made to work, either. Would this make sense? What do you think?

delcypher:
Doing this makes sense but I'm hesitant in doing this before we make a release.

I think we should use one of the temporary solutions I suggested (or don't hide any symbols) so we can make a release and after releasing we can start tackling the API issue. What would you rather do?

msoos:
We have been working a lot on hiding symbols. It mostly works and so this pull request is stale. I'm closing because we are much further ahead and other pull requests may be useful but this is too old now, in a good way :)

PR #194, opened 2015-08-21
fix python3 version comparison

prokls:
The current implementation fails eg. for::

    >>> sys.version_info
    sys.version_info(major=21, minor=0, micro=1, releaselevel='final',
    serial=0)
    >>> '21.0.1' > '3'
    False
    >>> sys.version_info >= (3, 0, 0)
    True

I admit it takes some time to get to python21 ;)


msoos:
Thanks!

Issue #195, opened 2015-08-21
python binding: (a << 1) != (1 << a)


prokls:
```python
import operator
import stp

def shift(op, param):
    s = stp.Solver()
    a = s.bitvec('a', 32)
    b = s.bitvec('b', 32)
    c = s.bitvec('c', 32)
    s.add(a == 3)
    s.add(b == op(a, param))
    s.add(c == op(param, a))
    assert s.check()
    assert s.model()['b'] == (3 << param if op is operator.lshift else 3 >> param)
    # works, but shouldn't
    assert s.model()['c'] == (3 << param if op is operator.lshift else 3 >> param)
    # does not work, but should
    #assert s.model()['c'] == (param << 3 if op is operator.lshift else param >> 3)

for i in range(0, 30):
    shift(operator.lshift, i)
    shift(operator.rshift, i)
```

I found a bug with the shifting operator. I assume the MWE is self-explanatory.

prokls:
```python
import stp

def lshift(param):
    s = stp.Solver()
    a = s.bitvec('a', 32)
    b = s.bitvec('b', 32)
    c = s.bitvec('c', 32)
    s.add(a == 3)
    s.add(b == a << param)
    s.add(c == param << a)
    assert s.check()
    assert s.model()['b'] == (3 << param)
    # works, but shouldn't
    assert s.model()['c'] == (3 << param)
    # does not work, but should
    #assert s.model()['c'] == (param << 3)

def rshift(param):
    s = stp.Solver()
    a = s.bitvec('a', 32)
    b = s.bitvec('b', 32)
    c = s.bitvec('c', 32)
    s.add(a == 3)
    s.add(b == a >> param)
    s.add(c == param >> a)
    assert s.check()
    assert s.model()['b'] == (3 >> param)
    # works, but shouldn't
    assert s.model()['c'] == (3 >> param)
    # does not work, but should
    #assert s.model()['c'] == (param >> 3)

for i in range(1, 30):
    lshift(i)
    rshift(i)
```

Let's get rid of the ``operator`` library. I'm not sure ``a << b`` equals ``operator.lshift(a, b)`` in every situation.

prokls:
The bug is obviously the [python datamodel](https://docs.python.org/dev/reference/datamodel.html?highlight=rlshift#object.__rlshift__) implemented wrongfully. See [bindings/python/stp/stp.py#L466](https://github.com/stp/stp/blob/master/bindings/python/stp/stp.py#L466):

```python
[...]

PR #196, opened 2015-08-27
Fixing python as per issue by  Lukas Prokop

msoos:
Related to issue #195


msoos:
Thanks Lukas! Some of your fixes went into #204 so the work wasn't lost :) Thanks again for everyone!

Issue #88, opened 2014-03-18
The "lit" tool should use its own Python interpreter


smcc:
The wiki page about the new testing infrastructure recommends that if you don't have the permissions to install the "lit" tool as part of your system-wide Python installation, you can use the virtualenv tool. However in my testing this didn't actually work, because the Makefile tried to run lit using its own favorite Python interpreter (e.g., /usr/bin/python), whereas virtualenv requires the script to be run with a special wrapped Python interpreter which is mentioned in the script's #! line. But it did work after I removed the explicit Python reference from the makefile. For reference my change is at:

https://github.com/bitblaze-fuzzball/stp/commit/2594114f7b53f1665da37a6b25a5fb862266019a

msoos:
Let's see if TravisCI likes it. If it does, I merge and we are good. In that case, it works in both cases so there is no need to have the python hard-coded and this issue can be closed :)

msoos:
Okay, so the cmake now has an option to specify the python interpreter explicitly (run `cmake-gui ...` from `build`) Also, the virtualenv should work out of the box. I am closing this. Please re-open if there are further issues with building!

PR #197, opened 2015-08-28
Making LIT use normal python as per Stephen McCamant's suggestion

msoos:
Diff as per issue #88

### DIFF
--- a/tests/query-files/CMakeLists.txt
+++ b/tests/query-files/CMakeLists.txt
@@ -10,14 +10,12 @@ configure_file(lit.site.cfg.in lit.site.cfg @ONLY)
 if (Boost_FOUND)
     add_custom_target(query-file-tests
                   DEPENDS stp
-                  COMMAND ${PYTHON_EXECUTABLE}
-                          ${LIT_TOOL} ${LIT_ARGS} .
+                  COMMAND ${LIT_TOOL} ${LIT_ARGS} .
     )
 else()
     add_custom_target(query-file-tests
                   DEPENDS stp_simple
-                  COMMAND ${PYTHON_EXECUTABLE}
-                          ${LIT_TOOL} ${LIT_ARGS} .
+                  COMMAND ${LIT_TOOL} ${LIT_ARGS} .
     )
 endif()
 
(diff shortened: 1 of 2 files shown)


msoos:
@smcc Can you please comment on whether you still need this? Maybe I can fix something around the underlying issue such that it doesn't break the build.

smcc:
I don't know that my relationship with this change has ever been as strong as "need". I do spend a larger than usual amount of time trying to get STP to run compiling entirely from source without installing anything in /usr, because of the centralized systems administration system our department has, but I'm aware that most other users are not in that situation any more.

My thinking when I made this change was that it would make the build process strictly more flexible, and it was motivated by a particular build environment that seemed to me like it should work. On the other hand Dan's opinion was that the configuration I was trying was not supposed to work and would not be worth supporting. I haven't been won over by his argument but lots of trade-offs like this have to be made in supporting a build process so my opinions probably shouldn't stand in the way of progress.

I believe that even within my limitations of not installing things in /usr, there are different configurations of the other tools that would allow me to build a vanilla STP correctly without this change. And since for now I'm already using a branched (and currently somewhat out of date) version of STP with FuzzBALL, no trunk changes will have any immediate effects on FuzzBALL users.

Understanding why this change breaks the Travis build might lead to an insight and a different way of making the change that would make everyone happy; it seems to clearly be some kind of counterexample to my earlier thought that the change would not break any builds. However in a minute or so of looking I couldn't understand what the problem was from the Travis reports that are linked to this pull request. Is it obvious to you which of the lines in the log represents the build failure?

msoos:
I suspect the build failure to be something entirely different. As for flexibility of the build system -- you are probably right on that. The python to be used should probably be optionally passable to the `cmake`. I don't have the environment to check that though. If you can create a diff for that to cmake it would be great.

As for drift of versions, I'm sorry to hear. I should work more on STP so it would make more sense to keep it updated inside projects. There have been some diffs and fixes over the months, but nothing major :( I have a lot of commitments, and I have been improving CryptoMiniSat in significant ways...

delcypher:
> The python to be used should probably be optionally passable to the cmake

Changing the value of (hidden by default) of ``PYTHON_EXECUTABLE`` in the cache will let you do this.

msoos:
Oh wow, let me make that explicit then. The commit https://github.com/stp/stp/commit/3785148da15919de445e476a6f20b06c881cf50c now fixes this issue. Thanks for everyone for their input!

PR #198, opened 2015-09-03
AST: simplify FatalError and mark as noreturn

jirislaby:
There is duplicated code in FatalError implementations. Remove that
and since now FatalError is de facto "noreturn", mark it as such to
silence the compiler yelling.

### DIFF
--- a/include/stp/AST/AST.h
+++ b/include/stp/AST/AST.h
@@ -37,8 +37,9 @@ namespace stp
 // TODO remove -- only used in c_interface.cpp NOT from main.cpp
 void process_argument(const char ch, STPMgr* bm);
 
-void FatalError(const char* str, const ASTNode& a, int w = 0);
-void FatalError(const char* str);
+void FatalError(const char* str, const ASTNode& a, int w = 0)
+                __attribute__((noreturn));
+void FatalError(const char* str) __attribute__((noreturn));
 void SortByExprNum(ASTVec& c);
 void SortByArith(ASTVec& c);
 bool exprless(const ASTNode n1, const ASTNode n2);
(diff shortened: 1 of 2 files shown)


include/stp/AST/AST.h
```diff
@@ -37,8 +37,9 @@ namespace stp
 // TODO remove -- only used in c_interface.cpp NOT from main.cpp
 void process_argument(const char ch, STPMgr* bm);
 
-void FatalError(const char* str, const ASTNode& a, int w = 0);
-void FatalError(const char* str);
+void FatalError(const char* str, const ASTNode& a, int w = 0)
+	__attribute__((noreturn));
```
delcypher: Huh is that clang-format doing the weird indentation on that attribute?

lib/AST/ASTmisc.cpp
```diff
@@ -267,14 +263,9 @@ void FatalError(const char* str)
...
-    assert(false);
-    exit(-1);
-  }
-  else
-  {
-    assert(false);
-    exit(-1);
   }
+  assert(false);
+  exit(-1);
```
delcypher: I think we should change this to ``abort()``. If debugging with gdb, gdb will catch an ``abort()`` but not a call to ``exit()``.

include/stp/AST/AST.h
jirislaby: This is vim. Where do you prefer the attribute to be indented?

jirislaby:
So I fixed all that, hopefully.

msoos:
I think we had an issue with noreturn before (Windows?). But I think it's the right approach. I'm merging and see if anybody complains :) We can detect GCC/Clang worst case and #ifdef -it. Thanks!

PR #204, opened 2015-11-17
Fix operator overloading in Python bindings

aclements:
This fixes various issues with operator overloading in the Python bindings.

Part of this is equivalent to PR #196, but doesn't have the confusing merges and multiple changes. This also adds a thorough test of overloaded operators to the Python API test suite.


msoos:
Again, great work. Thanks!

PR #205, opened 2015-11-17
Fix vc_getCounterExample for arbitrary expressions

### LINKED ISSUE #203: vc_getCounterExample does not work for all expressions
Based on the documentation of vc_getCounterExample, I would expect to be able to pass an arbitrary expression to it and get the concrete value of that expression in the current counterexample. However, it often simply returns the constant 0. For example,

```python
import stp
from stp.stp import _lib as _stplib

s = stp.Solver()
x = s.bitvec("x", 32)
s.add(x > 1610545248)
y = (x >> 13) + 1
s.check()
xval = _stplib.vc_getCounterExample(s.vc, x.expr)
yval = _stplib.vc_getCounterExample(s.vc, y.expr)
print _stplib.getBVUnsignedLongLong(xval), _stplib.getBVUnsignedLongLong(yval)
```

prints "4294967295 0", but given the value of x, it should have printed 524288 for y.

Given the implementation of AbsRefine_CounterExample::GetCounterExample, this isn't surprising. If the expression isn't already a constant, it looks it up in a map and returns constant 0 if it isn't found. Hence, there are a finite number of non-constant expressions for which GetCounterExample can possibly return a non-0 expression.

I've found that calling vc_simplify on the expression before passing it to vc_getCounterExample seems to work, but I don't know if I'm just getting lucky and my expressions happen to simplify to things that are in the CounterExampleMap, or if there's a fundamental reason this works. If there is the "right" thing to be doing, it could be better documented.

However, I've also found that the following trivial change to  AbsRefine_CounterExample::GetCounterExample works. It seems much more robust to me (and it passes the test suite), but I don't fully understand the CounterExampleMap or the implications of doing this. If this is a sensible change, I'm happy to make a PR (or you can just apply the diff).

```diff
diff --git a/lib/AbsRefineCounterExample/CounterExample.cpp b/lib/AbsRefineCounterExample/CounterExample.cpp
index f446e97..ce2536b 100644
--- a/lib/AbsRefineCounterExample/CounterExample.cpp
+++ b/lib/AbsRefineCounterExample/CounterExample.cpp
@@ -649,7 +649,7 @@ ASTNode AbsRefine_CounterExample::GetCounterExample(const ASTNode& expr)
   if ((it = CounterExampleMap.find(expr)) != CounterExampleMap.end())
     output = TermToConstTermUsingModel(CounterExampleMap[expr], false);
   else
-    output = bm->CreateZeroConst(expr.GetValueWidth());
+    output = TermToConstTermUsingModel(expr, false);
   return output;
 } 
```

aclements:
Currently vc_getCounterExample only works for constant expressions,
boolean expressions, and for bitvector expressions that happen to be
in the counterexample map. If none of these cases hold, it simply (and
confusingly) returns a constant 0 expression. However, the prototype
of vc_getCounterExample suggests that it can compute an arbitrary
expression's value in the counterexample and the documentation does
nothing to dissuade this. Furthermore, a caller generally doesn't know
what's in the counterexample map, which limits it to usefully querying
only constant and variable expressions and nothing more complex.

All of the mechanism to get the counterexample value of an arbitrary
expression already exists. In fact, by substantially *simplifying*
AbsRefine_CounterExample::GetCounterExample to just call
ComputeFormulaUsingModel or TermToConstTermUsingModel according to the
expression type, we can make it far more general. Most of the logic
currently in GetCounterExample either duplicates logic in
TermToConstTermUsingModel or needlessly restricts it, so by simply
ripping it out, we make GetCounterExample work on arbitrary
expressions.

We also add a test that checks the computed values of various
expressions in the counterexample. Without the change to
GetCounterExample, this test fails.

Fixes #203.


msoos:
Wow, thanks! Great work :) You seem to be quite knowledgeable. The project needs some love (as you probably have noticed -- though we have been doing some cleanup). If you feel like hacking more, go ahead, I'd love to merge in fixes and cleanups etc. And by the way, the gtest unit test is absolutely fantastic, we need more of that stuff everywhere :)

Thanks again for the work!

Mate

PS: I need to fix the coveralls integration, seems broken, reports failure.

PR #228, opened 2016-07-17
Fix static linking

MartinNowack:
Fixes the static linking with cryptominisat5.

The changes allow the libraries be installed in arbitrary locations.

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -10,7 +10,7 @@ generators.
 
 - ``CMAKE_BUILD_TYPE`` - The build type (e.g. Release)
 - ``CMAKE_INSTALL_PREFIX`` - The prefix for install (e.g. /usr/local )
-- ``DBUILD_STATIC_BIN``- Build static binary and library
+- ``BUILD_STATIC_BIN``- Build static binary and library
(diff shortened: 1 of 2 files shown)


msoos:
Thanks! First one seem good, but I need to check! The second one is wrong, we need minisat from Niklas' github, not our own :) I'll look through this soon, tonight is already a bit late :S In the coming days definitely!

MartinNowack:
@msoos Ok, I rolled back all the changes, except the location fix.
The major issue I had was that I used the following arguments:
```
cmake -DCMAKE_INSTALL_PREFIX=/tmp/ubuntu/local -DBUILD_SHARED_LIBS=OFF -DENABLE_PYTHON_INTERFACE:BOOL=OFF  ../
```

If `BUILD_SHARED_LIBS=OFF`  but `BUILD_STATIC_BIN=ON` is *not* provided, cmake tries to link the binaries with the dynamic libs, which of course fails.
Maybe it would be useful to add an additional check here.
Also, `BUILD_STATIC_BIN=ON` does not imply `BUILD_SHARED_LIBS=OFF`, makes sense to me now, but it just hit me ;)

The reason I changed the URL was because INSTALL.md  still refers to the STP version.

msoos:
Thanks a lot, looks good!

PR #231, opened 2016-08-05
Use internal unique ID for ASTNodes for ordering

MartinNowack:
Hashing based on memory addresses changes the
order of elements being iterated over depending on the assigned addresses.

This patch introduces a unique ID generated during creation
of the node. And use this for comparison and
hashing functions.

This is one step towards solving #218, it doesn't solve the whole problem yet.
There are a couple of more cases in the code.

### DIFF
--- a/include/stp/AST/ASTInternal.h
+++ b/include/stp/AST/ASTInternal.h
@@ -104,6 +104,9 @@ protected:
    *******************************************************************/
   unsigned int _value_width;
 
+  uint64_t node_uid;
+  static uint64_t node_uid_cntr;
+
   /****************************************************************
    * Protected Member Functions                                   *
    ****************************************************************/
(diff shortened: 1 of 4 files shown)


msoos:
Wow you are quite amazing. Thanks! Sorry, I feel like I've been keeping you
back rather than helping :S I promise to do more.

delcypher:
@MartinNowack Awesome. Out of curiosity what approach did you use to find this?

msoos:
Wow. Great work, thank you! There should be some tools to detect stuff like this. Hackish code. Works, but usually more frustrating to work with than the speed/memory advantage it gives. Thanks so much!

MartinNowack:
@msoos Thanks a lot for merging!
@delcypher It's similar to the problem we have in KLEE.
If we want to preserve a specific order when we iterate over collections (e.g. to be deterministic), the data structure should not depend on pointers or it should have the appropriate comparator and hash function.
So, I was looking at the code for exactly that pattern, that was the first one I spot ;)

PR #237, opened 2016-10-19
simplifier: use abort() instead of assert(false)

jirislaby:
assert becomes nop when built with -DNDEBUG. So use abort in
getConstantBit instead of assert.


msoos:
Actually, can you please do:

```
assert(false);
abort();
```

Instead? We catch bugs by looking for assert fail. So I'd rather have both. It doesn't make things any slower, compilers see that there is an abort in case of debug due to the `assert()` and the `abort()` gets opimised out in that case. Otherwise, `abort()` is in.

By the way, we highly discourage `-DNDEBUG` builds :)

PR #246, opened 2017-04-17
MSVC Compatibility

fkutzner:
I've made some small changes to STP in order to build it with MSVC on Win64. The largest one is related to the Kind enum clashing with the definitions of TRUE and FALSE via `Windows.h`. I wasn't sure whether renaming constants in the enum Kind would break third-party code, so I restructured the code in `windows/` so that `Windows.h` does not get included in STP header files.

Hope you like the commits. This PR touches several parts of the code, so if you'd like to have several smaller PRs instead of this one, let me know.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -218,6 +218,8 @@ if(WIN32)
       set(STP_DEFS_COMM ${STP_DEFS_COMM} -D_CRT_SECURE_NO_WARNINGS)
       set(STP_INCL_COMM windows/winports windows/winports/msc99hdr ${STP_INCL_COMM})
       include_directories(${STP_INCL_COMM})
+      add_subdirectory(windows)
+      set(PLATFORM_COMPAT_LIBRARIES ${PLATFORM_COMPAT_LIBRARIES} windows_compat)
 
       # stack size of MSVC must be specified
       string(REGEX REPLACE "/STACK:[0-9]+" "" CMAKE_EXE_LINKER_FLAGS ${CMAKE_EXE_LINKER_FLAGS})
--- a/include/stp/AST/AST.h
+++ b/include/stp/AST/AST.h
@@ -27,13 +27,12 @@ THE SOFTWARE.
 
 #include "UsefulDefs.h"
 #include "ASTNode.h"
-
+#include "stp/Util/Attributes.h"
 
 namespace stp
 {
-void FatalError(const char* str, const ASTNode& a, int w = 0)
-                __attribute__((noreturn));
-void FatalError(const char* str) __attribute__((noreturn));
+ATTR_NORETURN void FatalError(const char* str, const ASTNode& a, int w = 0);
+ATTR_NORETURN void FatalError(const char* str);
 void SortByExprNum(ASTVec& c);
 void SortByArith(ASTVec& c);
 bool exprless(const ASTNode n1, const ASTNode n2);
--- a/include/stp/AST/ASTBVConst.h
+++ b/include/stp/AST/ASTBVConst.h
@@ -30,7 +30,6 @@ THE SOFTWARE.
 namespace stp
 {
 class STPMgr;
-void FatalError(const char* str);
 
 /******************************************************************
  *  Class to represent internals of a bitvector constant          *
--- a/include/stp/Simplifier/constantBitP/FixedBits.h
+++ b/include/stp/Simplifier/constantBitP/FixedBits.h
@@ -25,6 +25,8 @@ THE SOFTWARE.
 #ifndef FIXEDBITS_H_
 #define FIXEDBITS_H_
 
+#include <stp/Util/Attributes.h>
+
 #include <vector>
 #include <iostream>
 #include <cassert>
@@ -35,7 +37,7 @@ namespace stp
 {
 class ASTNode;
 typedef unsigned int* CBV;
-void FatalError(const char* str);
+ATTR_NORETURN void FatalError(const char* str);
 }
 
 namespace simplifier
--- a/include/stp/Simplifier/constantBitP/FixedBits.h
+++ b/include/stp/Simplifier/constantBitP/FixedBits.h
@@ -30,6 +30,7 @@ THE SOFTWARE.
 #include <vector>
 #include <iostream>
 #include <cassert>
+#include <algorithm>
 
 class MTRand;
 
--- a/include/stp/Util/Attributes.h
+++ b/include/stp/Util/Attributes.h
@@ -1,0 +1,36 @@
+/********************************************************************
+ * AUTHOR: Felix Kutzner
+ *
+ * BEGIN DATE: Apr, 2017
+ *
+Permission is hereby granted, free of charge, to any person obtaining a copy
+of this software and associated documentation files (the "Software"), to deal
+in the Software without restriction, including without limitation the rights
+to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+copies of the Software, and to permit persons to whom the Software is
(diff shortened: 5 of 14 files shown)


msoos:
Wow! That's some serious love you gave the codebase! Thanks! What SAT solver do you use with it? CryptoMiniSat is supposed to compile on Windows nowadays, I'm curious if you are using it?

fkutzner:
Thanks :) I've been using MiniSat with STP, but I'll check out CryptoMiniSat.

msoos:
It'd be great if you tried, actually. CryptoMiniSat now compiles under MSVC reliably (there is an automated check). You can check the README.md of CryptoMiniSat, it tells how to compile it step-by-step.

Actually... how do you compile STP under MSVC? Can you please give a short set of commands you use? I'd like to put it up on the README.md of STP. That would be super-duper helpful. Thanks a ton in advance!

fkutzner:
Building it isn't entirely untethered from POSIX yet:

    cmake -G "Visual Studio 14 2015 Win64" -DCMAKE_INSTALL_PREFIX=<InstallPath> -DCMAKE_PREFIX_PATH=<PathToMinisatAndZlibInstallation>;<PathOfCygwinRootDir> -DMINISAT_LIBRARY=<pathOfMinisat.libFile> <PathOfSTPSource>
    cmake --build . --config Debug
    cmake --build . --config Debug --target install

I'm using Cygwin's flex and bison, but run the build process from the Git bash (a gcc-less MinGW environment which also doesn't contain MinGW C/C++ header files). CMake generates a Visual Studio project and uses MSBuild to build STP, so it is possible to use the Visual Studio IDE - it's no fun, though, since MSVC emits a tremendous amount of warnings, slowing down the UI. Boost is missing from my compilation mainly because I haven't gotten around to add it yet. If you'd like, I could write the readme entry (including build instructions for zlib and boost), probably by next week.

I'm currently working on a few things:

- reducing the amount of warnings. Most of them don't seem to be very helpful, e.g. the thousands of lines about struct paddings and removing unused C++ library inline functions. These make building STP with MSVC take ages
- getting the automated tests to run. I'm halfway there, but there are a lot of small issues like getting the tests to use the non-Cygwin Python installation and needing to conjure up a "not" program (I did some manual testing for the PR, though, and ran the automated tests on macOS to make sure STP still works with other toolchains.)
- finding minisat in the CMAKE_PREFIX_PATH
- enabling shared library builds

The last bit is tricky with MSVC, as we would need to explicitly mark the symbols we whish to be exported. Adding them for the C++ interface without sprinkling export tags all over the code seems pretty cumbersome. I'm looking into ways to make a Unix-style shared library with all symbols exported. What do you think? It would be minimally intrusive for the STP code, but complicate the build process a bit since the symbols would need to be read from the object files and gathered in a definition file. Should I create an issue for this as soon as I know this is viable?

Building the static version of CryptoMiniSat like above was a breeze! I haven't installed m4ri yet, though, which prevented me from building STP with CryptoMiniSat. I'll look into what went wrong next weekend. The dynamic (default) build didn't work, since a static one is forced for MSVC and no DLL is generated against which the executable can be linked. MSVC only generated a warning about clean_to_int in solverconf.h:53. not always returning a value. However, I couldn't build it in the IDE, since xxd.exe could not be found. It worked fine from the Git bash, though.

msoos:
Wow, that's some really good work, thanks! I'll get it to compile under AppVeyor and then I'll see the warnings too and I can fix some/most as well. I did that for CryptoMiniSat and it seems like I did OK. The symbol export thing is a valid concern. We should actively try to fix it for STP. I fixed it for CryptoMiniSat and it wasn't that bad at all. Essentially the same thing (with some #define's) can be used to do both POSIX and WIN-style symbol export fixing.

As for CryptoMiniSat:
* I need to fix the dynamic building. Sorry, I was concentrating on releasing an executable.
* xxde.exe works when vim is installed and cmake is used to build. I guess/hope that worked as you could build it (only statically, though, but still)

So, thanks again! I'll get to this a bit later this week!

msoos:
Actually, @delcypher had a go at the symbol-hiding here: https://github.com/stp/stp/pull/185

fkutzner:
Sounds great, I'd really appreciate it! I've thought a bit about symbol visibility - there are two separate issues: on the one hand, there's the question of which symbols should be exposed, and on the other hand, there's the issue of exposing symbols consistently (as far as possible) across platforms. I'm currently using `CMAKE_WINDOWS_EXPORT_ALL_SYMBOLS` (https://cmake.org/cmake/help/v3.4/prop_tgt/WINDOWS_EXPORT_ALL_SYMBOLS.html, https://blog.kitware.com/create-dlls-on-windows-without-declspec-using-new-cmake-export-all-feature/) to build a Windows DLL mimicking the Unix shared library. However, this solution has a few drawbacks:
- if a client uses a global variable defined within the DLL, the corresponding declaration still needs a `__declspec(dllimport)` attribute.
- CMake 3.4 is required, and due to a bug in CMake, symbols which are defined in object libraries cannot be automatically added to the .def file (i.e. exported) by CMake versions lesser than 3.7. Thus, building the DLL would require CMake 3.7.
- AFAIK, MSVC can optimize calls into DLLs when `__declspec(dllimport)` attributes are set. Those attributes would be missing with this approach.

Currently, I've solved the first issue by simply adding `__declspec(dllimport)` (via an STPAPI #define) to global variable declarations, but defining accessor functions for the global variables would be nicer, probably eliminating the strict need for `__declspec(dllimport)` altogether.

As soon as the visibility issue is resolved, we could use proper `__declspec(dllimport/dllexport)` attributes and remove `CMAKE_WINDOWS_EXPORT_ALL_SYMBOLS`. IMO, this is not too hackish - would this be an acceptable short-term approach for shared library building with MSVC?

msoos:
Actually, I can't even compile MiniSat right now: https://ci.appveyor.com/project/msoos/stp It gives me two kinds of errors:

```
[00:04:21] C:\projects\stp\minisat\minisat\utils\System.cc(94): error C2244: 'memUsedPeak': unable to match function definition to an existing declaration [C:\projects\stp\minisat\build\minisat-lib-shared.vcxproj]
```

and

```
[00:04:21] C:\projects\stp\minisat\minisat\core\Solver.cc(996): error C3688: invalid literal suffix 'PRIu64'; literal operator or literal operator template 'operator ""PRIu64' not found [C:\projects\stp\minisat\build\minisat-lib-shared.vcxproj]
```

I can fix these, but I'm just curious, how did you compile MiniSat? I thought MiniSat is _really_ meant to compile out-of-the-box on Windows, no? Thanks in advance!

fkutzner:
Ouch, sorry, I forgot about that. https://github.com/niklasso/minisat seems unmaintained. I use minisat from https://github.com/stp/minisat, where the first issue is already fixed. For the second one, I just added whitespaces around PRIu64 in Solver::printStats().

fkutzner:
I've looked through the AppVeyor file - does AppVeyor bring perl, flex and bison? I use these tools from my Cygwin installation. Also, a Python installation (outside of Cygwin) is needed. I've got the automated tests up and running (see https://github.com/fkutzner/stp/tree/win64dll), and needed to adjust the CMake parameters by passing -DPYTHON_EXECUTABLE, -DPYTHON_LIB_INSTALL_DIR (on Windows, the Python installation directory) and -DLIT_TOOL (path to lit-script.py, not the lit binary). Furthermore, to run the tests I needed to install an implementation of _not_ in a %PATH% directory.

Having a Windows box at hand, I could try to have a stab at setting up AppVeyor for STP, however it would take some time since I don't have experience with AppVeyor yet.

Thanks for putting time into this! :)

PR #249, opened 2017-04-26
documented the entire old C API interface header with nice doxygen comments

Robbepop:
This commit should improve the lifes of people that use the old STP C API.

No other changes were made, besides renaming a few function arguments which has no effect and won't break any code.

This was overdue!

A few parts could be further improved, but not by me since I do not know enough of STP to document them.

### DIFF
--- a/include/stp/c_interface.h
+++ b/include/stp/c_interface.h
@@ -48,24 +48,45 @@ typedef void* Type;
 typedef void* WholeCounterExample;
 #endif
 
-
+//! \brief Processes the given flag represented as char for the given validity checker.
+//! 
+//! The following flags are supported:
+//!  - 'a': Disables optimization. TODO: What kind of optimization is meant here?
+//!  - 'c': Enables construction of counter examples.
+//!  - 'd': Enables construction and checking of counter examples. Superseeds flag 'c'.
+//!  - 'm': Use SMTLib1 parser. Conflicts with using SMTLib2 parser.
+//!  - 'n': Enables printing of the output. TODO: What is meant with output here?
+//!  - 'p': Enables printing of counter examples.
+//!  - 'q': Enables printing of array values in declared order.
+//!  - 'r': Enables accermannisation.
+//!  - 's': Sets the status flag to true. TODO: What consequenses does this have?
+//!  - 't': Enables quick statistics. TODO: What is this?
+//!  - 'v': Enables printing of nodes.
+//!  - 'w': Enables word-level solving. TODO: What is mean with this?
+//!  - 'y': Enables printing binaries. TODO: What is meant with this?
+//! 
+//! This function panics if given an unsupported or unknown flag.
 void process_argument(const char ch, VC bm);
 
-
-// o  : optimizations
-// c  : check counterexample
-// p  : print counterexample
-// h  : help
-// s  : stats
-// v  : print nodes
-
-// The "num_absrefine" argument isn't used at all. It's left for compatibility
-// with existing code.
-// TODO remove these two functions, it's an ugly way of controlling settings
+//! \brief Deprecated: use process_argument instead!
+//! 
+//! Sets flags for the validity checker.
+//! For more information about this look into the documentation of process_argument.
+//! 
+//! Parameter num_absrefine has no effect in the current implementation.
+//! It is left for compatibility with existing code.
+//! 
 void vc_setFlags(VC vc, char c, int num_absrefine _CVCL_DEFAULT_ARG(0));
+
+//! \brief Deprecated: use process_argument instead!
+//! 
+//! Sets flags for the validity checker.
+//! For more information about this look into the documentation of process_argument.
+//! 
 void vc_setFlag(VC vc, char c);
 
 //! Interface-only flags.
+//! 
 enum ifaceflag_t
 {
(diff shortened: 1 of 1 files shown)


msoos:
WOW, that's some crazy good work, thank you! You are more than welcome to make changes, if you have ideas about cleanup or more docs, etc, we'd be really happy to see them go in! Just hack around and make a merge request :) Thanks again, this is really great work!

Robbepop:
Thank you for the warm words!

I actually did the old C-API clean-up because I was studying it while trying to design a possible new C-API v2.

You can examine the current state of the C-API v2 in the following links:
 - HEADER: [LINK](https://github.com/Robbepop/stp/blob/c-interface-v2/include/stp/c_interface_v2.h)
 - README: [LINK](https://github.com/Robbepop/stp/blob/c-interface-v2/include/stp/README.md)

However, I am not sure if I am the right person to design it, nor do I know if one is still needed or what all the important requirements for a new STP C-API are.

Comments from your side are appreciated.

---

Also I forgot to mention that I annotated some comments in the old C-API with a few `TODO`s that are mainly questions that should be answered and documented by a person with more knowledge about STP.

msoos:
Wow, I'll need more time to read through this, but it looks like you have tried to fix a lot of the issues with the API. The API is in all honesty pretty crappy, and a good C API would mean we can put some really nice wrappers around it, e.g. for Python, C++, etc. which might help a lot. Also, it might improve our sanity. Maybe @TrevorHansen  could take a peek? Trevor, what do you think?

Robbepop:
I too think that a decently designed C-API would bring huge benefits to the development of STP.

There are still many things in the design of the v2 C-API that I do not like and many things are still leaning heavily on the old C-API on a design perspective.
This could be changed and in fact the v2 C-API is still under heavy development and experimentation.

My problems are:
 - I am not an expert with SMT or SAT solvers.
 - I do not know STP internals.
 - I have no idea what the specific requirements for such a C-API for STP are.

So I just hacked along and tried my best during my free-time.

---

The README covers the most important parts of what is exactly broken with the current C API (imo) and I also lost a few very important sentences about the design guidelines of the new C-API that are fundamental to it.

msoos:
No worries. Nobody is perfect at everything regarding STP, it's a complicated beast and few, if anybody, know every bit of it. I suggest you add some test cases for the new API, see if you can do basic stuff with it and check if the results are what you'd expect. Currently we have some tests for the v1 API, maybe you can take a peek and see if you can adapt them to the v2? That might clear some things up about the usability and may give you  ideas or just confidence about the new API :)

Robbepop:
Yes, a test infrastructure will be of great help once the implementation is done.

However, we are still at the early phases with it and need to decide on some fundamental designs.
For example, the serialization API i.e. `stp_serializeExpr` etc. is currently quite complex and could be simplified.
Besides that there are open questions, such as:
 - Do we need a logging API?
     - A very simple logging API is already declared.
 - What is the exact story for error handling?
     - Currently we have `stp_Result` but I may want to extend it further to make it possible
        to put more information into such a `stp_Result` object.
 - Strong typing or weak typing? I think it is better to go for one instead of trying to do both.
     - As far as I can tell the old C-API tried both but was stuck to weak typing. But far from everybody is happy with that.
 - How is ownership of STP objects handled?
     - I would like that STP takes ownership of as many entities as possible and to make APIs like `stp_deleteExpr` unnecesary.

msoos:
* Error handling is currently pretty bad. I have changed a lot of things to just exit the hell out. Throwing an exception would probably be the best, but memory cleanup is not always trivial. But that would be the best. The current (non-working) callback mechanism is unworkable.
* Ownership should be mostly STP except for major things. One of the biggest problems with the API is that nobody knows who owns what and what should be freed by whom.

The rest need to be answered by others I think :)

Robbepop:
I think for the new C-API the best error handling method would be one where the user can choose how errors are handled by STP.

For this I currently designed the new API in a way that most API functions have an `stp_Result* outResult` parameter.

If the user wishes to receive an `stp_Result` indicating what error happend (if any) he receive it through this out parameter and if not, the user may parameterize this argument with `NULL` which tells STP that it can panic on errors - and of course best with some kind of logging to let the user still know what happened.

TrevorHansen:
Thanks for all the work on the interface Robbepop. It's been in the too hard-basket for us all for a long time. It's great to see progress.

I've not worked on interfaces seriously before. So I don't have any insights (other than the current one is terrible). If I was to start work on it, i'd begin by copying the good stuff from other's interfaces. The ones that come to mind are:
https://github.com/ahorn/smt-kit
https://github.com/agra-uni-bremen/metaSMT
http://fmv.jku.at/boolector/doc/cboolector.html

Managing ownership seems the hardest problem to me. Internally STP has reference counting which works well within C++, but in C, it seems much harder.

Robbepop:
Thank you very much for the links to the other SMT solver interfaces. That may help a lot!
About reference counting: So far I have leaned the new API towards the API structure of OpenCL which I find pleasant to work with.

---

There they got reference counting for all of their objects via calls like `clRetainContext` and `clReleaseContext` that increases or decreases the reference count of an object instead of having functions like `clDeleteContext`.
I pretty much liked this interface to reference counting as it enables C++ based interfaces to build upon those quite easily in an object oriented way.

Issue #247, opened 2017-04-19
Fix AppVeyor


msoos:
Apparently it's not that difficult.

Pull request discussion here: https://github.com/stp/stp/pull/246

Copy from thread:

---------------
@fkutzner wrote: 

Building it isn't entirely untethered from POSIX yet:

cmake -G "Visual Studio 14 2015 Win64" -DCMAKE_INSTALL_PREFIX=<InstallPath> -DCMAKE_PREFIX_PATH=<PathToMinisatAndZlibInstallation>;<PathOfCygwinRootDir> -DMINISAT_LIBRARY=<pathOfMinisat.libFile> <PathOfSTPSource>
cmake --build . --config Debug
cmake --build . --config Debug --target install
I'm using Cygwin's flex and bison, but run the build process from the Git bash (a gcc-less MinGW environment which also doesn't contain MinGW C/C++ header files). CMake generates a Visual Studio project and uses MSBuild to build STP, so it is possible to use the Visual Studio IDE - it's no fun, though, since MSVC emits a tremendous amount of warnings, slowing down the UI. Boost is missing from my compilation mainly because I haven't gotten around to add it yet. If you'd like, I could write the readme entry (including build instructions for zlib and boost), probably by next week.

I'm currently working on a few things:

reducing the amount of warnings. Most of them don't seem to be very helpful, e.g. the thousands of lines about struct paddings and removing unused C++ library inline functions. These make building STP with MSVC take ages
getting the automated tests to run. I'm halfway there, but there are a lot of small issues like getting the tests to use the non-Cygwin Python installation and needing to conjure up a "not" program (I did some manual testing for the PR, though, and ran the automated tests on macOS to make sure STP still works with other toolchains.)
finding minisat in the CMAKE_PREFIX_PATH
enabling shared library builds
The last bit is tricky with MSVC, as we would need to explicitly mark the symbols we whish to be exported. Adding them for the C++ interface without sprinkling export tags all over the code seems pretty cumbersome. I'm looking into ways to make a Unix-style shared library with all symbols exported. What do you think? It would be minimally intrusive for the STP code, but complicate the build process a bit since the symbols would need to be read from the object files and gathered in a definition file. Should I create an issue for this as soon as I know this is viable?

[...]

PR #252, opened 2017-05-01
DLL building on Windows and C interface symbol visibility

fkutzner:
This PR enables DLL building on Windows (see the discussion in #247).

To get the tests running, some adjustments in the CMake files were unavoidable - namely, generating files depending on which configuration (Debug, Release, ...) is used when invoking MSBuild, since output directory of binaries depends on the configuration. This is done with generator expressions, for which CMake 3.0.2 is needed.

I originally made most of the CMake changes in https://github.com/fkutzner/stp/tree/win64dll (where all symbols get exported from DLLs) and ported them into this branch. In win64dll, the tests ran fine, however with this branch, the query-file-tests seem to hang. On the other branch, I needed to increase the EXE stack size limit for the query-file-tests.

I've also deactivated boost in AppVeyor for now. stp_simple can be built, but stp cannot, since two different versions of the debugging database stp.pdb are generated - one for the DLL, another one for the EXE, causing the linker to refuse linking the executable.

To make sure I didn't inadvertently break anything, I've compiled and tested STP on macOS. It runs fine :)

Hope you like the PR :)


msoos:
Wow, amazing work, thanks! I'll let the builds finish. There have been a number of issues with builds lately, so it will fail, but I want to see how it fails :) Your fixes _really_ bring Windows right back, I mean tests working is completely crazy amazing. Well done! Fingers crossed :)

The boost thingy is an issue, I love boost and it should work. We'll fix that one later. It really is pretty close to working, and I managed to make it work more than once.

fkutzner:
Thanks :) Boost and tests are on my todo list. Boost should be easy - on my machine, it was really just the linker refusing to write an stp.pdb file (for the exe) when taking a an stp.pdb in the same directory (the dll .pdb) as an argument. The hanging tests may be some weird DLL problem. But I'm optimistic, since the tests passed with the all-symbol-exporting DLL.

fkutzner:
There's another thing about running the tests, which I forgot to mention: the _not_ program needs to be in %PATH%. I've installed one in my Windows directory, but it should be pretty easy to add a small _not_ implementation to STP.

msoos:
Oh, weird. Where do we call it? By the way, I just fixed all the tests so they should compile statically too, at least on Linux. I'm a bit confused about Windows and DLLs and Windows testing. I'm looking forward to you working on it when you have time :) It's going quite well now, we have moved to an upgraded Linux, fixed a bunch of issues related to dynamic and static linking, etc. It's a whole lot better now. Thanks for the lots and lots of contributions! :)

fkutzner:
IIRC the lit tool calls _not_. I'll try to get the tests to run in AppVeyor later this week and then tackle the other issues.

msoos:
Yeah, sorry, I kinda broke python for the moment on Windows. I'm trying to fix it. It's slightly complicated... The `file(GENERATE OUTPUT ${CMAKE_CURRENT_BINARY_DIR}/$<CONFIG>/stp.py` broke the Linux build, but I see how the `$<CONFIG>` is needed in Windows. Without the $<CONFIG> it works in Linux, but that breaks Windows, and vice-versa. So I'm trying to fix it back :) I'll let you know when it's fixed.

fkutzner:
No problem, thanks for fixing it :) I probably won't be able to commit much in the next few days, but that's not related to the current AppVeyor issues. They're curious, I cherry-picked the python-binding commit from https://github.com/fkutzner/stp/commits/win64dll and that branch passed Travis. I only had to configure Travis to use a more recent CMake version (>= 3.0.2), but that was the old setup anyway. Building it on macOS also didn't cause any trouble. I'm using CMake 3.8.0 on Windows and 3.7.2 on macOS.

PR #254, opened 2017-05-03
fixed #253: DLL_IMPORT in c_interfaces leaks API and breaks installat…

### LINKED ISSUE #253: c_interface.h installation broken with DLL_PUBLIC
Since `DLL_PUBLIC` was introduced in `c_interface.h` it now has got a new include to `<stp/Util/Attributes.h>` and thus the `c_interface.h` is no longer self-contained.  

This should be fixed in one of the following ways:

- Add the `<stp/Util/Attributes.h>` to the public interface - I think this should **not** be done as we should avoid increasing the size of STP's API. Also we should prevent STP's internal macros from leaking into the official API!
- Make `c_interface.h` self-contained again by defining the `DLL_PUBLIC` macro within it - ugly due to code duplication but better than extending the API
- Split `<stp/Util/Attributes.h>` into two files: one for internal macros (e.g. `NO_RETURN`) and attributes and one API attributes file. These should be included as new installation files in `cmake` and to prevent macros defined in them from leaking into the API (e.g. we should not leak `DLL_PUBLIC` and `DLL_LOCAL` into the API) we need to `#undef` them both at the end of our interface header files. This would be my preferred fix. Question: Where to store the API file for attributes? I'd place them in `stp/Utils` but move them to the same folder as `c_interface.h` and `cpp_interface.h` upon installing.  This may be problematic since include paths would be different.

Or maybe there is another trick?

I have filed a PR that fixes this problem temporarly.

Robbepop:
Fixed c_interface DLL_PUBLIC macro usage from breaking STP's installation and from leaking internal macro definitions to the API.

### DIFF
--- a/include/stp/c_interface.h
+++ b/include/stp/c_interface.h
@@ -36,7 +36,41 @@ extern "C" {
 #endif
 
 #include <stdio.h>
-#include <stp/Util/Attributes.h>
+
+/////////////////////////////////////////////////////////////////////////////
+/// STP API INTERNAL MACROS FOR LINKING
+/// 
+/// These are undefined at the end of this file to prevent them from leaking
+/// into code that includes it.
+/////////////////////////////////////////////////////////////////////////////
+
+#if defined(_MSC_VER)
+    // NOTE: for now, we need STP_SHARED_LIB for clients of the statically linked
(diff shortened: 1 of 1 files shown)


msoos:
I never got to merge this. Should I?

msoos:
Actually, I'm an idiot, I merged this already :) Thanks!

PR #256, opened 2017-05-09
Replaced node typedef in c_interface implementation

Robbepop:
From code:
// FIXME: These typedefs are stupid. They make the code hard to understand!
// These typedefs lower the effort of using the keyboard to type (too
// many overloaded meanings of the word type)

Done.


msoos:
Thanks a lot, looks much cleaner! Cheers!

PR #257, opened 2017-05-09
Enabled tests on AppVeyor

fkutzner:
This PR enables testing in AppVeyor. Curiously, the tests pass on my local Windows machine and fail in AppVeyor (running in Release mode in both cases - see the comments in #247).

Also, building against static boost libraries is enabled now and a tiny part of LLVM is built to get a _not_ utility on Windows for running the tests. To keep build times down, the boost build coming with AppVeyor is used and the LLVM files are cached between AppVeyor runs.

GitHub does not display CI information for me right now. The logs for the most recent commit are:
https://travis-ci.org/fkutzner/stp/builds/230315036
https://ci.appveyor.com/project/fkutzner/stp/build/1.0.49

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -71,9 +71,11 @@ build_script:
   - mkdir myinstall
   - cd build
   - cmake -G %CMAKE_GENERATOR% -DCMAKE_INSTALL_PREFIX=%ZLIB_ROOT% ..
-  - if %PLATFORM%==x86 call msbuild %MSBUILD_FLAGS% /t:Build /p:Configuration=%CONFIGURATION% /p:Platform="x86" zlib.sln
-  - if %PLATFORM%==x64 call msbuild %MSBUILD_FLAGS% /t:Build /p:Configuration=%CONFIGURATION% /p:Platform="x64" zlib.sln
-  - msbuild %MSBUILD_FLAGS% INSTALL.vcxproj
+  - #if %PLATFORM%==x86 call msbuild %MSBUILD_FLAGS% /t:Build /p:Configuration=%CONFIGURATION% /p:Platform="x86" zlib.sln
+  - #if %PLATFORM%==x64 call msbuild %MSBUILD_FLAGS% /t:Build /p:Configuration=%CONFIGURATION% /p:Platform="x64" zlib.sln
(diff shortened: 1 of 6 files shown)


delcypher:
@fkutzner You don't need to build `not` from LLVM. You could just use the `not` tool that is part of [OutputCheck](https://github.com/stp/OutputCheck/tree/master/bin)

msoos:
Wow, good work! I think what's happening has to do with gtest etc. I have fixed all of this for CryptoMiniSat, so I'll probably do that for STP later. I am busy this week, 15th of May is the deadline for the SAT competition, but later I can do it :)

fkutzner:
@delcypher: Thanks for the suggestion, I had overlooked that script. Unfortunately, the Windows loader does not support the shebang notation. Can you recommend a way of compiling it? `pyinstaller --onefile [...]/bin/not` aborted with an exception.

I'm also a bit puzzled by OutputCheck's `not`: STP returns 0 for successful runs and nonzero values for failures. The `not` script returns 1 when the called program returns a negative value - is this intended?

delcypher:
@fkutzner Yes shebang won't work on Windows. You'd have to invoke it like `python not`. One thing I could do is make the `pip install OutputCheck` also install the `not` utility but under a different name. I'd rather not do that though.

> I'm also a bit puzzled by OutputCheck's not: STP returns 0 for successful runs and nonzero values for failures. The not script returns 1 when the called program returns a negative value - is this intended?

Returning 1 when the program's exit code is negative is intentional. Note that [we will only get a negative exit code on UNIX like systems when the program is killed by a signal](https://docs.python.org/2/library/subprocess.html#subprocess.Popen.returncode). If STP crashes (i.e. gets sent SIGSEGV) I still want this to be considered a failure because crashing is never good.

delcypher:
> I haven't tested it on AppVeyor yet, but if it works out, would it be okay to include it in the OutputCheck repository?

I'd rather not do that. We either just need to teach `OutputCheck`'s distutil setup to also install the `not` binary or we should just have a copy of it inside STP's source tree.

We should also fix the way `not` is used in the tests. We should make it a substitution (i.e. `%not`) and set that in the `lit` configuration to be whatever we want. On windows for example we could make that `python \full\path\to\not`.

PR #261, opened 2017-06-19
cmake: fix ENABLE_PYTHON_INTERFACE option

mchalupa:
This option was ignored as it was set always when working
python was found. Let the user forbid it.


msoos:
Thanks!

PR #262, opened 2017-06-26
Some improvements in CMakeLists.txt

csinz:
- Enable RPATH on MacOS X: the STP dynamic library can thus be found more reliably by applications using it
- Do not search for Cryptominisat, if NOCRYPTOMINISAT is set to ON
- Added configuration option to disable Python interface (ENABLE_PYTHON, default ON)


csinz:
I removed the changes about the Python interface from my pull request, so no clashes should occur now.

msoos:
Thanks, merged! Sorry, I clicked on the wrong button at first :) Thanks again, great work!

PR #265, opened 2017-08-06
[CMake] Fix #192

### LINKED ISSUE #192: running `make` without `make AST` first fails
build fails because AST/ASTKind is not present; building the AST target first works though.

I'm using cmake 3.0.2 on osx.

delcypher:
The build steps for the `parser` step were missing a dependency
on the `ASTKind_header` target which would trigger a compilation
failure on macOS when using the "UNIX Makefiles" generator.

This was discovered recently in https://github.com/klee/klee/issues/714
.

I first tried a different approach to fixing this by generating
`ASTKind.cpp` and `ASTKind.h` at configure time and having
CMAKE_CONFIGURE_DEPENDS directory property depend on the files used for
generation. Although this worked it caused STP to be rebuilt from
scratch everytime a re-configure happened.  Although it would be
possible to work around this by only replacing the files if they've
changed at configure time this seems too much work for very little
return. So I went for the simpler change in this commit.


msoos:
Looks good, thanks! I'm currently in the process of fixing the STP compile in Travis&AppVeyor, so I am skipping wainting for that check to finish. Thanks again!

msoos:
It worked for me, including tests. So I think it's OK. CMS is broken, so some of the builds will break. It should be back to norm shortly.

PR #266, opened 2017-08-14
circumvent Xcode bug with CMake's TARGET_OBJECTS

csinz:
STP currently cannot be built with CMake using Xcode as a generator.
This is due to a bug that occurs with TARGET_OBJECTS in current Xcode versions.
The submitted fix is also used similarly in other projects.


msoos:
Thanks! Went through both checks, it's all good :)

PR #268, opened 2017-10-03
Provide thread safety if C++11

andrewchi:
Enable STP to be initialized and called from multiple threads simultaneously. This becomes active if there exists C++11 support for the `thread_local` keyword.  If C++11 is not supported, then nothing changes.

If this causes a measurable performance hit, then it may be worth providing a compilation flag (in Attributes.h) to enable/disable this.


andrewchi:
I'm investigating why the STATIC_CMS build fails the tests. Input welcome, obviously.

msoos:
Hey,

Woooow. I'll fix the CMS build fail, I probably broke it. Don't worry, this
is some amazing work! I'll merge ASAP :) Thanks so much! I'll fix CMS
today, before merging, too :)

Thanks again and no worries, CMS failing probably has nothing to do with
this :)

Thanks again!

Mate

andrewchi:
Thanks! My use case for this code was a multi-threaded fork of KLEE that runs 16 threads, each with its own STP solver.  It seems to work there.

Granted, there is some functionality that I didn't exercise. For example, there I'm using minisat rather than CMS (are there threading interactions?). It's also possible that marking `_string_lit` as `thread_local` in lib/Parser/smt2.lex and lib/Parser/smt.lex interacts badly with the parser generator, but that seems odd to me.

But more importantly, I assume you want some unit tests exercising this functionality (e.g., run one of the existing tests simultaneously in multiple threads). I wasn't sure how you wanted that designed, so feel free to point me in the right direction.

rgov:
Sounds like a good feature for STP but did you consider just running 16 instances of KLEE with different starting seeds so they explore different parts of the program state? This is the strategy that AFL uses on multicore systems, and it simplifies the fuzzer's implementation.

andrewchi:
We did think about doing it that way as well, but our work's actual objective was slightly different from AFL and KLEE's normal use case of finding bugs quickly.  Instead, our objective was to verify that communication over an untrusted network is consistent with legitimate client software (https://www.usenix.org/conference/nsdi17/technical-sessions/presentation/chi), an strategy that could have caught, e.g., a Heartbleed attack without using Heartbleed-specific configuration. Translating it back to the symbolic execution setting, the goal was to look for a code path that explains the packets we've seen over the network.  In the end, some of our search logic required more inter-thread coordination than just telling each thread, "go off into your corner and search that space."  But I agree that the implementation would have been _far_ simpler if we could have just run separate instances. I may yet revert to that strategy in future work.

We had been using a much older STP for our work, but as we brought everything up-to-date, I figured I'd try to contribute a useful piece of it.  It might be handy to others to be able to have multiple STP solvers going at the same time.

msoos:
Of course, it is useful indeed. Let us know what other improvements you
have made, I'm eager to incorporate them too, in case they are open source.
I'll look at your link later, seems really interesting :) Thanks again!

include/stp/Util/Attributes.h
```diff
@@ -57,4 +57,11 @@ THE SOFTWARE.
     #define DLL_LOCAL
 #endif
 
+// Provide thread safety if possible
+#if __cplusplus >= 201103L
+    #define THREAD_LOCAL thread_local
+#else
+    #define THREAD_LOCAL
```
msoos: We use C++11 so thread_local is there. If it's not there, setting it empty is probably a bad idea, as for some it would fail to run multi-threaded, and for some it would not... leading to very hairy debugging issues :O

andrewchi: Ok, thanks. Then in that case maybe we should just use `thread_local` directly instead of using the macro.  There are a couple of places where I inserted #include "stp/Util/Attributes.h" in order to get the macro definition (lib/extlib-abc/aig.h, lib/extlib-abc/aig/kit/kitTruth.c, and lib/extlib-constbv/constantbv.cpp) those can be removed too.

msoos:
Well... after reading the code, I am not convinced this is the best solution, but it's definitely a move in the right direction. The issue is that the THREAD_LOCAL define has effectively masked the issue of C not having thread_local, which means all C code could potentially have a race, if it's called. For example, the lexer is most likely NOT called in a multi-threaded scenario, as the library is being used -- but it could be, it's part of the c_interface file, and the lexer produces code can can definitely race. For the issues with libabc, I still need to investigate -- it might not be used in the code, but it might be, we need to check. In any case, this is not bulletproof and some issues might arise and debugging it could be pretty difficult. I will look at this, but it's not a perfect solution -- but a good start!

Thanks again,

Mate

andrewchi:
Right. I originally thought about having the macro use either `__thread` or `__declspec(thread)` (depending on GCC or Clang), which would accomplish the same thing for the C parts.  But I wasn't sure whether you would ever want to compile those separately or whether everything was just going to be built by a C++ compiler.

msoos:
Yep, good points. I'll think about it and let you know what I came up with :) I might actually fix the underlying issue... I'm not sure yet. In any case, you started the fire, now I'm gonna try to finish it :D But maybe I'll resort to those in case I cannot fix it another (clear, maybe) ways.

andrewchi:
Thanks for the fixes you made in the master branch; I think they'll work.

I just ran my fork of KLEE (with 24 STP instances) in Intel Inspector and it found a data race in a couple of global variables I missed: `propagatorNotImplemented` and `iterations` in the header file `include/stp/Simplifier/UnsignedIntervalAnalysis.h`.  Neither affects correctness of the solver, but we ought to get it right.

I'm now testing a version with them both marked THREAD_LOCAL to make sure there aren't any more data races in STP, or at least ones detectable by Intel Inspector.

msoos:
Nice, by the way, Intel Inspector. Never had access to it. Seems like it is doing a good job :)

PR #269, opened 2017-10-05
Mark two more global variables as thread_local

andrewchi:
I missed these in the first version of the "thread-safe" code.

I discovered this by running my fork of KLEE (with 24 instances of STP) under Intel Inspector. Neither of these affects correctness of the solver, but one does affect correctness of diagnostic output. In any case, we should avoid all data races.

### DIFF
--- a/include/stp/Simplifier/UnsignedIntervalAnalysis.h
+++ b/include/stp/Simplifier/UnsignedIntervalAnalysis.h
@@ -45,8 +45,8 @@ namespace stp
 {
 using std::make_pair;
 
-unsigned propagatorNotImplemented =0;
-unsigned iterations =0;
+THREAD_LOCAL unsigned propagatorNotImplemented =0;
+THREAD_LOCAL unsigned iterations =0;
 
 void print_stats()
 {


msoos:
Thanks!

msoos:
So on Windows, I am getting this: 

See in detail here: 

https://ci.appveyor.com/project/msoos/stp/build/1.0.134

```
c:\projects\stp\include\stp\globals\globals.h(88): error C2492: 'GlobalSTP': data with thread storage duration may not have dll interface [C:\projects\stp.build\lib\AST\AST.vcxproj]
c:\projects\stp\include\stp\globals\globals.h(89): error C2492: 'GlobalParserBM': data with thread storage duration may not have dll interface [C:\projects\stp.build\lib\AST\AST.vcxproj]
c:\projects\stp\include\stp\globals\globals.h(90): error C2492: 'GlobalParserInterface': data with thread storage duration may not have dll interface [C:\projects\stp.build\lib\AST\AST.vcxproj]
  ASTNode.cpp
[... keeps repeating for all files...]
```

Did you encounter this? Did you compile this under Windows?

andrewchi:
Unfortunately, I didn't compile this on Windows, and I clearly should have.  I'm not sure how to fix this without breaking the existing interface (i.e., requiring programs using STP to be recompiled).

msoos:
Sure, no worries. I am in the process of fixing it, was up until 3am yesterday trying to make the bison&flex reentrant, which is the better way of doing this :) Note that thread_local, is not really viable in the long term as the whole STP needs to be re-architected, but is a good temporary work-around! Thanks again!

PR #289, opened 2018-04-25
Riss

conp-solutions:
This commit series adds Riss as yet another SAT backend for STP. Furthermore, the new backend is added to Travis, and the resulting stp is tested with a fuzzer.

Furthermore, the build directory is added to the .gitignore file.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,3 +8,4 @@ docs/_build
 
 .DS_Store
 .direnv
+build
--- a/.travis.yml
+++ b/.travis.yml
@@ -29,6 +29,7 @@ env:
     - STP_CONFIG=COVERAGE
     - STP_CONFIG=STATIC_CMS
     - STP_CONFIG=INTREE_BUILD
+    - STP_CONFIG=STATIC_RISS
 
 matrix:
   include:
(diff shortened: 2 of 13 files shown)


msoos:
Thanks! Sounds good actually! I'm interested in others' views before we can merge through :)

One thing I would change is to let the user decide which solver to use: riss, minisat, or cms. I think such an option, on the command-line, would make sense and may increase the use-case of STP.

conp-solutions:
If I understand the request correctly, you want an option --sat=<solver> {cms,minisat,riss} ? Isn't this exactly what the --cryptominisat, --minisat and --riss options already do? I can add another option that competes with those three, but this does not feel right. I'd either keep the 3 that currently exist, or replace them with a single option that has the same effect.

msoos:
Ah, sorry, I thought that you replaced minisat with Riss :) I'll take a
second look and merge :)

Mate

msoos:
OK, looks good except you maybe want to use

https://cmake.org/cmake/help/v3.0/command/find_library.html

To find Riss? Otherwise it might be painful for others to compile it with Riss? I'm merging but maybe you want to create another pull request to fix that so it's easier for other to use it with Riss

PR #291, opened 2018-05-22
Starexec

conp-solutions:
To prepare a starexec package, I used a script and added some template files.

### DIFF
--- a/scripts/make-starexec.sh
+++ b/scripts/make-starexec.sh
@@ -1,0 +1,117 @@
+#!/usr/bin/env bash
+#
+# Copyright (c) 2018 Norbert Manthey
+#
+# This script creates a package for starexec, including statically linked
+# third-party SAT solvers
+#
+# The script will install m4ri on the system from source.
+
+# make sure we notice failures early
+set -e -x
+
+# make sure we know where the code is
+SOLVERDIR=$(pwd)
+BRANCH=$(git rev-parse --abbrev-ref HEAD)
+
+if [ ! -x "$SOLVERDIR"/scripts/make-starexec.sh ]
+then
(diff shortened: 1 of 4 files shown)


msoos:
Wow, great work, thanks!!

msoos:
:) By the way, do you know who is preparing STP-riss and STP-cms? Are you?

msoos:
Oh, the PDF... I have just added you access to write as well:

https://github.com/stp/docs

Yeah, mostly the only thing that changed is the solver backend :) You can write just one description with riss and cms backends both, if you want. Up to you.

conp-solutions:
Sure, I can do that. I just need to get access to starexec until then. For STP-CMS-mt, I basically take the same starexec configuration file, but add the option "--threads $(nproc)" to the cryptominisat command line, correct? Given the way STP-mt works, using a simple portfolio solver instead of a single one sounds promising, but I will leave that for next year ;)

conp-solutions:
I submitted the mentioned variants except Riss-mt, since Riss parallel variant is not well tested or tuned.As all current submissions use the SAT solver as standalone binary via a wrapper script around STP, I wonder more whether we should submit an STP that uses an internal Solver. This might be closer to what users of STP get, no? Do you feel any need for this kind of submission?Best,Norbert

msoos:
Dunno... you are right about "this is what users are getting" -- and I usually really try to optimise that rather than optimising for winning, but I am not really a pro at SMT so I don't know the trade-offs involved. I think it's good the way it is now. I am now thinking of forking STP and "making it my own" -- open-source of course. It will take forever before I manage to do that :S Anyway, thanks so much for submitting :) We will probably need to update for the latest Riss and CMS before the final deadline though :)

PR #292, opened 2018-05-25
starexec: add multicore configuration

conp-solutions:
As the SMT competition supports the parallel execution of SAT solvers,
allow to use cryptominisat in its parallel mode, using all cores that
are available automatically.


msoos:
Thanks so much!

PR #304, opened 2018-12-14
try to get >64 bit consts working.

### LINKED ISSUE #303: Arbitrary-length bitvector constants?
I see that bitvector constants over `sizeof(unsigned long long int)` are an error per: https://github.com/stp/stp/blob/c68788eb05731ae08e4f730a6b1c93803620e2b9/lib/STPManager/STPManager.cpp#L187
Is there no way to create a larger bitvector constant value?

TrevorHansen:
I've tried to fix #303 by exposing the function that creates bit-vector constants from decimal strings.

### DIFF
--- a/bindings/python/stp/stp.py.in
+++ b/bindings/python/stp/stp.py.in
@@ -275,6 +275,13 @@ class Solver(object):
         expr = _lib.vc_bvConstExprFromLL(self.vc, width, value)
         return Expr(self, width, expr)
 
+    def bitvecvalD(self, width, value):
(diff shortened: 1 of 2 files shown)


TrevorHansen:
Hi @msoos , do you or anyone else think I should clean up the patch before applying?

msoos:
Hey @TrevorHansen @maxzinkus 

Looks good to me :) As long as you the python name of the function is not insane (I think it looks OK), I think this should be fine :)

msoos:
Thanks so much!

PR #311, opened 2019-03-29
Allow bitvecs of length >64

maxzinkus:
PR #304 enables this, but this check prevents it from working. Not sure why the tests aren't catching it.
Also, `width <= 0` is equivalent to `width == 0` as `width` is unsigned.

@TrevorHansen

### DIFF
--- a/lib/STPManager/STPManager.cpp
+++ b/lib/STPManager/STPManager.cpp
@@ -184,7 +184,7 @@ bool STPMgr::LookupSymbol(const char* const name, ASTNode& output)
 ASTNode STPMgr::CreateBVConst(unsigned int width,
                               unsigned long long int bvconst)
 {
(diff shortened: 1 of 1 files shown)


rgov:
Is there an existing test that is passing when you don't think it should? Could you please make a test that fails without your change and passes with it?

maxzinkus:
There are multiple definitions of `CreateBVConst`, some take Str, DecStr, or Int. The DecStr one is in the test, and works just fine in the correct test case. The Int one is used internally when bvconsts are created, and has the width check I'm removing here. So the tests are actually fine, but this makes the interface consistent

rgov:
Ok, I see that. Thanks.

maxzinkus:
Thanks!

Issue #287, opened 2018-03-28
SMTCOMP


TrevorHansen:
If STP is submitted practically unchanged, perhaps with a novel SAT solver, to SMTCOMP. Does anyone think we should ask for anything, or be involved in some way, beyond just an acknowledgement in the system description that STP is used?

The SMTCOMP rules already have the notion of a "wrapper tool", which covers the need to acknowledge:
_Wrapper tools. A wrapper tool is defined as any solver which calls one or more SMT solvers
not written by the author of the wrapper tool. The other solvers are called the wrapped solvers. A
wrapper tool must explicitly acknowledge any solvers that it wraps. Its system description should
make clear the technical innovations by which the wrapper tool expects to improve on the wrapped
solvers._

msoos:
What novel SAT Solver? :) Norbert wrote to me that he wants to submit with his SAT solver, Riss. I'm happy for that but if so, he probably needs to add an issue here, so we can track it, right? I think it's important to be transparent and I urged him to create a pull request.

By the way, CMS has been significantly improved and you should see massive speedup compared to last year(s). So it might be an interesting challenge to see if CMS or Riss would be the best to use -- we should submit with both. I am all-in for some friendly competition, but for that to happen we need to have this tracked under a GitHub issue.

conp-solutions:
As Mate already said, I'm eager to see whether Riss can improve the performance of STP on the one hand, and I'd love to see how Riss performs as a SMT backend in general. The source of the solver is hosted here: https://github.com/conp-solutions/riss

I can create the pull request and make Riss another SAT backend in the list of backends, I just want to know what's a reasonable way. Concerning the submission: I'm happy to go the solver wrapper way, or anything else. As Mate, I'm all for submitting more versions. Honestly, I do not know whether Riss performs better than the currently used solvers in the SMT context.

That being said: can you shed some light on how the multi-threaded version of STP works? Is it "just" replacing the sequential SAT backend with a parallel one, or do you do more?

TrevorHansen:
Hi Norbert,

You're welcome to integrate Riss with STP. In the "SAT" directory are examples of how the other SAT solvers integrate. If it doesn't make sense, let me know and we can discuss.

STP isn't multi-threaded itself, it just calls a multi-threaded SAT solver.

[...]

PR #312, opened 2019-04-03
starexec: add portfolio script

conp-solutions:
When running with multiple cores, a plain portfolio approach for
several SAT solvers is a viable approach for a small amount of
solvers. When more resources are available, smarter solutions with
sharing should be considered.

This commit adds a script that implements the above idea to a certain
extend. Based on the available resources, the SAT backend uses
the solvers CryptoMiniSat, Riss and MiniSat in parallel. While Riss and
MiniSat get at most one core, CryptoMiniSat is assigned all remainint
resources. This way, we use several solvers, and CryptoMiniSat can share
information about its parallel solver threads.

#287 Implements a first portofolio prototype

### DIFF
--- a/scripts/starexec/bin/multi-solver.py
+++ b/scripts/starexec/bin/multi-solver.py
@@ -1,0 +1,126 @@
+#!/usr/bin/python
+
+import os
+import shutil
+import subprocess
+import sys
+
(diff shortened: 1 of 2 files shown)


msoos:
Sounds great, thank you so much!

PR #314, opened 2019-04-18
Fix two format specifiers.

jamesjer:
The loop variables have type size_t.  On a big endian 64-bit architecture, such as s390x, the %d specifier instructs sprintf to examine only the higher-order 32 bits, which are probably all zero.


rgov:
Thanks!

PR #315, opened 2019-04-29
Python unit test fixes

andrewvaughanj:
This branch fixes three issues with the Python tests:

1. In the "solver" tests, there were no assertion messages, meaning that failures were hard to debug;
2. Also in the "solver" tests, the calculation of the "other solvers" was actually incorrect and raised a Python exception -- it seems these **never** worked (see below);
3. It actually seems that any failures in the Python tests did not record a test failure (`add_test` in CMake only records a failure on a non-zero error code; with the code as it was, `python` would return `0` irrespective as to what happened during test execution!).

### DIFF
--- a/tests/api/python/tests.py.in
+++ b/tests/api/python/tests.py.in
@@ -322,7 +322,7 @@ class TestSTP(unittest.TestCase):
         # all other solvers should not be in use
         #
         if supported:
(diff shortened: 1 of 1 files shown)


msoos:
Thanks! Looks cool :)

PR #316, opened 2019-04-29
Non-install 'library_path.py' now searches 'lib' and 'bin' folders (now consistent with 'library_path.py.in_install')

andrewvaughanj:
On Windows, STP's CMake places `libstpwin.dll` inside of the `bin` directory (`CMAKE_RUNTIME_OUTPUT_DIRECTORY`), and not inside of the `lib` directory (`CMAKE_LIBRARY_OUTPUT_DIRECTORY`), which it does on Linux.

This means that the construction of `library_path.py`, which is different to `library_path.py.in_install`, does not work correctly on Windows (as it searches `lib` and not `bin`).

In effect, the file `./bindings/python/stp/library_path.py` gets auto-populated to have Python's search path set to only look in `lib`. This causes the automated Python tests to fail on Windows.

**Note**: I was unsure of the best way to fix this. One thought was to have two `.in` files (one for Windows and one for non-Windows). However, I decided to keep there being one `.in` file, but have some code commented out if `configure_file` is called on non-Windows. This is what `WIN32PYTHONCOMMENT` achieves.

### DIFF
--- a/bindings/python/stp/CMakeLists.txt
+++ b/bindings/python/stp/CMakeLists.txt
@@ -30,6 +30,18 @@
 
 get_target_property(LIBSTP_BASENAME stp OUTPUT_NAME)
 set(LIBSTP_FILENAME "${CMAKE_SHARED_LIBRARY_PREFIX}${LIBSTP_BASENAME}${CMAKE_SHARED_LIBRARY_SUFFIX}")
+
+#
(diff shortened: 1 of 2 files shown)


msoos:
Well, it looks like a very messy fix but if it works.... maybe not so bad :)

PR #317, opened 2019-05-01
Making sure STP's '__init__.py' is consistent with '__all__' in 'stp.py'

andrewvaughanj:
It seems that we do not have consistency between STP's `__init__.py` and the symbols that we try to export via `__all__` in `stp.py`.

STP's Python modules are structured like this:

     buildings/python/stp/stp.py

Therefore, if you set PYTHONPATH path to point to `buildings/python` (i.e., the parent directory of the `stp` **folder**) then you do not get our newly-exported symbols, as `import stp` imports the folder (i.e., `__all__.py`) and not the module (i.e., `stp.py`).

This PR (manually) ensures that what's in `__init__.py` is consistent with `__all__`.

### DIFF
--- a/bindings/python/stp/__init__.py.in
+++ b/bindings/python/stp/__init__.py.in
@@ -1,5 +1,8 @@
-#author: Dan Liew
-#date: Sep, 2014
+# author: Dan Liew
+# date: Sep, 2014
+#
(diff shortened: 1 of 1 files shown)


msoos:
Thanks, this makes a lot of sense!

PR #318, opened 2019-05-01
Exposing the underling `_lib` object from STP's Python API

andrewvaughanj:
STP's `Solver` Python class only has wrappers for _some_ of STP's C-based API (e.g., it does not expose `vc_bvConcatExpr` or `vc_isBool`).

In some instances, it there is therefor beneficial to be able to work directly with the `ctypes.CDLL` object, rather than the "proxy" `Solver` class.

In the longer term, it is probably worthwhile identifying all of the calls that are missing, and then exposing these via the `Solver` class.

However, in the interim, this PR does resolve that some of these calls are currently inaccessible, and therefore makes STP's Python API "productive".

### DIFF
--- a/bindings/python/stp/__init__.py.in
+++ b/bindings/python/stp/__init__.py.in
@@ -35,4 +35,5 @@ from .stp import (
     get_git_version_sha,
     get_git_version_tag,
     get_compilation_env,
+    get_lib,
 )
--- a/bindings/python/stp/stp.py.in
+++ b/bindings/python/stp/stp.py.in
@@ -34,7 +34,7 @@ import sys
 
 __all__ = [
     'Expr', 'Solver', 'stp', 'add', 'bitvec', 'bitvecs', 'check', 'model',
-    'get_git_version_sha', 'get_git_version_tag', 'get_compilation_env'
+    'get_git_version_sha', 'get_git_version_tag', 'get_compilation_env', 'get_lib'
 ]
 
 Py3 = sys.version_info >= (3, 0, 0)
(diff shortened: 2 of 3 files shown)


msoos:
I understand the intention here, but I am afraid -- if we do this, people will start relying on this and then we will _never_ be able to change the C API. That would be very sad, don't you think? It's a pretty broken API and I'd rather just do the wrapper if possible? Which ones do you really need? Maybe I could just work on exposing those and then we should be fine? Then we can change the underlying API and fix the issues without affecting Python usage. What do you think?

andrewvaughanj:
Oh, I do (genuinely) agree that it isn't ideal ... unfortunately, a while ago, I wrote something that I am now trying to "productise" that relies on this. One thing I can do (if we don't want to merge this, and that is **100% fine**) is just patch my shipping version of `stp.py`. I really should have created a PR rather than creating my own proxy object, but I wasn't aware of how amenable STP was to PRs :)

I guess on the question of the C API -- if people write something _today_ in C, they'd be bound to it, right? I'm not sure there's any difference in having those "breaking changes" either in the direct C API or in the Python API.

As an example, in my "internal" usage, we have our "type cache", which means I'm more than happy to bypass `Solver.bitvec` and just use `_lib.vc_varExpr` directly.

Right now (and from a very hacky way of checking), there are 49 calls that are in `_lib`, but which are not reflected in the `Solver` class:

```
for i in $(grep "DLL_PUBLIC" include/stp/c_interface.h | egrep -oh "vc_[^(]*"); do echo -n "$i "; grep -c $i /home/avj/stp/stp/bindings/python/stp/stp.py.in; done | grep "1" | wc -l
```

msoos:
OK, well, it's sad, but we'll have to live with it then :( Merging. We need to fix this thing for good at one point...

PR #322, opened 2019-05-20
Support time-based timeout when calling Cryptominisat

andrewvaughanj:
This PR introduces functionality to utilise Cryptominisat's time-based timeout functionality, to allow checking assertions with a given timeout.

To work entirely correctly, this PR depends on [CMS PR 555](https://github.com/msoos/cryptominisat/pull/555). Without this PR, STP's solver timeouts are based on the starting age of the underlying process and not the age of the solver.

Given this change in CMS, I think it would be best to add more tests to STP that have timeouts across multiple checks/solvers, and where the timeout is shorter than the process age, but still a valid timeout (e.g., a 1 second timeout on a 10 second old process).

I am opening this PR now to allow for early review/comments.

### DIFF
--- a/lib/Sat/CryptoMinisat5.cpp
+++ b/lib/Sat/CryptoMinisat5.cpp
@@ -103,7 +103,7 @@ bool CryptoMiniSat5::solve(bool& timeout_expired) // Search without assumptions.
   if (ret == CMSat::l_Undef)
   {
     timeout_expired = true;
-    // TODO: do we want an assertion on the cause of the timeout
+    // TODO: do we want an assertion on the case of a timeout
   }
   return ret == CMSat::l_True;
 }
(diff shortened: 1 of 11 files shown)


lib/Sat/CryptoMinisat5.cpp
```diff
@@ -90,11 +95,20 @@ bool CryptoMiniSat5::solve(bool& timeout_expired) // Search without assumptions.
...
+  if (max_time > 0) {
+     s->set_max_time(max_time);
+  }
+
   CMSat::lbool ret = s->solve();
   if (ret == CMSat::l_Undef)
   {
     timeout_expired = true;
-    assert(s->get_sum_conflicts() >= max_confl);
+    // TODO: do we want an assertion on the case of a timeout?
```
msoos: No, **definitely** not. Please remove this TODO -- it should not be there. The time can run while e.g. someone had a debug point set or the time can be weirdly linear and even non-linear (think of summer time, winter time, people changing the time by hand, etc). So time is VERY weird and we should definitely not assert-fail because of that.

andrewvaughanj: Done!

msoos:
Thanks! Merged :)

PR #325, opened 2019-05-21
Removing the (broken) timeout tests to restore builds on Travis

andrewvaughanj:
This is a quick PR to suppress the time-based timeout tests while I can resolve them.

This is not a permanent solution, but it should resolve the broken build.


andrewvaughanj:
This PR removed the time-based timeout tests and left the conflict-based ones behind. What's odd is that, on Travis, the conflict ones are still failing.

I'm working on this.

PR #326, opened 2019-05-21
Correcting timeout-based tests to avoid failures on Travis

andrewvaughanj:
Three changes:

- Conflict-based timeout tests are now performed with both CMS and Minisat (both Python and C)

- Python timeout tests are now much more intensive to ensure we hit a timeout (and avoid false negatives during testing)

- C timeout tests have a significantly reduced timeout to ensure we hit a timeout (and avoid false negatives during testing)

Tested locally on my machine, but unclear how the performance of my machine matches Travis. Hopefully my machine is more powerful, and therefore if I hit a timeout, Travis should do as well.

### DIFF
--- a/tests/api/C/timeout.cpp
+++ b/tests/api/C/timeout.cpp
@@ -106,7 +106,6 @@ TEST(timeout_conflicts_minisat, one)
 }
 
 #ifdef USE_CRYPTOMINISAT
-#if 0
 /*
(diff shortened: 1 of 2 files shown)


andrewvaughanj:
Okay, so that's Travis clean!

msoos:
Thanks!

Issue #306, opened 2019-01-23
Bitvecval value in python


steseb:
It seems that the Python interface doesn't provide the right name/value for the a constant bit vector.

Please see the small example below:
```
import stp
a = stp.Solver()
x = a.bitvec('x')
x.name
>>> 'x'
x.value
>>> 0
y = a.bitvecval(32, 18)
y.name
>>>
y.value
>>> {'x': 0}
```

if `x.value == 0` could just stand for a placeholder for the symbolic bit vector, I don't understand why `y.value` returns `x`

msoos:
Ah that sounds crappy... could you please take a look in the python interface code and check what could be causing this? Maybe you could find it? I'd be really grateful :) We are a bit busy doing a lot of other stuff so a helping hand would be great! Thanks :)

steseb:
I tried to have a look.

Assuming that it is ok to get 0 for the `value` of a new `bitvec`, the problem with the `bitvecval` seems that `Expr.value()` searches in the `Solver.model()` by name, but BV constants don't have one.

If no name/key is provided, `model` returns all the expressions having a key i.e., only the `bitvec`s. 

I can't solve this.

andrewvaughanj:
@steseb I forgot about this; there's a chance this is done -- see: #328 

If you're happy, can you close the issue?

PR #327, opened 2019-05-31
Support for getting the value of a constant bitvec in the Python API

andrewvaughanj:
Attempt to correct the behaviour specified in #306.

Looks like this works and passes the Python suite.

New test added to validate what's listed in #306.


andrewvaughanj:
I don't really know what happened here ... I (temporarily) closed this to fix something, but then I couldn't re-open the PR (hence why another one exists).

¯\_(ツ)_/¯

PR #331, opened 2019-06-16
Static

conp-solutions:
This change is to not have the static vs dynamic binary problems for the SMT Comp submission.


msoos:
Cool, thanks!

PR #335, opened 2020-01-09
Better support for python in build

mimi1vx:
Python devel libraries aren't needed for simple ctypes bindings
Remove restriction on python version to allow build also with other pythons than 2.7


msoos:
Hi,

I believe this is wrong. The devel headers in my experience are needed, the bindings won't build without them. The python executable is needed due to the python binding being tested. Also, you removed the printing of

```
message(STATUS "PYTHON_EXECUTABLE:FILEPATH=${PYTHON_EXECUTABLE}")
```

Which prints the `PYTHON_EXECUTABLE` which you actually check for hence is definitely needed as per your own check. However, again, the libraries and headers are needed in my experience -- if the include_dir is not found then we will have trouble including the python.h which we need to build. Similarly, the linking will fail in case the libraries are not included. Are you absolutely _certain_ these are not needed? Have you tried this in at least 5-10 environments. Our user base is _extremely_ wide. These checks are there due to our experience. Do you have tested this change _at least_ on each of Windows, Mac, and Linux? We cannot accept any build patch that has not been tested at least on one of each of these environments.

rgov:
@msoos When you say the bindings won't "build" without them, what do you mean?

https://github.com/stp/stp/tree/b4b63b859323d4494924be3bf98d2131fdaa8f07/bindings/python/stp

The bindings are all .py files, what Python-specific C code is built anywhere? At a quick glance I didn't see any files that `#include <Python.h>` which makes me think the Python headers really aren't needed.

msoos:
Damn, that was late at night -- I confused it with CryptoMiniSat, I'm so
sorry! Thanks for noticing this, I basically just copy pasted the CMS
version and that indeed was wrong, good catch! Oh okay, so only the Python
2vs3 needs to be checked (but bit enforcing a version seems like the right
thing to do) and the executable found needs to be printed and we are good
to go :)

msoos:
Sorry again!

mimi1vx:
for python3 ... it need only small change in stp.py --> fix import from 'library_path' and probadly check if is list of library PATHS generated correctly

see: https://build.opensuse.org/package/view_file/devel:tools:statica/stp/py3.patch?expand=1

mimi1vx:
@msoos  tested on linux, and now on way to openSUSE :D as other systems, I haven't any of them

andrewvaughanj:
What's the `.2.3` about? How does that work on Windows?

andrewvaughanj:
Also, how does that PATHS change work for testing? When testing works, it wants to load the STP code from the CMake build folder -- it looks like your changes has it loading it from the install directory. This might mean that you can't run tests against a development (i.e., non-installed) version of STP.

mimi1vx:
> What's the `.2.3` about? How does that work on Windows?

ctypes needs exact path to library to load it, so this is adding the missing part on linux. 
libstp is installed to:

`/usr/lib/libstp.so.2.3` (32bit system) or `/usr/lib64/libstp.so.2.3`

unfortunately current code in `library_path.py.in_install generates something like `/somethig//usr/lib/libstp.so/`

andrewvaughanj:
Doesn't `make install` (or similar) symlink `libstp.so` to `libstp.so.2.3`? Also, I believe that `2.3` comes from the version of STP (currently 2.3.x), so hard-coding that value probably isn't the best.

Plus, `@LIBSTP_FILENAME@'` should be `libstp.dll` (or similar) on Windows. If you slap `2.3` on the end of that, it then won't be found by Windows' loader.

msoos:
Thanks so much, and sorry for being confused about it at the beginning!

Please let us know if there is anything else that needs fixing -- seems like there is a discussion going on here about some changes that may need to be made? Could one of you please create a pull request if indeed it's needed/useful? Thanks a lot in advance! Would be nice to have it all fixed :)

andrewvaughanj:
@msoos my comments were on @mimi1vx‘s changes related to building as part of OpenSuSE — the merged PR doesn’t worry me for the Windows build.

mimi1vx:
btw, instead of a list of libraries to try load, we should use `ctypes.util.find_library('libname')`

msoos:
@mimi1vx  Sounds interesting! Do you wanna do a pull request to fix that?

PR #336, opened 2020-02-12
Warnings

bpfoley:
Quell several GCC and one bison warning that shouldn't affect the meaning of the code,
but remove some worrying warnings during build.


msoos:
Thanks for the cleanup!

PR #337, opened 2020-02-12
Remove unused ForDecl node from the grammar.

bpfoley:
Bison was producing warnings about this.

I'm not sure if this was ever used, but it's not referenced in the grammar,
and isn't mentioned in the docs.


rgov:
Historical note: It was introduced in bd16295ccb50264ba2113397126851f6babc1759.

I would agree with its removal. @msoos?

msoos:
Sure, merging :)

PR #338, opened 2020-02-12
Fix bison warning "rule useless in parser due to conflicts"

bpfoley:
The following rules were conflicting:

ArrayUpdateExpr : Expr WITH_TOK Updates
Updates : '[' Expr ']' ASSIGN_TOK Expr
        | Updates WITH_TOK '[' Expr ']' ASSIGN_TOK Expr

I think it was intended to allow

a [ e1 ] := val
OR
a WITH [ e1 ] := val

### DIFF
--- a/lib/Parser/cvc.y
+++ b/lib/Parser/cvc.y
@@ -1021,13 +1021,13 @@ Expr            :      TERMID_TOK { $$ = new ASTNode(GlobalParserInterface->letM
 ;
 
 /*Grammar for Array Update Expr*/
-ArrayUpdateExpr : Expr WITH_TOK Updates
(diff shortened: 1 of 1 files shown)


msoos:
Actually, CVC should probably ripped out entirely... plus I don't know about these rules well enough. Anyone else who could take a look perhaps? @TrevorHansen perhaps?

rgov:
@msoos In the meantime, what about accepting this change? It is less destabilizing than yanking CVC altogether.

msoos:
Sure, I'm merging -- but I don't really see the point of putting work into fixing up a system that we are currently not using anyway. This seems noisy and not making much progress :(

PR #340, opened 2020-02-14
Quell 0 length array warnings if present, as STP uses this C++ extension

bpfoley:


msoos:
Hi,

Actually, this one should be fixed correctly, rather than squashing the warning. This could be a real issue in the future, and I am not sure squashing the warning is the right solution here. There are ways of fixing it right.

PR #323, opened 2019-05-21
Provide an implementation of 'vc_bvRemExpr' to allow for API consistency

andrewvaughanj:
Inside of STP's Python API there was a call to a (non-existent) function of `vc_bvRemExpr`. If using the SMT front-end, the operation `bvurem` gets directly mapped to `BVMOD`.

Given that someone had "already made the mistake" of presuming that `vc_bvRemExpr` existed (e.g., `vc_sbvRemExpr` *does* exist), I provided an implementation that maps (at the API-level) from `vc_bvRemExpr` to `BVMOD`.

An alternative would have been to fix the Python API such that `rem` and `rrem` called `vc_bvModExpr` -- however, adding `vc_bvRemExpr` facilitate API consistency seemed like the better approach.

Tests have been added that would fail on older versions of STP, but give a correct value after this PR.

Signed-off-by: Andrew V. Jones <andrew.jones@vector.com>

PR #347, opened 2020-03-31
Ensuring that 'vc_bvRemExpr' is properly exposed to avoid a SIGSEGV via Python

andrewvaughanj:
In #323, the new API call `vc_bvRemExpr` was not correctly exposed via the Python API, leading to a segmentation violation during calls `rem` or `rrem`.

This PR fixes that.


msoos:
Ah, nice catch! Would have _never_ found that one! What an obscure bug, I understand why it wasn't obvious. I ran the tests and they run fine on my machine, too :) Thanks again!

PR #348, opened 2020-03-31
Fixed operator in STPConfig.cmake.in

ccadar:
@msoos many thanks for the quick fix!  You just got the string comparison operator wrong, and we get this error in KLEE:
```
CMake Error at /Users/cristic/stp/stp.git/build/STPConfig.cmake:15 (if):
  if given arguments:

    "" "==" "ON"
```
This trivial patch fixes it.


msoos:
Oh, of course, sorry, thank you!

Issue #349, opened 2020-04-09
SMTCOMP 2020


TrevorHansen:
The draft SMTCOMP 2020 rules have been released:

https://smt-comp.github.io/2020/rules20.pdf

The big change for us is that they're asking us to limit the number of configurations we submit to 2. I read this as making it difficult for us to submit STP with a range of different SAT solvers - like we did last year?

Do others read it the same way?

Trev.

andrewvaughanj:
Here’s the part:

> Limit the Number of Solvers. This year solver development teams may only submit more than one entry if they are substantially different. A justification must be provided for the difference. We strongly encourage the teams to keep the number of solvers per team per category at at most two.
>
> Rationale: The organizers wish to keep the total run-time of the competition small, in case of unforeseen problems that require to rerun the whole competition. The idea in allowing up to two submissions is to encourage the development of new, experimental techniques by allowing an “alternative solver” to be submitted while keeping the competition manageable.

I’m not sure it would be viable to even submit STP with two backends, as this does not even “substantially different”.

Also, if you (Trevor) are on the teams for both STP and Minkeyrink, you need to decide if they are “substantially different” enough to meet the intent of what the organisers wrote.

andrewvaughanj:
I feel I should clarify what I wrote, as I fear it could be interpreted in a "negative way".

I'm a 100% not saying that submitting Minkeyrink and STP is "not possible", only that there is a question about the "spirit of the law" (vs. the "letter of the law") and which people are members of the "STP team" and the "Minkeyrink team".

As Minkeyrink is closed-source, I have no idea how it differs from STP. It may already differ "substantially", in which case, submission of both STP and Minkeyrink is a no-brainer.

W.r.t. the original question: my interpretation of the "spirit of the law" is that submitting the same SMT solver with differing SAT backends would not be substantially different, and also that the "alternative solver" part doesn't cover this. My interpretation of the "alternative solver" is that you might have two _widely different_ ways of handling the SMT instance (e.g., MBQI vs. e-matching) inside of one tool. Submitting that tool in MBQI-mode vs. e-matching-mode would be fine.

msoos:
I think what we have to take into account is that many times these kinds of
rules are not really against Minkeyrink or STP. They are instead to deter
people who have been making minor changes and are not considered serious
contributors to the field. In other words, it's against abuse by parties
who are not contributing back in a significant way and only wish to profit
from the potential fame. This, I believe, cannot be said of those working
[...]

PR #354, opened 2020-05-24
Mergesat

conp-solutions:
This series adds MergeSat as another SAT backend. During working on it, I noticed that instead of just adding MergeSat, I could have actually implemented a compile-time replacement strategy for MiniSat.

As I currently only wanted to integrate MergeSat, I'll leave the other task open for now.

This series has not been tested extensively.

This is related to https://github.com/stp/stp/issues/349


conp-solutions:
The Travis checks fail with an error that is hard to understand given the current output. Interestingly, it does not seem to fail for Riss, alghouth Riss uses almost the same cmake and make invocation. Digging through this might take a while.

andrewvaughanj:
@conp-solutions sorry if this is a silly question, but why are the Riss and MergeSat back-ends only compiled statically? We recently disabled static builds through Travis (as they didn't work).

I'd love to get your stuff moving along, but I'd like to *not* build it statically if possible. I have nothing against the static builds, there's just work to do in getting them reinstated, so I don't want to block your PRs because of that.

conp-solutions:
There is no clear reason. I quickly managed to produce a static build, hence, i just used that. It might be possible to just drop the static keyword, I simply did not test that route, as static binaries have been the easier approach for the SMT competition.

Best,
Norbert

andrewvaughanj:
@conp-solutions do you think we can split off https://github.com/stp/stp/pull/354/commits/8e4ff44e5bb3dac529b916918590b797721c9680 into a separate PR? I don't think you need this for this branch (as it seems this branch *only* has `#define MINISAT_NSPACE Minisat` in it).

Genuinely, I get the intent, but it makes it hard to see what's Mergesat related vs. that change. If it works like I believe it does, then it would be good to have a CMake option to set which solver to use for `MINISAT_NSPACE`.

rgov:
This is a cool contribution but unfortunately it has since diverged from the codebase and would need some work to merge in.

@conp-solutions - do you still have interest in pursuing this?

conp-solutions:
I currently have no further plans of pursuing this, thanks. Let me close this pull request.

Issue #357, opened 2020-07-09
Get bitvector model as a vector of bool


mikhailramalho:
I'm using the `vc_bvConstExprFromStr` methods to create some large bitvectors (width > 64), but is there a way to get its value from the model also as an array of bools or a string of 0's/1's? I only found the `getBVInt`, `getBVUnsigned` and `getBVUnsignedLongLong` functions.

andrewvaughanj:
Joining the dots: https://github.com/Z3Prover/z3/issues/4568#issuecomment-655743461

Are you using the C or Python API? Do you want an STP `vc_boolType` or a language-level Boolean type (e.g., `bool` in C99/Python)?

andrewvaughanj:
So I was just taking a look at this for you, and I noticed this:

- [`getBVInt`](https://github.com/stp/stp/blob/master/lib/Interface/c_interface.cpp#L1607) calls `GetUnsignedConst`

- [`getBVUnsigned`](https://github.com/stp/stp/blob/master/lib/Interface/c_interface.cpp#L1621) calls `GetUnsignedConst`

The code for [`GetUnsignedConst`](https://github.com/stp/stp/blob/master/lib/AST/ASTNode.cpp#L235) seems to be limited to unsigned numbers that fit in a machine unsigned int (the message says 32, but I think it would be supported for a 64-bit value if you're on a 64-bit host).

When you say you're "creat[ing] some large bitvectors (width > 64)", have you created BVs that are larger than your machine-size `unsigned int` and successfully retrieved them over the API?

**Anyway** back to your original issue: behind the scenes [`getBVUnsignedLongLong`](https://github.com/stp/stp/blob/master/lib/Interface/c_interface.cpp#L1625) calls [`BitVector_to_Bin`](https://github.com/stp/stp/blob/master/lib/extlib-constbv/constantbv.cpp#L1489), which looks like it is close to what you need.

If an array of `char`s (as `0`/`1`s) would work for, I can probably expose this via the API. Do you need `bool`s, or is `char` okay?

andrewvaughanj:
Should be done; take a look at: https://github.com/stp/stp/pull/358

To be clear: this PR needs to be agreed by one of: @msoos @TrevorHansen @rgov -- it is possible that they might not want this API call.

mikhailramalho:
> So I was just taking a look at this for you, and I noticed this:
> 
[...]

PR #358, opened 2020-07-09
Adding an API routine to allow for obtaining a BV as a bit string

andrewvaughanj:
## Issue

#357 requested the ability to obtain a bit string from a given constant BV.

## Resolution

This PR adds an additional API call (`vc_printBVBitStringToBuffer`) that allows the user to obtain this representation.

As per the doc-string, the "output parameter" pointer *belong to the callee*.

Test added: `tests/api/C/bit_string.cpp`.


andrewvaughanj:
Travis is sad due to #359

rgov:
I'm not opposed to this but is it preferable to get this out as an ASCII string rather than some other sort of bignum representation? I guess this saves us the hassle of needing to choose a bignum library to make part of the interface.

andrewvaughanj:
That is how it is implemented; it returns an ASCII `char*` for the bit string.

rgov:
Right, I'm just saying that format is somewhat of a computational dead end without a bignum library that then parses it back into machine representation.

Just noticing that OpenSSL's `BN_bin2bn` doesn't parse binary strings, so that's out. GMP seems to have `mpz_set_str` which accepts whatever base.

andrewvaughanj:
Ah! Gotcha! Sure, I'm not sure why the user wants this, but I would guess it is because they have some parsing already for what Boolector returns ([the `.assignment` will be a bit string](https://boolector.github.io/docs/pyboolector.html)).

andrewvaughanj:
Probably this is a better reference [`const char *boolector_bv_assignment(Btor *btor, BoolectorNode *node)`](https://boolector.github.io/docs/cboolector_index.html):

> Returns
>
> String representing a satisfying assignment to bit-vector variables and a consistent assignment for arbitrary bit-vector expressions. The string representation will use binary [...]
>

mikhailramalho:
I was asking if we could get a binary string because it seems consistent with your C API; there is a method `vc_bvConstExprFromStr` that creates a bit-vector from a binary string after all.

I don't really mind using gmp, but it seems a bit of an overkill to add it as dependency only because of this function. We have our own bigint library in our tool, which converts to and from binary strings.

msoos:
I think this is a fine setup, and looking through the code, it all makes sense. No need to explicitly take a bignum library, the user can do it themselves quite easily I think :) Merging!

Issue #363, opened 2020-07-16
Process interaction stalls on `get-value`


robdockins:
When connecting to STP via pipes, the external process will stall when waiting for models or values from `get-value` or `get-model` commands.  It appears the output paths for these operations are missing a `flush` operation.  I was able to un-stall my use case by adding `flush(cout);` lines in the following linked locations.

https://github.com/stp/stp/blob/f6fd26830ded125fd31d15e07c465b6feea3ced6/lib/Interface/cpp_interface.cpp#L587
https://github.com/stp/stp/blob/f6fd26830ded125fd31d15e07c465b6feea3ced6/lib/Interface/cpp_interface.cpp#L610

I would not be surprised if there are other locations that need a similar treatment, but I'm not familiar with this code at all.

andrewvaughanj:
What happens if you wrap your call to STP with `stdbuf -i0 -o0 -e0` or `unbuffer`? See: https://stackoverflow.com/a/25548995

It might be worthwhile adding an option to ensure that _no_ output is buffered (see: http://www.cplusplus.com/reference/cstdio/setvbuf) which then means that we don't need to add flushes everywhere. I previously done that for other projects: https://github.com/googleprojectzero/halfempty/pull/16

robdockins:
I'm on MacOS, which unfortunately does not have `stdbuf` or `unbuffer` preinstalled.  I was able to install the `expect` package via homebrew (which has the `unbuffer` utility).  However, when I wrapped `stp` with `unbuffer` in a shell script, that does not seem to help, and actually seems to make things worse.  The `NSUnbufferedIO` env variable mentioned in the linked stackoverflow thread also does not seem to help.

Some way to force no buffering or line buffering would be one way to fix the problem.  Line buffering should probably be the default for an executable like this, IMO.

andrewvaughanj:
Do you have an example or anything that makes it clear how things break when the output is buffered?

I can create an STP branch that does `setvbuf`, but I don't know how to set that it "works" for you.

andrewvaughanj:
I have no idea if this is going to work for you: https://github.com/andrewvaughanj/stp/tree/use_linebuf

but it implements what I was suggesting

robdockins:
Unfortunately, it's a little tricky to write a short, encapsulated example demonstrating the problem.  Most of the standard utilities use pseudo-ttys to work around problems like this.  I'll see if that branch works for me.

andrewvaughanj:
[...]

PR #364, opened 2020-07-16
Use linebuf

andrewvaughanj:
Ensuring that STP's output is (at most) line buffered.

See: https://github.com/stp/stp/issues/363


andrewvaughanj:
This was originally a mess because I started with the wrong base branch and didn't check; forced pushed to correct that.

Issue #365, opened 2020-07-16
Additional SMTLib2 support


robdockins:
Would it be easy to add support for the `:global-declarations` option and the `get-info` operation? They are convenient for certain kinds of interactive use cases. Cf http://smtlib.cs.uiowa.edu/papers/smt-lib-reference-v2.6-r2017-07-18.pdf

```
$ stp
(set-option :global-declarations true)
unsupported

$ stp
(get-info :version)
(error "syntax error: line 1 syntax error  token: get-info")

$ stp
(get-info :error-behavior)
(error "syntax error: line 1 syntax error  token: get-info")
```

andrewvaughanj:
I might be able to take a look at `global-declarations` -- do you have a small, self-contained example of where this makes sense/is necessary?

This is where the SMTLIB2 front-end handles `pop`: https://github.com/stp/stp/blob/master/lib/Interface/cpp_interface.cpp#L356

So, **I think**, it is just a case of ensuring that we don't remove the symbols that you don't want `pop`ped.

It seems those `get-info` commands you've listed are required to claim SMTLIB2.6 support (see 4.1.8 of the SMTLIB reference), but I don't actually know if STP claims this.

andrewvaughanj:
If you care about `global-declarations`, do you also care about `reset-assertions`?

robdockins:
Sure, I can come up with a test case.

However, the larger context is quite a bit less self-contained.  What I'm doing involves symbolic execution of programs and I use interactive solving with lots of push/pops to keep track of the control-flow conditions that lead to assertions in the program.  We don't know in advance which parts of the potential problem space will be relevant to a solver query, so terms are streamed to the solver on-demand.  These terms can get *quite* large, so it's a shame to have to stream them again if they are forgotten on a `pop` (and the bookeeping on my end is much trickier to keep track of which terms need to be streamed again).

And, yes, good catch: `reset-assertions` is also relevant to this use case.

andrewvaughanj:
Baby-steps: if you can provide an easy/minimal/self-contained example, I can take look at getting that working (but no promises).

Then, once the easy bit is done, we can take a look at some larger examples :)

If you can give an example with and without `reset-assertions` that would also help!

robdockins:
I was working on a small example to demonstrated the desired behavior.  Somewhat to my surprise, STP already behaves as though `global-declarations` is set.

[STP-online-quickstart.txt](https://github.com/stp/stp/files/4939062/STP-online-quickstart.txt)

[...]

PR #367, opened 2020-07-20
Initial work on improving scopes such that symbols are correctly removed

andrewvaughanj:
## Issue

As identified in https://github.com/stp/stp/issues/365,  the creation of functions can allow for symbols to escape having their reference count decreased to zero, and therefore their lifetimes are longer than we would expect.

### Example

```
(push 1)
(declare-fun x!2 () Bool)
(define-fun x!3 () Bool (not x!2))
(pop 1)
(assert (not x!2))
(check-sat)
```

After popping `x!2` and `x!3`, it should not be possible to refer to `x!2`, but STP allows this.

## Solution

This PR reworks the symbol/function scopes inside of STP such that popping a frame *also* removes the functions declared in that scope.

### Discussion

I opted to keep a pointer to the "global" function context and then remove the popped functions during destruction of a frame.

An alternative would be change `Cpp_interface::applyFunction` to search through all of the frames (starting from the deepest) to find which to apply.

I have initially opted to tweaking how the scopes were stored, as this seemed like the less-invasive approach. Now that we have a `SolverFrame` class, we can improve `Cpp_interface::applyFunction` in the future to also use a "frame" of functions (rather than a global map).

### DIFF
--- a/lib/Interface/cpp_interface.cpp
+++ b/lib/Interface/cpp_interface.cpp
@@ -1,5 +1,5 @@
 /********************************************************************
- * AUTHORS: Trevor Hansen
+ * AUTHORS: Trevor Hansen, Andrew V. Jones
  *
  * BEGIN DATE: Apr, 2010
  *
@@ -39,7 +39,7 @@ namespace stp
 void Cpp_interface::checkInvariant()
 {
   assert(bm.getAssertLevel() == cache.size());
-  assert(bm.getAssertLevel() == symbols.size());
+  assert(bm.getAssertLevel() == frames.size());
 }
 
 void Cpp_interface::init()
@@ -48,7 +48,9 @@ void Cpp_interface::init()
   alreadyWarned = false;
 
   cache.push_back(Entry(SOLVER_UNDECIDED));
-  symbols.push_back(ASTVec());
+
+  // create an initial, empty frame
+  frames.push_back(SolverFrame(&functions));
 
   if (bm.getVectorOfAsserts().size() == 0)
     bm.Push();
@@ -63,6 +65,16 @@ Cpp_interface::Cpp_interface(STPMgr& bm_, NodeFactory* factory)
     : bm(bm_), letMgr(new LETMgr(bm.ASTUndefined)), nf(factory)
 {
   init();
+}
+
+ASTVec& Cpp_interface::getCurrentSymbols()
+{
+  return frames.back().getSymbols();
+}
+
+vector<std::string>& Cpp_interface::getCurrentFunctions()
+{
+  return frames.back().getFunctions();
 }
 
 void Cpp_interface::startup()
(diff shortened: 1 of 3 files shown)


msoos:
+1 looks good to me but @TrevorHansen may need to check as he is more familiar with this part of the code :)

lib/Interface/cpp_interface.cpp
```diff
@@ -613,4 +640,47 @@ void CNFClearMemory()
...
+//
+// This ensures that the reference counting for any symbols used in the
+// function declarations are correctly decremented.
+Cpp_interface::SolverFrame::~SolverFrame()
+{
+  // Iterate on the function names in our current scope
+  for (vector<std::string>::const_iterator scoped_function_name =
+           getFunctions().begin();
+       scoped_function_name != getFunctions().end(); ++scoped_function_name)
+  {
```
TrevorHansen, line 661: Feel free to use modern style if you'd like:
``
for (const auto& scoped_function_name : getFunctions())
``

lib/Interface/cpp_interface.cpp
```diff
@@ -613,4 +640,47 @@ void CNFClearMemory()
...
+    // Find this function in the global context
+    std::unordered_map<std::string, Function>::iterator function_to_erase =
```
TrevorHansen, line 663: You could make make this:
``
    std::unordered_map<const std::string, Function>::const_iterator function_to_erase =
``
or
```
const auto function_to_erase =
```
if you wanted to.

andrewvaughanj:
@TrevorHansen for:

> Feel free to use modern style if you'd like
>

Do you have a _preference_? Basically, I'm not a huge fan of `auto` because I feel it makes it unclear what is what.

For example, if I use an `auto` on the return of `getFunctions` and then I create an `auto` iterator on that return, without looking up the definition of `getFunctions`, I don't actually know what type you get from dereferencing the iterator.

If you don't use `auto`, you know from the immediate context what the types are. That being said, the [CVC4 guidelines *explicitly encourage*](https://github.com/CVC4/CVC4/wiki/Code-Style-Guidelines#auto) the use of `auto` for the iterator instance.

Maybe the compromise here is that iterators should use `auto` but other code is "up to the developer"?

rgov:
My usual approach is: when in doubt, steal from the [LLVM Coding Standards](https://llvm.org/docs/CodingStandards.html) or whichever major project you like. Try to match nearby code so future readers don't need to switch between styles. They have a stance on [`auto`](https://llvm.org/docs/CodingStandards.html#use-auto-type-deduction-to-make-code-more-readable).

andrewvaughanj:
Shall we set up a `CODING_GUIDELINES` document that says "if in doubt, defer to LLVM"? Basically, I feel what's in in the LLVM document aligns with what I already wrote on `auto` (e.g., it is okay to use `auto` for iterators to make the code more concise).

TrevorHansen:
I don't mean to come across like I have strong feelings about ```auto''', or coding guidelines. I was looking for something to comment on mostly so that it was obvious that I took @andrewvaughanj 's contribution seriously enough to read through it. 

If it'd make it easier for people to contribute to STP, then yes, I'm very keen on coding guidelines. Otherwise, I'm happy to let people exercise their judgement - we get so few patches through it's hard for me to imagine it all becoming chaos.

Thanks again for fixing up the function stuff!

rgov:
Haven't followed the discussion closely: @andrewvaughanj if this is ready to go, please merge.

andrewvaughanj:
> Haven't followed the discussion closely: @andrewvaughanj if this is ready to go, please merge.

I think this is good -- @TrevorHansen was happy + Travis is green.

Will merge!

PR #368, opened 2020-07-20
Disable static builds on Travis

andrewvaughanj:
Temporarily shutting off static Travis builds to at least get CI builds to be green, and then we can re-enable these.


andrewvaughanj:
I don't want this to be permanent, but having any ❌ makes it hard to know (quickly) if we're making things better or worse.

If we can get non-static builds happy (which I think we are), we can then reenable static builds and go through the process to get the builds clean.

rgov:
Seems fine. I'd love to move away from Travis to GitHub Actions. Travis is probably on life support and GitHub supports Windows builds so we don't need to separately maintain AppVeyor or whatever we're using.

andrewvaughanj:
@rgov agreed! That's high-up on my list of things I'd like to look at (otherwise, there's an element of "building on a foundation of sand").

PR #370, opened 2020-07-21
Creating a non-static CMS Travis build

andrewvaughanj:


msoos:
BTW, M4RI should work both in static & dynamic mode, so in theory you would not need to disable it :)

andrewvaughanj:
@msoos I'm just abusing Travis to work out what does/doesn't work.

Weirdly, builds don't work with `ccache` (which is the first thing I turned off) and then they build, but it couldn't find the M4RI shared library at run-time (hence why second attempt of turning off M4RI).

Just to be clear: this *is not* ready, I'm just trying to get things back in a state where both both Riss and CMS are tested in CI.

andrewvaughanj:
@msoos does this https://travis-ci.org/github/stp/stp/jobs/712276831#L2166 make any sense to you?

```
ccache: error: Failed to create temporary file for /home/travis/.ccache/c/d/2c4accbe774d6c74e3b63313130f43-209396.o.tmp.stdout: Permission denied
```

msoos:
Ah, to damn with ccache :) I'm suprised M4RI doesn't work, but to be honest, the new CMS is so damn fast at Gauss-Jordan elimination, it's not that important anyway.

andrewvaughanj:
My plan (currently) is to get CMS + Riss working under CI, and then look at replacing Travis for GH Actions.

If the tests pass, are you happy with CMS (without M4RI) in CI then?

msoos:
Sure, sounds great, thanks!

rgov:
Closing because we're not using Travis anymore.

PR #371, opened 2020-07-30
Update docs

andrewvaughanj:
Some minor readme tweaks:

- document that you need to do `git submodule init && git submodule update` before running CMake

- point out that STP has its own minisat fork (upstream minisat can't be built with a new-ish `clang`)

- provide some documentation on the "hints" you can provide to CMake if you don't want install STP's deps system-wide

### DIFF
--- a/README.markdown
+++ b/README.markdown
@@ -84,14 +84,14 @@ make(1), Visual Studio, Xcode, etc.
 Here are a few interesting configuration variables. These apply to all
 generators.
 
-- ``CMAKE_BUILD_TYPE`` - The build type (e.g. Release)
-- ``CMAKE_INSTALL_PREFIX`` - The prefix for install (e.g. /usr/local )
-- ``ENABLE_ASSERTIONS`` - If TRUE STP will be built with asserts.
-- ``ENABLE_TESTING`` - Enable running tests
-- ``ENABLE_PYTHON_INTERFACE`` - Enable building the Python interface
-- ``PYTHON_EXECUTABLE`` - Set python executable in case you have more than one python installed
-- ``SANITIZE`` - Use Clang's sanitization checks
-- ``STATICCOMPILE`` - Build static libraries and binaries instead of dynamic
+- `CMAKE_BUILD_TYPE` - The build type (e.g. Release)
+- `CMAKE_INSTALL_PREFIX` - The prefix for install (e.g. /usr/local )
+- `ENABLE_ASSERTIONS` - If TRUE STP will be built with asserts.
+- `ENABLE_TESTING` - Enable running tests
+- `ENABLE_PYTHON_INTERFACE` - Enable building the Python interface
+- `PYTHON_EXECUTABLE` - Set python executable in case you have more than one python installed
+- `SANITIZE` - Use Clang's sanitization checks
+- `STATICCOMPILE` - Build static libraries and binaries instead of dynamic
 
 ### Dependencies
 STP relies on : boost, flex, bison and minisat. You can install these by:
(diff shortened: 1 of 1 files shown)


README.markdown
```diff
@@ -111,6 +124,15 @@ $ sudo make install
 $ sudo ldconfig
 ```
 
+#### Building against non-installed libraries
+
+If you wish to build STP's dependencies without installing them, you can tell CMake where to find the non-installed artefacts; for example:
+
+* ``-DMINISAT_INCLUDE_DIRS:PATH=<path>`` and ``-DMINISAT_LIBDIR:PATH=<path>`` -- the paths to ``minisat/core/Solver.h`` and the ``minisat`` libraries (respectively)
```
rgov, line 131: In Markdown only one backtick is needed to make text monospaced.

README.markdown
```diff
@@ -99,6 +100,18 @@ STP relies on : boost, flex, bison and minisat. You can install these by:
...
 ```
 
+If your distribution does not come with minisat, STP maintains an updated fork; it can be built as follows:
+
+```
+$ git clone https://github.com/stp/minisat
+$ cd minisat
+$ mkdir build && cd build
+$ cmake ..
+$ make
```
rgov, line 110: I encourage `cmake --build .` instead of `make`, to use whichever build tool is configured.

README.markdown
```diff
@@ -99,6 +100,18 @@ STP relies on : boost, flex, bison and minisat. You can install these by:
 $ sudo apt-get install cmake bison flex libboost-all-dev python perl minisat
 ```
 
+If your distribution does not come with minisat, STP maintains an updated fork; it can be built as follows:
```
rgov, line 103: I would rather use periods instead of semicolons in both instances in this change; use whichever you feel is appropriate.

README.markdown
```diff
@@ -99,6 +100,18 @@ STP relies on : boost, flex, bison and minisat. You can install these by:
...
+
+```
+$ git clone https://github.com/stp/minisat
+$ cd minisat
+$ mkdir build && cd build
+$ cmake ..
+$ make
+$ sudo make install
+$ sudo ldconfig
```
rgov, line 112: `ldconfig` only applies to Linux, right? maybe: `command -v ldconfig && sudo ldconfig`

README.markdown
andrewvaughanj: Hahaha, yeah, I know ... weirdly, there were other instances of double-back tick in this markdown, so I copied it for consistency. I'll fix it everywhere.

README.markdown
andrewvaughanj: Easy change; I'll make it consistent through-out the whole document.

README.markdown
andrewvaughanj: I have no preference, I'm happy for `.` over `;`.

README.markdown
andrewvaughanj: I don't normally install packages like this system wide (ever), so I'm not sure (I know what `ldconfig` does, but I am not sure how essential it is here).

I can add the test if we prefer (I only don't prefer this as it might not be clear to people what `command -v <thing>` does).

rgov: How about a note after it that says, "The final command may not be necessary on some systems."

README.markdown
andrewvaughanj: Done!

README.markdown
andrewvaughanj: Done!

README.markdown
andrewvaughanj: Done!

README.markdown
andrewvaughanj: Done!

PR #378, opened 2020-08-03
Working towards using GitHub Actions for CI

andrewvaughanj:
First attempt at getting STP to build with GitHub Actions.

Some points:

- Travis is not yet disabled

- Moves towards using `contrib` style scripts (similar to Boolector and CVC4) for building STP's dependencies

- Submodules for Google Test and Output Check have been moved to `contrib` scripts

- Currently builds STP with CMS under both `gcc` and `clang`


rgov:
Is this WIP or ready to be merged?

.github/workflows/ci.yml
```diff
@@ -0,0 +1,71 @@
...
+    - uses: actions/checkout@v2
+
+    - name: Environment
+      run: |
+        echo "::set-env name=CC::${{ matrix.env.CC }}"
+        echo "::set-env name=CXX::${{ matrix.env.CXX }}"
+
+    - name: Packages
+      run: |
+        sudo mkdir -p /usr/share/man/man1
```
rgov, line 33: I'm curious why this is needed?

contrib/setup-cms.sh
```diff
@@ -0,0 +1,21 @@
+#!/usr/bin/env bash
+
+set -e -o pipefail
```
rgov, line 3: Can you also add `-u` (expansion of unset variable fails)?

contrib/setup-cms.sh
```diff
@@ -0,0 +1,21 @@
...
+
+dep_dir="deps"
+install_dir=$(readlink -fm ${dep_dir}/install)
+
+[ ! -d "${install_dir}" ] && mkdir -p "${install_dir}"
+
+dep="cms"
+
+cd ${dep_dir}
```
rgov, line 12: Please quote variable expansions. Maybe there's a shell linter we can apply to make sure it's all good.

contrib/setup-cms.sh
```diff
@@ -0,0 +1,21 @@
...
+git clone https://github.com/msoos/cryptominisat ${dep}
+cd ${dep}
+mkdir build && cd build
+cmake -G Ninja -DCMAKE_INSTALL_PREFIX:PATH=${install_dir} ..
```
rgov, line 16: Do we need to use the Ninja generator? It adds a prereq we don't need.

msoos:
I don't think we need ninja, it works without it I'm sure. I never used
ninja to be honest so I don't know what it brings but in my understanding,
cmake and make should do the trick :) Maybe others know what the benefits
are?

andrewvaughanj:
Ninja is just a “higher performance” generator that (in my opinion) does a better job of no-op builds.

It is used as the default generator for projects such as LLVM and Chrome (https://ninja-build.org). Plus, I *exclusively* build STP using Ninja (and any other project that works with it).

That being said, I’m fine to not use Ninja for this :)

msoos:
Ah then it's good! Again, I have no idea what it brings, so I'll trust you
on this :)

.github/workflows/ci.yml
andrewvaughanj: Hmm, I was testing stuff with https://github.com/nektos/act and hit some issue related to installing some dependancy, but then I changed the "base" image I was working and didn't remove this.

It can (probably) go -- I'll check!

contrib/setup-cms.sh
andrewvaughanj: 🤦 if someone had asked me, I would have been certain I had `-u`!

contrib/setup-cms.sh
andrewvaughanj: Done.

Scripts have been checked with https://github.com/koalaman/shellcheck (scripts were already formatted in a way that https://github.com/mvdan/sh made no changes).

contrib/setup-cms.sh
andrewvaughanj: So, I actually removed Ninja *from the contrib scripts* as I agree that not everyone will have it.

I **have not** removed it from the GitHub Action.

andrewvaughanj:
> Is this WIP or ready to be merged?

I think it is ready for a second round of review. Since the last time, I've:

- Removed Travis

- Updated the top-level README to tell people to use the `contrib` scripts

I think before we totally go ahead, we should make sure that Mate is happy that this action *does less* than Travis at the moment (no Coverity, no static builds, etc.).

Just so we're all clear, I have a number of follow-up bits I want to do this with:

- Add static builds

- Add builds with the other solvers (e.g., Riss)

- Migrate the Windows builds to use GitHub Actions

- Maybe we want to use https://docs.github.com/en/github/finding-security-vulnerabilities-and-errors-in-your-code/about-code-scanning over Coverity (I already have access)


Right now, this is baby steps towards that.

andrewvaughanj:
Scripts have been renamed from the "contrib" directory into the "scripts/deps" folder (and corresponding documentation updated).

msoos:
Wow, going along nicely!! Thanks for all the hard work :)

andrewvaughanj:
@msoos how does this look for you? It appears to be clean, and up to date with master!

msoos:
Sounds great to me! GitHub Actions are pretty fast and neat. I support this change for sure, thanks so much! :+1:

Hanh668 (commented):
> 
7AE44F6B-2213-42BA-8608-921CFA74B9E8

rgov:
Thank you @andrewvaughanj. This was way overdue.

There are currently test case failures but it looks like CI is working OK. I also updated the CodeQL workflow to match the latest template.

PR #379, opened 2020-08-04
Looking to enable GitHub Actions

andrewvaughanj:


andrewvaughanj:
This only existed to enable GitHub Actions on the STP repo -- work actually exists here https://github.com/stp/stp/pull/378

See https://docs.github.com/en/actions/reference/events-that-trigger-workflows#pull-request-events-for-forked-repositories for why:

> Workflows don't run on forked repositories by default. You must enable GitHub Actions in the Actions tab of the forked repository.
>

PR #387, opened 2020-10-26
Ensuring that we do not create double frames when creating a new frame (closes #385)

### LINKED ISSUE #385: Incorrect function look-up with scopes
[As reported by](https://github.com/stp/stp/commit/54416cc6f086163d1b5147e3e6db7897496c4848#commitcomment-43571894) @hailinzeng:

```
;t.smt2
(set-logic QF_ABV)
(define-fun |a| () Bool true)
(assert |a|)
(push 1)
(check-sat)
(pop 1)
```

Gives:

```
sat
Fatal Error: Trying to apply function which has not been defined.
STP Error: Trying to apply function which has not been defined.
```

Before the merge for function scopes, this does not give these errors.

andrewvaughanj:
When I implemented scopes, there was a bug where the copy constructor would get called twice when creating and then adding a scope to the vector of all scopes.

This "double creation" meant that functions were cleaned-up at the wrong time.

This PR changes it such that frames are only created once. An option here would have been to define a move constructor and use `emplace_back` (rather than frames being a vector of pointers) -- @msoos do you have a preference here?

### DIFF
--- a/include/stp/cpp_interface.h
+++ b/include/stp/cpp_interface.h
@@ -104,7 +104,7 @@ class Cpp_interface
   };
 
   // The vector of all frames that have been created by calling push
-  std::vector<SolverFrame> frames;
+  std::vector< SolverFrame* > frames;
 
   // Obtain the symbols/functions for the current frame
   ASTVec& getCurrentSymbols();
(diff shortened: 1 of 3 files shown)


msoos:
:+1: Looking great!

Issue #396, opened 2021-02-25
Unsound result with "alternative" CMS builds


TrevorHansen:
I get a different result to Boolector, and a crash with -d.

```
~/solvers/boolector/build/bin/boolector -i bad.txt 
unsat
```

```
./stp ~/bad.txt 
sat
```

```
./stp --SMTLIB2 -d ~/bad.txt
terminate called after throwing an instance of 'std::out_of_range'
  what():  vector::_M_range_check: __n (which is 1341) >= this->size() (which is 0)
Aborted (core dumped)
```

```
./stp --version
STP version 2.3.3
STP version SHA string 9a59a72e82d67cefeb88d8baa34965f70acb5d1c
STP compilation options CMAKE_CXX_COMPILER = /usr/bin/c++ | CMAKE_CXX_FLAGS =  -Wall -Wextra -Wunused -pedantic -Wsign-compare -Wtype-limits -Wuninitialized -Wno-deprecated -Wstrict-aliasing -Wpointer-arith -Wpointer-arith -Wformat-nonliteral -Winit-self -Wparentheses -Wunreachable-code | COMPILE_DEFINES =  -DNDEBUG -D__STDC_LIMIT_MACROS -DUSE_CRYPTOMINISAT | ONLY_SIMPLE = OFF | Boost_FOUND = TRUE | STATICCOMPILE = OFF | TUNE_NATIVE = OFF | COVERAGE = OFF | FORCE_CMS = OFF | LIBS = /usr/local/lib/libminisat.so | ENABLE_TESTING = OFF | ENABLE_PYTHON_INTERFACE = ON | PYTHON_EXECUTABLE = /usr/bin/python | PYTHON_LIBRARY =  | PYTHON_INCLUDE_DIRS =  |  | compilation date time = Feb  2 2021 11:51:13
c compiled with gcc version 9.3.0
```

[bad.txt](https://github.com/stp/stp/files/6039726/bad.txt)

TrevorHansen:
Reducing it down
```
(set-logic QF_BV)
(declare-fun v58395 () (_ BitVec 16))
(declare-fun v58396 () (_ BitVec 14))
(assert 
(let ((e58407 (ite (bvugt v58395 (_ bv0 16)) (_ bv1 1) (_ bv0 1)))) 
(let ((e58412 (bvmul v58395 ((_ sign_extend 15) e58407)))) 
(let ((e58499 (not (distinct (bvsrem e58412 v58395) (_ bv0 16))))) 
[...]

PR #397, opened 2021-02-28
`ToSATAIG::runSolver` now returns the value of `SATSolver::solve`

andrewvaughanj:
Issue #396 documented a case where STP was crashing when using a build
of CMS that had 'BreakID' enabled. Such a configuration of CMS can
return `l_False` for a `solve` check, but *also* return `true` on a call
to `okay`.

This led to a crash in `ConstructCounterExample` where STP expected the
underlying SAT solver to have a satisfying model, when actually it had
reported UNSAT.

This commit makes it such that `ToSATAIG::runSolver` results the value
of `SATSolver::solve` and not `SATSolver::okay`.

Signed-off-by: Andrew V. Jones <andrewvaughanj@gmail.com>

### DIFF
--- a/lib/ToSat/ToSATAIG.cpp
+++ b/lib/ToSat/ToSATAIG.cpp
@@ -202,13 +202,13 @@ void ToSATAIG::mark_variables_as_frozen(SATSolver& satSolver)
 bool ToSATAIG::runSolver(SATSolver& satSolver)
 {
   bm->GetRunTimes()->start(RunTimes::Solving);
-  satSolver.solve(bm->soft_timeout_expired);
+  bool result = satSolver.solve(bm->soft_timeout_expired);
   bm->GetRunTimes()->stop(RunTimes::Solving);
 
   if (bm->UserFlags.stats_flag)
     satSolver.printStats();
 
-  return satSolver.okay();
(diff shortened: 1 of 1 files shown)


msoos:
:+1:  Wow, thanks, looks great!

andrewvaughanj:
@msoos can you look here https://github.com/stp/stp/blob/101aee2ad9b30c38bef54883187091b90588d63c/lib/AbsRefineCounterExample/CounterExample.cpp#L1063 please?

Is it feasible for CMS (or MiniSat) to return SAT *but not be `okay`*? This PR fixes the issue where we had UNSAT and `okay`, but I don't know about the opposite.

andrewvaughanj:
Also, the only test I have takes between 70 and 80 seconds to run -- I *have not* added this a test (as, not only does it take forever, it needs BreakID).

TrevorHansen:
This looks good to me, too.

I'd guess it's impossible to be satisfiable, but not OK.

 @msoos are you good to merge?

msoos:
Sure, I confirmed above :) Sorry, I was confused, I thought we use the :+1:  as a sign of OK to merge!

PR #398, opened 2021-03-02
Treat warnings as errors

andrewvaughanj:
Enable warnings as errors (`-Werror`) while building with GCC/clang, and
fix code (and associated tests) such that STP builds across: gcc-7,
gcc-9, gcc-10, clang-7, clang-8, clang-9, clang-10, clang-11 on Linux.

Errors from external libraries (e.g., MiniSat) are marked as `SYSTEM`
(via CMake) and errors in 'internal' libraries (e.g., extlib-abc) are
suppressed via `pragmas`, which work with both GCC and clang. Selective
files in STP have explicit warnings disabled via CMake.

Treating warnings as errors is not enabled when compling with MSVC.

Signed-off-by: Andrew V. Jones <andrewvaughanj@gmail.com>

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -373,6 +373,18 @@ if (COVERAGE)
     add_cxx_flag("--coverage")
 endif()
 
+option(WERROR "Build with warnings as errors" OFF)
+if (WERROR)
+    if (NOT MSVC)
+        message(STATUS "Enabling treating warnings as errors")
+        # Let's make things be errors -- we can't test for this
+        set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Werror")
+        set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -Werror")
+    else()
+        message(STATUS "Warnings as errors is not enabled under MSVC")
+    endif()
+endif()
+
 option(USE_THREAD_LOCAL "Use thread-local storage when available" ON)
 add_feature_info(ThreadLocal USE_THREAD_LOCAL "Enables use of thread-local storage when available")
 
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -546,7 +546,7 @@ else()
     endif()
 endif()
 
-include_directories(${MINISAT_INCLUDE_DIRS})
+include_directories(SYSTEM ${MINISAT_INCLUDE_DIRS})
 set(LIBS ${LIBS} ${MINISAT_LIBRARIES})
 
 # -----------------------------------------------------------------------------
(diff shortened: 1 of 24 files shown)


andrewvaughanj:
Codacy isn't wrong; there was some code that was `if (false)` that I made into `bool print_debug = false; if (print_debug)` to silence a warning:

- [tests/api/C/b4-c2.cpp:2553](https://github.com/stp/stp/pull/398/files#diff-adb33604c7e72b7f0f727b4b4d1b4a6dbf7717e9973ad37c613328d8874bbe18R2553)

Honestly, I could just nuke the "dead" code.

Thoughts?

msoos:
Sorry to chirp in but this has hurt me really bad previously. GCC or Clang
will add another warning (or catch one more issue due to advanced
detection) 5 years down the line, someone wants to build old STP, and
things will go bottom up. Actually I have bumped into this myself sometimes
in less than a year after it built flawlessly. I have seen researchers
abandoning trying to use my tool because a warning errored out and they
understandibly decided not to track down something they thought would take
hours to fix. I strongly suggest not having -Werror as default :)

Just my 2 cents,

Mate

andrewvaughanj:
> I strongly suggest not having -Werror as default :)

Reasonable; are you happy with these changes if I put `-Werror` under a CMake option?

andrewvaughanj:
Honestly, I care significantly more about fixing (or, in some cases, silencing) warnings in the current code, such it is "painfully obvious" if I introduce a new warning when adding code.

andrewvaughanj:
💥 `-Werror` is now behind a make option `-DWERROR:BOOL=ON`.

msoos:
Thanks, perfect! And thanks so much for all your work!!

rgov:
Let's enable this in CI, at least, so that we catch anything when there's a pull request (which we don't see very often), without increasing the support burden from the build breaking on every user's distinct compiler configuration.

andrewvaughanj:
Agreed; can we get this in, then I'll do the GitHub Actions branch and enable WERROR there?

rgov:
LGTM. I'd rather delete or `#ifdef` debug code than `if (false)` but no big deal.

andrewvaughanj:
>  I'd rather delete or #ifdef debug code

I went for the `#if` approach.

Honestly, I'd rather nuke this code altogether, but I tried to make this branch do as little as possible, while getting me to `-Werror`.

msoos:
Wow, looks pretty nice! Are you planning to merge or is there some more work you are planning to do first?

andrewvaughanj:
This is good if you’re happy to merge it?

msoos:
Of course it's good! Here's my :+1:

Issue #399, opened 2021-03-10
SMT-COMP 2021


TrevorHansen:
There's new a cloud division at SMT-COMP 2021 that runs on 100 machines.

Overview:  https://smt-comp.github.io/2021/news/2021-02-05-parallel-and-cloud-tracks
Sample submission: https://github.com/aws-samples/aws-batch-comp-infrastructure-sample

The easiest seems to be to create a CNF from STP, then send it to a distributed SAT solver. 

@msoos is it easy to use CMS for distributed solving?

msoos:
Hi,

Sorry for the late reply. There is an MPI version of CryptoMiniSat -- I
haven't used it in a while, but it's not supposed to be too difficult to
make work. I'll play with it and get back to you :)

Cheers,

Mate

conp-solutions:
By now, I extended the MergeSat solver to be capable of being used as backend of the HordeSat for MPI (and the issues from last year have been addressed). Similarly to a distributed CryptoMiniSat, a distributed STP would have to translate the input problem to CNF first, before using a distributed SAT solver. Either, STP is integrated into the distributed SAT solver and each node translates the formula, or the translated CNF would have to be stored back to S3 so that each compute node could download it from there for further solving. Not sure whether distributing these CNFs via MPI broadcast would be faster.

I wonder we should submit the portfolio solve of last year to this years parallel track. Last years best parallel SAT solver only used 32 out of the 64 available cores - given virtualization and resource sharing, that makes sense. @msoos, is it sensible to combine a 28core parallel CryptoMinisat with 2 MergeSat (SAT and UNSAT configuration), 1 Riss and 1 MiniSat ?

andrewvaughanj:
> 1 MiniSat 

Do we have instances where the MiniSat configuration beats any of these other configurations?

For the cost of one thread, it feels it is probably fine to "waste" it, but I am just curious if we have any examples where vanilla MiniSat beats the other configurations.

msoos:
Hi All,

First of all, thanks for the great points!

Regarding the parallel system: sure, a multi-threaded MergeSat  + a multi-threaded CMS + a single-threaded MiniSat sounds like a good parallel setup! Re @andrewvaughanj : hmmm... not sure about MiniSat adding too much, actually, good point. @conp-solutions what do you think?

[...]

PR #404, opened 2021-05-29
Smtcomp 2021

conp-solutions:
This branch is a proposal for the parallel track of the SMT Comptition 2021 (See issue #399 )

The Dockerfile now
 * also installs MergeSat
 * uses transparent huge pages by default
 * uses the starexec portfolio configuration
 * only uses every 2nd CPU (to not run into hyperthread sharing)
 * does not support the stdin use case anymore
 * can access input from S3

A few related scripts have been adapted as well.

### Note

This change is not meant to be (fully) merged, but rather to aid the discussion for SMT comp.

### Testing Done

The change has been tested with the following sequence of commands:

```
# build verbose (to see steps), get the container ID
docker build -f Dockerfile . && STP_CONTAINER=$(docker build -q -f Dockerfile .) && echo "Container: $STP_CONTAINER"

# setup example input
export FILE_DIR=$PWD/tests/query-files/unit_test; export INPUT_FILE=bvsgt.smt2

# run the container with the input file
docker run --rm -i -v $FILE_DIR:$FILE_DIR:ro -e INPUT_FILE=$FILE_DIR/$INPUT_FILE "$STP_CONTAINER"
```

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -16,7 +16,6 @@ LICENSE
 LICENSE_COMPONENTS
 papers
 README.markdown
-scripts
 stp.kdev4
 stuff*
 tests
--- a/Dockerfile
+++ b/Dockerfile
@@ -87,6 +87,10 @@ RUN apt-get update && \
         unzip
 RUN pip install supervisor awscli
 
+# Enable transparent huge pages, always
+RUN DEBIAN_FRONTEND=noninteractive apt -y install hugepages
+RUN hugeadm --thp-always
+
 # Finally run the tool
 CMD /stp/tools/container-default-command.sh
 
--- a/Dockerfile
+++ b/Dockerfile
@@ -47,6 +47,8 @@ RUN make install
 RUN mkdir /stp
 WORKDIR /stp
 COPY . /stp
+# make all starexec tools available
+COPY scripts/starexec/bin/* /usr/local/bin/
 RUN mkdir build
 WORKDIR /stp/build
 RUN cmake -DSTATICCOMPILE=ON ..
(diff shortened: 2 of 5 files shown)


msoos:
:+1:

conp-solutions:
For the actual competition submission, where should this code land? I would suggest to create a separate branch in the main repository, or, as done for the cloud track, create a separate repository. (I would propose branch, as i only modify tooling around STP).I currently cannot check whether i have write permissions to the STP namespace. Otherwise, could somebody else create such a branch?I will rebase this branch to the current master, and would create a PR for that branch, so that we have the details for the submission.

andrewvaughanj:
@conp-solutions you're not currently on the STP "team", which means, even if I create a branch on the main STP repo, you won't be able to push to it.

The quickest way for me to resolve this now is just to create a secondary repo and give you push access (as per your second suggestion).

Would you be okay with that for now?

msoos:
Actually, I'd propose that we can of course add @conp-solutions to the STP team. Would that be okay with others?

conp-solutions:
Yes, that works.For the dedicated branch: i could at least create the pull request, and when merging without merge commit, the git hash would stay the same.Just let me know what you prefer.

conp-solutions:
This is the branch merged on top of the current origin/master. This is what I'd hand in to the competition.

TrevorHansen:
> Actually, I'd propose that we can of course add @conp-solutions to the STP team. Would that be okay with others?

Yes! If Norbert would like to.

conp-solutions:
That works for me. However, i unfortunately do not have much bandwidth to contribute other than the few bursts per year to adapt to competition configurations and related nits. If that is okay, feel free to add me.

msoos:
Hey! This has not been merged yet. I invited you @conp-solutions to the team. You can accept and then merge :)

aytey:
Closing this; it can be re-opened if we need it

PR #405, opened 2021-06-06
Cleaner CMS include and better include location

msoos:
@andrewvaughanj can you please check and merge if OK? Thanks!


andrewvaughanj:
Looks fine; what's the purpose of the change? Just to ensure that CMS doesn't "leak" into any other parts?

msoos:
Yep, it's cleaner :) Thanks!

PR #406, opened 2021-06-12
Let's not use macros instead of "const"-s

msoos:
It messed with namespaces, and in particular, messes with CryptoMiniSat's new constexpr setup. STP will not compile with the newest (MPI) cryptominisat with this inside.

Note that there **MAY** be a **SLOWDOWN**. We'd need to test. So let's not merge this before some performance testing. In my understanding, modern compilers (less than 5-7 years old) should not create any slowdown with this code. Also, it's clearly the right way to do it, separating namespaces :)

### DIFF
--- a/lib/Sat/MinisatCore.cpp
+++ b/lib/Sat/MinisatCore.cpp
@@ -23,7 +23,6 @@ THE SOFTWARE.
 ********************************************************************/
 
 #define __STDC_FORMAT_MACROS
-#define MINISAT_CONSTANTS_AS_MACROS
 #include "stp/Sat/MinisatCore.h"
 #include "minisat/core/Solver.h"
 //#include "utils/System.h"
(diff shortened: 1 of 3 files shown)


andrewvaughanj:
What does `MINISAT_CONSTANTS_AS_MACROS` change in MiniSat?

msoos:
It **only** changes these two things:

```
#if defined(MINISAT_CONSTANTS_AS_MACROS)
#define var_Undef (-1)
#else
  const Var var_Undef = -1;
#endif
```

and

```
#if defined(MINISAT_CONSTANTS_AS_MACROS)
  #define l_True  (lbool((uint8_t)0)) // gcc does not do constant propagation if these are real constants.
  #define l_False (lbool((uint8_t)1))
  #define l_Undef (lbool((uint8_t)2))
#else
  const lbool l_True ((uint8_t)0);
  const lbool l_False((uint8_t)1);
  const lbool l_Undef((uint8_t)2);
#endif
```

Obviously, `MINISAT_CONSTANTS_AS_MACROS` is very damaging to the namespace, since if any other system uses e.g. `l_True` then things will go haywire. And indeed, they do go haywire, since CryptoMiniSat uses its own `l_True` etc. The `MINISAT_CONSTANTS_AS_MACROS` is a hack, it's supposed to be faster. But this does not longer seem to be the case (I tested with CMS) with "modern" (less than 5 years old) compilers.

Issue #375, opened 2020-07-30
Evaluate moving to recent ABC


rgov:
If we can avoid having our own internal copy of ABC, we should. However, we should carefully evaluate the impact of doing so.

The copyright dates in the code that we have date back to 2007, but it's unclear when the code was imported. I looked for the first commit in which some of these files were introduced:

    aig.h -> c5277d3334e3dbca556fbf82bbe1c0cacdc85cb1 Version abc70712
    aigCheck.c -> c5277d3334e3dbca556fbf82bbe1c0cacdc85cb1 Version abc70712
    aigDfs.c -> c5277d3334e3dbca556fbf82bbe1c0cacdc85cb1 Version abc70712
    aigFanout.c -> 1647addf5e3ce4f82cc35cd4983bc5f7b9730290 Version abc70714
    aigMan.c -> c5277d3334e3dbca556fbf82bbe1c0cacdc85cb1 Version abc70712
    aigMem.c -> c5277d3334e3dbca556fbf82bbe1c0cacdc85cb1 Version abc70712
    aigMffc.c -> 64dc240b904adafee78e2a66a1fc31b717f8985f Version abc70723
    aigObj.c -> c5277d3334e3dbca556fbf82bbe1c0cacdc85cb1 Version abc70712
    aigOper.c -> c5277d3334e3dbca556fbf82bbe1c0cacdc85cb1 Version abc70712
    aigOrder.c -> 054e2cd3a8ab4ada55db4def2d6ce7d309341e07 Version abc70726
    aigPart.c -> 054e2cd3a8ab4ada55db4def2d6ce7d309341e07 Version abc70726
    aigRepr.c -> 054e2cd3a8ab4ada55db4def2d6ce7d309341e07 Version abc70726
    aigRet.c -> 9e4213e202b516c6c920d7e0faaf603273d1795d Version abc70817
    aigScl.c -> ddc6d1c1682a18e293399b7d6c9f4a9018c30c70 Version abc70828
    aigSeq.c -> c5277d3334e3dbca556fbf82bbe1c0cacdc85cb1 Version abc70712
    aigShow.c -> fefd8b901d89ad0d977db8896c12123cc747e3d7 Version abc70730
    aigTable.c -> c5277d3334e3dbca556fbf82bbe1c0cacdc85cb1 Version abc70712
    aigTime.c -> d62ee0a90d14fe762015906b6b3a5ad23421d390 Version abc70911
[...]

PR #411, opened 2021-07-29
Beautify CMake messages

Shatur:
At the moment, `stp` can be conveniently used together with [FetchContent](https://cmake.org/cmake/help/latest/module/FetchContent.html) in other projects.
But CMake output from `stp` a bit noisy. And can be confusing for other developers, because they can see warning messages like "Testing is disabled" that comes from `stp` and unrelated their projects.

In this PR I changed messages type from `STATUS` to `DEBUG` because the information like "Boost -- adding '${Boost_LIBRARY_DIRS}' to link directories" is for the `stp` developers. [The documentation](https://cmake.org/cmake/help/latest/command/message.html) says that the purpose of `DEBUG` is "Detailed informational messages intended for developers working on the project itself as opposed to users who just want to build it.". So looks like a good fit to me because usually the official modules (which are provided by CMake itself) don't print that much information.

To provide the information about what things are included, I used [FeatureSummary](https://cmake.org/cmake/help/latest/module/FeatureSummary.html).

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -47,6 +47,7 @@ set(CMAKE_CXX_STANDARD 14)
 
 include(GenerateExportHeader)
 include(GNUInstallDirs)
+include(FeatureSummary)
 
 # Search paths for custom CMake modules
 set(CMAKE_MODULE_PATH ${PROJECT_SOURCE_DIR}/cmake/modules)
@@ -63,7 +64,7 @@ set(STP_TIMESTAMPS ON CACHE BOOL "Enable build with timestamps")
 set(build_types Debug Release RelWithDebInfo MinSizeRel)
 if(NOT CMAKE_BUILD_TYPE)
 
-      message(STATUS "You can choose the type of build, options are:${build_types}")
+      message(STATUS "You can choose the type of build, options are: ${build_types}")
       set(CMAKE_BUILD_TYPE RelWithDebInfo CACHE STRING
           "Options are ${build_types}"
           FORCE
@@ -105,6 +106,7 @@ else()
     # Note this definition doesn't appear in the cache variables.
     add_definitions(-DNDEBUG)
 endif()
+add_feature_info(Assertions ENABLE_ASSERTIONS "Enables assertions")
 
 # -----------------------------------------------------------------------------
 # Enable LLVM sanitizations.
@@ -116,11 +118,11 @@ endmacro()
 
 include(MacroPushRequiredVars)
 
-option(SANITIZE "Use Clang sanitizers. This will force using clang++ as the compiler" OFF)
+option(SANITIZE "Use Clang sanitizers" OFF)
 if (SANITIZE)
     # Set in Cache so user can tweak it later
     SET(CMAKE_CXX_COMPILER "clang++" CACHE FILEPATH "" FORCE)
-    message("Forcing compiler:${CMAKE_CXX_COMPILER}")
+    message(STATUS "Forcing compiler: ${CMAKE_CXX_COMPILER}")
     add_cxx_flag("-fsanitize=return")
     add_cxx_flag("-fsanitize=bounds")
     add_cxx_flag("-fsanitize=integer")
@@ -132,6 +134,7 @@ if (SANITIZE)
     add_cxx_flag("-fsanitize=address")
     add_cxx_flag("-Wno-bitfield-constant-conversion")
 endif()
+add_feature_info(Sanitizers SANITIZE "Enables Clang sanitizers, will force using clang++ as the compiler")
 
 # -----------------------------------------------------------------------------
 # Let the user decide if they want to build shared or static client library.
@@ -144,6 +147,7 @@ if (STATICCOMPILE)
     set(BUILD_SHARED_LIBS OFF)
     set(ENABLE_PYTHON_INTERFACE OFF)
 endif()
+add_feature_info(Static STATICCOMPILE "Compiles static library and executable")
 
 if ((${CMAKE_SYSTEM_NAME} MATCHES "Windows") AND (NOT BUILD_SHARED_LIBS))
     set(Boost_USE_STATIC_LIBS ON)
@@ -370,6 +374,7 @@ if (COVERAGE)
 endif()
 
 option(USE_THREAD_LOCAL "Use thread-local storage when available" ON)
+add_feature_info(ThreadLocal USE_THREAD_LOCAL "Enables use of thread-local storage when available")
 
 if(WIN32)
   set(FLEX_PATH_HINT "e:/cygwin/bin" CACHE STRING "Flex path hints, can be null if on your path")
@@ -436,17 +441,18 @@ endif()
 
 if(NOT Boost_FOUND)
     set(ONLY_SIMPLE ON)
-    message(STATUS "Only building executable with few command-line options because the boost program_options library were not available")
 else()
     message(STATUS "Boost -- found at library: ${Boost_LIBRARIES}")
     message(STATUS "Boost -- adding '${Boost_LIBRARY_DIRS}' to link directories")
     link_directories(${Boost_LIBRARY_DIRS})
 endif()
+add_feature_info(SimpleOptions ONLY_SIMPLE "Simplifies command-line interface, only a few command-line options will be available")
 
 # -----------------------------------------------------------------------------
 # Find Riss
 # -----------------------------------------------------------------------------
 option(USE_RISS "Try to use Riss" OFF)
+add_feature_info(Riss USE_RISS "Enables Riss solver")
 
 if (USE_RISS)
     add_definitions(-DUSE_RISS)
@@ -463,7 +469,6 @@ if (NOT NOCRYPTOMINISAT)
   set(cryptominisat5_DIR "" CACHE PATH "Path to directory containing cryptominisat5Config.cmake")
   find_package(cryptominisat5 CONFIG)
   if (cryptominisat5_FOUND AND HAVE_FLAG_STD_CPP11 AND (NOT NOCRYPTOMINISAT))
-    message(STATUS "Found CryptoMiniSat 5.x and C++11, allowing --cryptominisat5 option")
     message(STATUS "CryptoMiniSat5 dynamic lib: ${CRYPTOMINISAT5_LIBRARIES}")
     message(STATUS "CryptoMiniSat5 static lib:  ${CRYPTOMINISAT5_STATIC_LIBRARIES}")
     message(STATUS "CryptoMiniSat5 static lib deps: ${CRYPTOMINISAT5_STATIC_LIBRARIES_DEPS}")
@@ -471,9 +476,9 @@ if (NOT NOCRYPTOMINISAT)
     add_definitions(-DUSE_CRYPTOMINISAT)
     set(USE_CRYPTOMINISAT ON)
   else()
-    message(WARNING "CryptoMiniSat5 (5.0 or above) or C++11 support not found, not allowing --cryptominisat option")
   endif()
 endif()
+add_feature_info(CryptoMiniSat USE_CRYPTOMINISAT "Enables CryptoMiniSat solver, allows --cryptominisat5 option")
 
 option(FORCE_CMS "Must build with CryptoMiniSat" OFF)
 if (FORCE_CMS AND NOT USE_CRYPTOMINISAT)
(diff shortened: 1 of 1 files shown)


andrewvaughanj:
Should `FeatureSummary` list all of the options? For example, including the "SUPPRESS_WARNINGS" you recently added?

In fact, should `add_feature_info` be called for _all_ options?

Shatur:
> Should FeatureSummary list all of the options?

As you wish. I see `FeatureSummary` as a list of options that affect features in an application. But if you want, I can add descriptions for all options as well.

andrewvaughanj:
> In this PR I changed messages type from STATUS to DEBUG because the information like "Boost -- adding '${Boost_LIBRARY_DIRS}' to link directories" is for the stp developers.
>

I suggest doing something like (not tested, written in the Github editor):

```
option(DEBUG_MESSAGES "enable debug messages from CMake" ON)

if (DEBUG_MESSAGES)
   set(DEBUG_STREAM STATUS)
else()
   set(DEBUG_STREAM DEBUG)
endif()

message(${DEBUG_STREAM} "You can choose the type of build, options are: ${build_types}")
```

So this leaves it "as-is" today but allows a "knowledgeable user" to silence this to the debug stream if necessary.

andrewvaughanj:
> I see FeatureSummary as a list of options that affect features in an application.

Sure, I can agree with that.

I guess then we need:

- COVERAGE
- ENABLE_ASSERTIONS
- NOCRYPTOMINISAT
- ONLY_SIMPLE
- SANITIZE
- STATICCOMPILE
- USE_RISS
- USE_THREAD_LOCAL

Does that seem like a reasonable enough list to you?

Shatur:
> I suggest doing something like (not tested, written in the Github editor):

Okay, will do!

> Does that seem like a reasonable enough list to you?

Hm... As to me `COVERAGE`, `ENABLE_ASSERTIONS` and `SANITIZE` not directly related to the features of the library. But will add, not problem :)

andrewvaughanj:
>  As to me COVERAGE, ENABLE_ASSERTIONS and SANITIZE not directly related to the features of the library.

Hmm, why do you think that? The behaviour of a build that enables `ENABLE_ASSERTIONS` could be "observably different" from a build that doesn't enable. Same for `SANITIZE` or `COVERAGE` -- linking a library with these could be different to linking against a library without them.

Shatur:
> Hmm, why do you think that?

Just because this features mostly for developing, not like a regular feature for user.

andrewvaughanj:
> Just because this features mostly for developing, not like a regular feature for user.

Nah, I don't agree with that -- a build without `ENABLE_ASSERTIONS` might be more tolerant of runtime issues *caused* by a user.

If you're happy, I think the above list a reasonable starting point.

Shatur:
> If you're happy, I think the above list a reasonable starting point.

OK, I do not mind :)

> option(DEBUG_MESSAGES "enable debug messages from CMake" ON)

How about to use [CMAKE_MESSAGE_LOG_LEVEL](https://cmake.org/cmake/help/latest/variable/CMAKE_MESSAGE_LOG_LEVEL.html#variable:CMAKE_MESSAGE_LOG_LEVEL) instead? Personally, it seems to me that it is better to set this parameter from the command line / IDE settings, but we can change the default log level if you want.

andrewvaughanj:
> How about to use CMAKE_MESSAGE_LOG_LEVEL instead? Personally, it seems to me that it is better to set this parameter from the command line / IDE settings, but we can change the default log level if you want.

https://cmake.org/cmake/help/latest/variable/CMAKE_MESSAGE_LOG_LEVEL.html#variable:CMAKE_MESSAGE_LOG_LEVEL

> New in version 3.17.

Could be too new 😞 

I think [@msoos's comment](https://github.com/stp/stp/issues/375#issuecomment-888237198) on 3.11 being fine is reasonable; I am not sure about 3.17 (I certainly don't want to bump to 3.17 *just* for `message`s 😂)

This raises a good point ... do we need to bump the CMake version for `FeatureSummary`?

Shatur:
> I think @msoos's comment on 3.11 being fine is reasonable; I am not sure about 3.17 (I certainly don't want to bump to 3.17 just for messages joy)

:smile: 
But everything should work on the old version, just not all messages will be shown until the user specifies `--log-level`.
**But** `DEBUG` was added in 3.15. So it's up to you if you want to bump it to 3.15. For Debian / Ubuntu CMake 3.15+ is available.

> This raises a good point ... do we need to bump the CMake version for FeatureSummary?

Bumping is not necessary, It's available in 3.3.

andrewvaughanj:
How much do you care about changing these messages? I don't want to break some of the upstream projects that use STP (our main user is KLEE, which is [currently at 3.5](https://github.com/klee/klee/blob/df04aeadefb4e1c34c7ef8b9123947ff045a34d9/CMakeLists.txt#L13) 🙀 )

Break is a strong term there: I want STP to work everywhere that KLEE works; if there are places where KLEE works and STP doesn't (because of CMake) that's a really dumb choice by us (== the STP project).

Shatur:
> How much do you care about changing these messages? I don't want to break some of the upstream projects that use STP (our main user is KLEE, which is currently at 3.5 scream_cat )

Oh, got it. So even 3.11 is unavailable?
Hm... Okay, probably we could left only `FeatureSummary`. I just found myself a lot of messages from `STP` annoying. Libraries usually don't print a lot of debug information, such as:

```
MESSAGE(STATUS "PROJECT_VERSION: ${PROJECT_VERSION}")
MESSAGE(STATUS "PROJECT_VERSION_MAJOR: ${PROJECT_VERSION_MAJOR}")
MESSAGE(STATUS "PROJECT_VERSION_MINOR: ${PROJECT_VERSION_MINOR}")
MESSAGE(STATUS "PROJECT_VERSION_PATCH: ${PROJECT_VERSION_PATCH}")
```

Or

```
message(DEBUG "You can choose the type of build, options are: ${build_types}")
```

`build_types` is a hardcoded list of default CMake options. Also this options do not have any effect for multi-configuration generators (MSVC or Ninja in multi-configuration mode).

I can leave with this. But can I suggest to use `STATUS` if cryptominisat is disabled instead of `WARNING` (for example, `FeatureSummary` also use regular for disabled optional components)? Also it would be nice to not print anything if testing is disabled.

andrewvaughanj:
How about:

```
if (SUPPRESS_WARNINGS)
   set(WARNING_STREAM STATUS)
else()
   set(WARNING_STREAM WARNING)
endif()

message(${WARNING_STREAM} "<cms message>")
```

?

Shatur:
> How about:

Are you suggesting to use the same `SUPPRESS_WARNINGS` variable as for compiler warnings?
The solution looks hacky to me. Is anyone really need the current behavior? I would just use regular messages as in `FeatureSummary` and remove some useless debug information like mentioned above.

andrewvaughanj:
> The solution looks hacky to me. Is anyone really need the current behavior?

Yes, the developers of STP (e.g., me).

I would actually argue that your approach of looking in the cache directly is significantly more advanced.

Shatur:
>Yes, the developers of STP (e.g., me).

Okay.
Overriding message type with another option will be too hacky. Let's leave the current behavior... I just tried to suggest how it usually done in other libraries. Yes, I understand why you can't bump the required CMake version, but right now I'm talking about not using `WARNING` for the library features. Then `FeatureSummary` will not fit well too.

> I would actually argue that your approach of looking in the cache directly is significantly more advanced.

Sure, but other two suggested approaches should be fine. And you almost never need this information. Especially printing stuff like `PROJECT_VERSION` or build types.

andrewvaughanj:
Does swapping `WARNING` for `STATUS` help you though? Because I guess I'm fine with that.

Just so you know: I'm not trying to be difficult; I'm trying to find the middle-ground between sticking with an older CMake, not changing the interface the current STP development team use *and* making life better for you -- not always easy :)

Shatur:
> Does swapping WARNING for STATUS help you though? Because I guess I'm fine with that.

Yes! This is almost exactly what I suggesting, sorry If I confuse you. Will do it and update `FeatureSummary` list.

> not always easy

Oh, totally understand :smile:

andrewvaughanj:
I think we got distracted.

Okay, so:

- `WARNING` -> `STATUS`
- Add `FeatureSummary`

🚀

andrewvaughanj:
It is probably "abusive", but how about those warnings going into `FeatureSummary` (and then the messages can go away all together)?

Shatur:
> how about those warnings going into FeatureSummary

Do I understand correctly that you suggest to use feature summary instead of "found / not found" messages?

For example, use only this:

```cmake
add_feature_info(CryptoMiniSat USE_CRYPTOMINISAT "Enables CryptoMiniSat solver, allows --cryptominisat5 option")
```

and remove the following "found" messages:

```cmake
message(STATUS "Found CryptoMiniSat 5.x and C++11, allowing --cryptominisat5 option")
message(STATUS "CryptoMiniSat5 (5.0 or above) or C++11 support not found, not allowing --cryptominisat option")
```

while keep other information if the library found, like

```
message(STATUS "CryptoMiniSat5 dynamic lib: ${CRYPTOMINISAT5_LIBRARIES}")
message(STATUS "CryptoMiniSat5 static lib:  ${CRYPTOMINISAT5_STATIC_LIBRARIES}")
message(STATUS "CryptoMiniSat5 static lib deps: ${CRYPTOMINISAT5_STATIC_LIBRARIES_DEPS}")
message(STATUS "CryptoMiniSat5 include dirs: ${CRYPTOMINISAT5_INCLUDE_DIRS}")
```

andrewvaughanj:
To me, that seems like it has some elegance and "improves" things (and hopefully makes life better for you). What do you think?

I don't think we can solve the `STATUS` ones that you listed at the bottom, but small incremental steps are always good ...

Shatur:
> To me, that seems like it has some elegance and "improves" things (and hopefully makes life better for you). What do you think?

Agree, will look much nicer, will do!

andrewvaughanj:
And so we'll do it for stuff that isn't "just" features? Like SANTIZE and TESTING? This was the "abusive" part.

Shatur:
> And so we'll do it for stuff that isn't "just" features? Like SANTIZE and TESTING? This was the "abusive" part.

Sure, going to add a summary for every item in the following list:

 * COVERAGE
 * ENABLE_ASSERTIONS
 * NOCRYPTOMINISAT
 * ONLY_SIMPLE
 * SANITIZE
 * STATICCOMPILE
 * USE_RISS
 * USE_THREAD_LOCAL

andrewvaughanj:
Won't things like `PYTHON_EXECUTABLE` cause you problems though?

```
message(WARNING "Cannot find python interpreter, cannot build python interface")
```

Same for `ENABLE_TESTING`:

```
message(WARNING "Testing is disabled")
```

Shatur:
> Won't things like PYTHON_EXECUTABLE cause you problems though?

Going to replace the following:

```cmake
if ((NOT DEFINED ENABLE_PYTHON_INTERFACE) OR ENABLE_PYTHON_INTERFACE)
    find_package (PythonInterp)
    if (PYTHON_EXECUTABLE)
        set(PYTHON_OK ON)
        message(STATUS "OK, found python interpreter")
    else()
        message(WARNING "Cannot find python interpreter, cannot build python interface")
    endif()

    if (PYTHON_OK AND BUILD_SHARED_LIBS)
        message(STATUS "Python found, enabling python interface")
        set(ENABLE_PYTHON_INTERFACE ON)
    endif()
endif()
```

with

```cmake
if((NOT DEFINED ENABLE_PYTHON_INTERFACE) OR ENABLE_PYTHON_INTERFACE)
    find_package(PythonInterp)
    if(PYTHON_EXECUTABLE AND BUILD_SHARED_LIBS)
        set(ENABLE_PYTHON_INTERFACE ON)
    endif()
endif()
add_feature_info(PythonInterface ENABLE_PYTHON_INTERFACE "Enables Python interface")
```

What do you think?

> Same for ENABLE_TESTING:

Similar here, going to represent it as a build "feature". Is it okay?

Shatur:
> Don't we need an OFF if PYTHON_EXECUTABLE isn't found?

Good catch!

andrewvaughanj:
Anyway, what you've shared is correct ... as long as the logic is correct :p

andrewvaughanj:
I need to play with this, but it seems like the changes look good.

Do you they achieve what you hoped to achieve? Do they solve the "why are these messages WARNINGS?" for you?

CMakeLists.txt
```diff
@@ -105,6 +106,7 @@ else()
     # Note this definition doesn't appear in the cache variables.
     add_definitions(-DNDEBUG)
 endif()
+add_feature_info(Assertions ENABLE_ASSERTIONS "Enables assertions")
```
andrewvaughanj, line 109: I just tried this; I think these messages need to *not* start with an upper-case, otherwise it reads weirdly with a capital letter in the middle.

andrewvaughanj (changes requested):
Tried it; works great!

One minor comment (well, a minor comment that applies to all of `add_feature_info` calls).

Shatur:
> Do you they achieve what you hoped to achieve? Do they solve the "why are these messages WARNINGS?" for you?

Yes, warnings were the main reason if this PR!

CMakeLists.txt
Shatur: I also thought about it. But in the [official example](https://cmake.org/cmake/help/latest/module/FeatureSummary.html) they use upper-case, do not know why. But we can use lower-case. What do you think?

andrewvaughanj: Well, now you've linked me to that, maybe we should just leave it. If you care, we can. If you don't, I'm not _that_ worried.

Shatur: Just checked [other projects](https://github.com/search?p=3&q=add_feature_info&type=Code), sometimes people start the description with lower-case, sometimes with upper-case. So I do not know how to make it right. I personally use upper-case as in the official examples. But it's up to you to decide.

andrewvaughanj: Leave it; it really is unclear which way is "best". I might tidy them up later on to work better with leading lower-case -- right now, they're just sentence fragments so it doesn't matter.

andrewvaughanj:
@Shatur are we good? Happy for me to merge?

Shatur:
@andrewvaughanj yes, totally!

andrewvaughanj (approved):

andrewvaughanj:
🚀

andrewvaughanj:
Thanks, @Shatur!

Shatur:
Thank you too!

Shatur:
@andrewvaughanj, I realized that to use `stp` as a library in `FetchContent` we need another option that disables executable compilation. Sorry for bothering so often. Opened another PR: #412.

PR #412, opened 2021-07-30
Add BUILD_EXECUTABLES option

Shatur:
Adds an option that disables the build of executable files. This is useful if `stp` is being used as a library. Otherwise, when using the `install` rule, the `stp` application is installed with the product.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -658,9 +658,12 @@ add_feature_info(PythonInterface ENABLE_PYTHON_INTERFACE "Enables Python interfa
 # -----------------------------------------------------------------------------
 
 message(STATUS "Final CXX flags ${CMAKE_CXX_FLAGS}")
+
+option(BUILD_EXECUTABLES "Build executables" ON)
 add_subdirectory(lib)
 add_subdirectory(tools)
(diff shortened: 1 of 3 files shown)


andrewvaughanj:
Would something like:

- [https://stackoverflow.com/a/14911387](https://stackoverflow.com/a/14911387)

be nicer here?

Soliciting opinions rather than saying it must be done one way or another.

Shatur:
> be nicer here?

This will fail the install rule. You will have the following error if you exclude a target from ALL rule:

```
CMake Error at tools/stp/cmake_install.cmake:52 (file):
  file INSTALL cannot find "/mnt/Files/Git/stp/build/stp": No such file or
  directory.
Call Stack (most recent call first):
  tools/cmake_install.cmake:47 (include)
  cmake_install.cmake:74 (include)
  ```

andrewvaughanj:
Guess not then; PR looks fine -- okay to merge?

Shatur:
> Guess not then; PR looks fine -- okay to merge?

Yep!

andrewvaughanj (approved):

andrewvaughanj:
Thanks, @Shatur!

PR #415, opened 2021-11-06
Fix bvExtract interface comment

Shangyint:
In the C interface, the comment for `vc_bvExtract(vc, e, high, low)` says the bit-width is `high - low`. However, as this operation is inclusive for both `high` and `low`, the resulting bit-width should be `high - low +1`


andrewvaughanj:
This change looks correct, but can you add parens around it? We want `(high_bit_no - low_bit_no) + 1` not `high_bit_no - low_bit_no + 1` (which is ambiguous).

Shangyint:
Thanks, fixed it.
Also spotted a error with the comments for left/right shift.

rgov:
@andrewvaughanj Can you please re-review and make a decision on merging?

rgov:
Thank you for your contribution!

PR #418, opened 2022-01-12
Modernize Boost usage

Shatur:
In this PR I modernized Boost usage by using [imported targets](https://cmake.org/cmake/help/latest/module/FindBoost.html#imported-targets). But the feature was introduced in CMake 3.5, so I bumped the version. But KLEE uses [the same version](https://github.com/klee/klee/blob/0ba95edbad26fe70c8132f0731778d94f9609874/CMakeLists.txt#L13), so I hope it's okay :)

Made the same PR for [cryptominisat](https://github.com/msoos/cryptominisat/pull/666).

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -18,7 +18,7 @@
 # OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
 # THE SOFTWARE.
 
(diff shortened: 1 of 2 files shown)


Shatur:
Hm... It looks like CI is broken. The error is unrelated and latest runs have the same issue.

rgov:
Looks like this was approved in CryptoMiniSat so I'll assume it's good here too. Thank you for contributing!

PR #420, opened 2022-02-28
Remove parallel flag because does not work on ctest 3.5.1.

TrevorHansen:
The --parallel flag doesn't seems to work on ctest 3.5.1.  All our build scripts work on 3.5.1 so it seems a mistake. Building gtest seems fast so I don't think we need parallel on it anyway.


rgov:
Seems to have been added in 3.12. https://cmake.org/cmake/help/latest/release/3.12.html#command-line

PR #427, opened 2022-05-30
Fixed build error due to namespace resolution

ennamarie19:


msoos:
Hi,

I think MiniSat can be build to make sure that the `l_False` etc are not `'#define`'s but instead are actual functions. Then this error doesn't exist, and `#define`'s don't pollute the namespace -- a win-win! Have you checked this build option for MiniSat? I think that should fix your build issue.

rgov:
Closing due to inactivity. Please leave a comment responding to the previous discussion if you'd like it reopened.

Issue #430, opened 2022-06-13
STP fails to build with gcc 7


ccadar:
Hi, we noticed in KLEE that the latest STP fails to build with gcc 7 now.  I attach (the start of) the error message below
What platform(s) are you using in the CI? 

Thanks!

```
[ 21%] Building CXX object lib/Simplifier/CMakeFiles/simplifier.dir/StrengthReduction.cpp.o
In file included from /usr/include/c++/7/bits/hashtable.h:35:0,
                 from /usr/include/c++/7/unordered_map:47,
                 from /data/stp/stp/include/stp/AST/UsefulDefs.h:49,
                 from /data/stp/stp/include/stp/AST/ASTInternal.h:28,
                 from /data/stp/stp/include/stp/AST/ASTNode.h:29,
                 from /data/stp/stp/include/stp/AST/AST.h:28,
                 from /data/stp/stp/include/stp/Simplifier/SubstitutionMap.h:28,
                 from /data/stp/stp/include/stp/Simplifier/Simplifier.h:28,
                 from /data/stp/stp/lib/Simplifier/StrengthReduction.cpp:25:
/usr/include/c++/7/bits/hashtable_policy.h: In instantiation of ‘struct std::__detail::__is_noexcept_hash<const stp::ASTNode, std::hash<const stp::ASTNode> >’:
/usr/include/c++/7/type_traits:143:12:   required from ‘struct std::__and_<std::__is_fast_hash<std::hash<const stp::ASTNode> >, std::__detail::__is_noexcept_hash<const stp::ASTNode, std::hash<const stp::ASTNode> > >’
/usr/include/c++/7/type_traits:154:31:   required from ‘struct std::__not_<std::__and_<std::__is_fast_hash<std::hash<const stp::ASTNode> >, std::__detail::__is_noexcept_hash<const stp::ASTNode, std::hash<const stp::ASTNode> > > >’
/usr/include/c++/7/bits/unordered_map.h:103:66:   required from ‘class std::unordered_map<const stp::ASTNode, stp::UnsignedInterval*>’
/data/stp/stp/lib/Simplifier/StrengthReduction.cpp:408:62:   required from here
/usr/include/c++/7/bits/hashtable_policy.h:87:34: error: no match for call to ‘(const std::hash<const stp::ASTNode>) (const stp::ASTNode&)’
  noexcept(declval<const _Hash&>()(declval<const _Key&>()))>
[...]

PR #431, opened 2022-06-13
Remove the unnecessary use of a SCARY iterator that may break on older compilers

danielschemmel:
Older compilers (see #430) currently fail building STP master due to the unnecessary use of a SCARY iterator. This PR simply replaces the written out type with an `auto` use, which will deduce the correct iterator type.


msoos:
Danke sehr as the German would say :) Thanks so much! I'd prefer the `(auto it: visited)` but I understand that may actually increase the required C++ version? We should move there IMHO, maybe one day I'll sit down and do it for fun :)

PR #438, opened 2022-07-27
CMakeLists: use absolute libdir in rpath handling

### LINKED ISSUE #437: rpath doesn't work correctly with relative CMAKE_INSTALL_LIBDIR
openSUSE build system defines:
```
 -DCMAKE_INSTALL_PREFIX:PATH=%{_prefix}
 -DCMAKE_INSTALL_LIBDIR:PATH=%{_lib}
```
where:
```
$ rpm -E '%{_prefix} XX %{_lib}'
/usr XX lib64
```
I.e. `CMAKE_INSTALL_LIBDIR` is relative. Which is perfectly fine -- in fact https://cmake.org/cmake/help/latest/module/GNUInstallDirs.html says:

> It should typically be a path **relative** to the installation prefix so that it can be converted to an absolute path in a relocatable way (see CMAKE_INSTALL_FULL_&lt;dir&gt;). However, an absolute path is also allowed.

So this test doesn't work properly:
https://github.com/stp/stp/blob/e8d153f196dcdf609add70e867f2188d58ffcc97/CMakeLists.txt#L191-L194

and rpath is set to "lib64". If you used `GNUInstallDirs`, you could use `CMAKE_INSTALL_FULL_LIBDIR` in the test above. I don't know how to solve it without `GNUInstallDirs`.

BTW. minisat has exactly the same problem:
https://github.com/stp/minisat/blob/37158a35c62d448b3feccfa83006266e12e5acb7/CMakeLists.txt#L86

jirislaby:
Commit a26083b634fe (Fixing libdirs for CMakeLists) switched from
`${CMAKE_INSTALL_PREFIX}/lib` to `${CMAKE_INSTALL_LIBDIR}`. It fixed the
issue with `lib` vs `lib64`. But while the former is absolute, the latter
needs not. That causes troubles as rpath can be set to something like
`lib64` only. That is bogus.

So fix this by using absolute `${CMAKE_INSTALL_FULL_LIBDIR}` from
`GNUInstallDirs` which is already included and handles the paths correctly.

Fixes #437.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -181,16 +181,16 @@ else ()
     # (but later on when installing)
     SET(CMAKE_BUILD_WITH_INSTALL_RPATH FALSE)
 
-    SET(CMAKE_INSTALL_RPATH "${CMAKE_INSTALL_LIBDIR}")
+    SET(CMAKE_INSTALL_RPATH "${CMAKE_INSTALL_FULL_LIBDIR}")
 
     # add the automatically determined parts of the RPATH
     # which point to directories outside the build tree to the install RPATH
     SET(CMAKE_INSTALL_RPATH_USE_LINK_PATH TRUE)
 
     # the RPATH to be used when installing, but only if it's not a system directory
-    LIST(FIND CMAKE_PLATFORM_IMPLICIT_LINK_DIRECTORIES "${CMAKE_INSTALL_LIBDIR}" isSystemDir)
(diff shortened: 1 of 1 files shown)


rgov:
@aytey Are you happy with this change following the discussion in #437? If so please merge.

TrevorHansen:
Thanks, merged in.

PR #439, opened 2022-08-03
Rewrite Dockerfile

rgov:
I rewrote the Dockerfile to address a few perceived deficiencies, though possibly the right thing is to do (in the future) is to use the `./scripts/deps/` scripts so that the logic is not duplicated.

1. I upgraded the builder image to Ubuntu 22.04 from Ubuntu 16.04.
2. I removed the `LABELS` since these were copied over from CryptoMiniSat and not appropriate for STP.
3. I moved away from building M4RI to installing it from the repository, as CMS instructs in its README.
4. I restructured the `apt-get install` to follow recommended Dockerfile best practices, e.g., with each dependency on its own line, sorted alphabetically, for cleaner diffs. I also reduced the number of times we have to run `apt-get update`.
5. I removed the use of experimental compiler toolchains (`ppa:ubuntu-toolchain-r/test`).
6. I switched from `make && make install` to using `cmake --build` and `cmake --install`.
7. I reduced the number of separate layers that are created during a build, combining some and removing others that are redundant (no effect on the final image).
8. I switched the produced image from `alpine:latest` to a completely blank image (since we are statically linked) which reduces the final image size from 10 MB to 5 MB.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,4 +1,4 @@
-# AUTHORS: Dan Liew, Ryan Gvostes, Mate Soos
+# AUTHORS: Dan Liew, Ryan Govostes, Mate Soos
 #
 # Permission is hereby granted, free of charge, to any person obtaining a copy
 # of this software and associated documentation files (the "Software"), to deal
@@ -152,7 +152,7 @@ if (STATICCOMPILE)
 endif()
 add_feature_info(Static STATICCOMPILE "Compiles static library and executable")
 
-if ((${CMAKE_SYSTEM_NAME} MATCHES "Windows") AND (NOT BUILD_SHARED_LIBS))
+if ((${CMAKE_SYSTEM_NAME} MATCHES "Windows") OR (NOT BUILD_SHARED_LIBS))
     set(Boost_USE_STATIC_LIBS ON)
     set(Minisat_USE_STATIC_LIBS ON)
 endif()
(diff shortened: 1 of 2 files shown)


CMakeLists.txt
```diff
@@ -152,7 +152,7 @@ if (STATICCOMPILE)
 endif()
 add_feature_info(Static STATICCOMPILE "Compiles static library and executable")
 
-if ((${CMAKE_SYSTEM_NAME} MATCHES "Windows") AND (NOT BUILD_SHARED_LIBS))
+if ((${CMAKE_SYSTEM_NAME} MATCHES "Windows") OR (NOT BUILD_SHARED_LIBS))
```
rgov, line 155: I'm not sure about the Windows logic here, but we definitely need to set `Boost_USE_STATIC_LIBS` whenever we are doing a static compilation, even on Linux, or else the build will fail.

rgov:
CI failures seem unrelated, they've been failing for a while?

Tested on `linux/arm64` and `linux/amd64` (under emulation).

Dockerfile
```diff
@@ -1,68 +1,61 @@
...
-WORKDIR /cms/build
-RUN cmake -DENABLE_ASSERTIONS=OFF -DCMAKE_BUILD_TYPE=Release -DSTATICCOMPILE=ON ..
-RUN make -j6
-RUN make install
-
-RUN mkdir /minisat
+RUN wget -O cryptominisat.tgz https://github.com/msoos/cryptominisat/archive/5.8.0.tar.gz \
+ && tar xvf cryptominisat.tgz --strip-components 1 \
+ && mkdir build && cd build \
+ && cmake .. \
```
TrevorHansen, line 31: @msoos do we need to add -DREQUIRE_M4RI=ON as well?

TrevorHansen (commented):
This looks great to me. Thanks for sorting this out!

PR #450, opened 2022-10-18
Use sysconfig, not distutils, to find python root

jamesjer:
The distutils module was deprecated in python 3.10, and will not be installed by default starting in python 3.12.  See https://peps.python.org/pep-0632/ for more information.  This PR uses sysconfig instead of distutils to find the python library directory.


numinit:
+1 for this patch, we are pulling it in: https://github.com/NixOS/nixpkgs/pull/327037

rgov:
Seems reasonable. This appears to be backwards compatible to Python 3.2.

PR #474, opened 2023-12-07
Remove nonsensical sbrk usage

brooksdavis:
This code called sbrk(4000000) and beyond error checking ignored the result meaning the 4MB of address space allocated could never be used without another call to sbrk and reference to INITIAL_MEMORY_PREALLOCATION_SIZE, but nothing in the repo does either.


TrevorHansen:
Thanks for investigating, @brooksdavis . I never understood what this does, but was too afraid to remove it.

PR #482, opened 2024-04-23
Install cadiback dependency

### LINKED ISSUE #481: CryptoMiniSat build script needs updating
CMS has added new dependencies. Their README now has the following installation script:

```sh
sudo apt-get install build-essential cmake

# not required but very useful
sudo apt-get install zlib1g-dev

git clone https://github.com/meelgroup/cadical
cd cadical
git checkout mate-only-libraries-1.8.0
./configure
make
cd ..

git clone https://github.com/meelgroup/cadiback
cd cadiback
git checkout mate
./configure
make
cd ..

tar xzvf cryptominisat-version.tar.gz
cd cryptominisat-version
mkdir build && cd build
cmake ..
make
sudo make install
sudo ldconfig
```

Specifically, it seems like the CadiBack dependency is missing.

Seems like there's two ways to fix:
- Build the CadiBack dependency in the following script: https://github.com/stp/stp/blob/master/scripts/deps/setup-cms.sh
- Allow the user to specify a specific CMS commit to be built when calling `setup-cms.sh`, which would allow us to use an old version pre-CadiBack dependency

Either way, should be a quick fix!

gussmith23:
Feel free to make whatever edits you want to this -- was my quick solution to the problem.

Closes #481.


gussmith23:
@aytey would love help on this. Unsure why CI is failing. This was the minimal set of changes to get it to work on my side, but I'm sure there's more that's needed!

aytey:
> @aytey would love help on this. Unsure why CI is failing. This was the minimal set of changes to get it to work on my side, but I'm sure there's more that's needed!

I'll try to look on Friday -- sorry.

As a note, I'm not sure if appveyor has been happy "in quite a while", so it failing might be something to ignore.

aytey:
I just approved the CI for gcc/clang -- sorry for not doing that sooner.

gussmith23:
Unsure why clang passed and gcc failed. I expected both to fail. gcc-built version getting errors like this:
```
/home/runner/work/stp/stp/build/tests/unit-tests/MergeSame_TestTests: error while loading shared libraries: libcadiback.so: cannot open shared object file: No such file or directory
```

This is because deps/cadiback/libcadiback.so and deps/cadical/build/libcadical.so need to be installed alongside STP. Currently, I don't think that's happening. I'm not quite sure what the right way is to make that happen with CMake. I'm assuming it's a simple fix!

TrevorHansen:
> Unsure why clang passed and gcc failed. I expected both to fail. gcc-built version getting errors like this:
> 
> ```
> /home/runner/work/stp/stp/build/tests/unit-tests/MergeSame_TestTests: error while loading shared libraries: libcadiback.so: cannot open shared object file: No such file or directory
> ```
> 
> This is because deps/cadiback/libcadiback.so and deps/cadical/build/libcadical.so need to be installed alongside STP. Currently, I don't think that's happening. I'm not quite sure what the right way is to make that happen with CMake. I'm assuming it's a simple fix!

Thanks for the fix. It's weird to me too, that the gcc-built version is failing. I don't know why either.

PR #488, opened 2024-06-30
Make timestamp in generated file use $SOURCE_DATE_EPOCH

keszybz:
Currently the generator puts the time of the build in the file, making the results different in each rebuild. Use $SOURCE_DATE_EPOCH to set the timestamp, if the variable is defined.

See https://reproducible-builds.org/docs/source-date-epoch/ for background and the definition of the variable.

The code modification is based on
https://reproducible-builds.org/docs/source-date-epoch/#perl.

Also, fix typo in the header.

### DIFF
--- a/lib/AST/genkinds.pl
+++ b/lib/AST/genkinds.pl
@@ -38,7 +38,7 @@ $maxkids = 0;
 %cat_index = ();
 
 if ($timestamp) {
-  $now = " " . localtime time;
+  $now = " " . localtime ($ENV{SOURCE_DATE_EPOCH} || time)
 } else {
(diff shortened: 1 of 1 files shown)


rgov:
Seems reasonable. Thank you for the contribution. Do any build systems set this environment variable automatically?

keszybz:
Yes, pretty much all build systems that try to build reproducibly will set it. At this point, I think all the major distros have it set automatically (Redora/RHEL/CentosStream/derivatives, OpenSUSE, Debian, Ubuntu, Arch, NixOs, etc.)

PR #495, opened 2024-09-12
Added dummy diagnostic-output-channel support

nindanaoto:
Hello, maintainers,

 I added the dummy (not changing any solver behavior) interface for the `diagnostic-output-channel` option support. 
 This is used in PySMT as a minimal requirement for an SMT solver, and this is why I added this.
  https://github.com/pysmt/pysmt/blob/71c35b0b82f7ea63240218777940bed06a18c60c/pysmt/smtlib/solver.py#L47
  
  Because this is a small, harmless change, I hope you will merge this.

### DIFF
--- a/lib/Interface/cpp_interface.cpp
+++ b/lib/Interface/cpp_interface.cpp
@@ -598,6 +598,13 @@ void Cpp_interface::setOption(std::string option, std::string value)
     else
       unsupported();
   }
+  else if (option == "diagnostic-output-channel")
+  {
+    if (value == "stdout")
+      success();
+    else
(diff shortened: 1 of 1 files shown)


aytey:
Is it worth adding a test here to validate that this works correctly?

nindanaoto:
Thank you for the response.
As far as I understood, currently there are no test for option setting related tests and since this PR adds almost trivial code, I don't agree with that we should add the test for this change. 
However, if you feel STP shuld include a test for all currently supported option related interfaces, I agree with that the interface I added should be included in that.

PR #376, opened 2020-07-31
Initial work to get STP to use the latest ABC from Git (current ABC rev: 5c8ee4a2)

aytey:
(the minute I open this PR, I am marking it as a draft)

This is PR to do the initial work to move from the version of ABC that STP includes in the repo to the latest version from https://github.com/berkeley-abc/abc (I did this with https://github.com/berkeley-abc/abc/commit/5c8ee4a2c142d133afe4cbfe567b300fe4d040a8).

There's some hard-coded stuff in here that needs to be fixed, but I thought it was useful to open-up the PR w.r.t. #375.

If you want to reproduce this, you need to ensure that ABC is compiled with `-fPIC`.

Very importantly, this ***passes `ctest` successfully***!!!

Signed-off-by: Andrew V. Jones <andrew.jones@vector.com>

PR #496, opened 2024-09-18
Use ABC as submodule

nindanaoto:
Dear maintainers,

 First of all, this PR is not intended for the merge because it currently includes undesired hacks to get the compilable implementation. 
 I encountered the error complaining that the ABC would not be installed to the same destinations. 
 This is a known issue for CMake.
  https://gitlab.kitware.com/cmake/cmake/-/issues/15415
 It seems to be fixed by introducing `$<BUILD_LOCAL_INTERFACE>` but this is introduced at CMake 3.26. 
 I'm aware that you prefer not to raise the CMake version requirement, so this might be not the final solution for you. 
https://github.com/stp/stp/issues/375#issuecomment-889182453
 One additional weird issue  I'm currently encountering is I could not compile a shared library version of STP, but it can be my environment-specific issue. 
 In addition to CMake related issue, I also had to define machine word size definition macros as you did in the `extabc_trevor` branch.
 https://github.com/stp/stp/blob/01367e30787180261a4bd73b6cc404e9e352aed6/CMakeLists.txt#L84
 
 Anyway, this PR shows the compilable code using the latest ABC and STP code without putting a hardcoded path. I just hope to share this with someone who wants to try such code immediately and to highlight potential issues in this direction.


rgov:
Hello Matsuoka-san, thank you for picking up the work of completing #376.

We do not have a policy of supporting a specific CMake version, but I think we should be compatible with Ubuntu LTS for several years. Ubuntu 20.04 packages CMake 3.16. Ubuntu 22.04 packages CMake 3.22. Therefore I do agree that `$< BUILD_LOCAL_INTERFACE>` seems too new for most users to be able to build STP.

Our CMakeLists.txt may not follow best practices, we have occasional reports of users having difficulty building it, and often the Python environments get confused. If there are improvements you recommend, please let us know.

I think we can merge this in once we have it working on a more common version of CMake. Hopefully it is not too difficult.

nindanaoto:
Hello Govostes-san (I was a bit surprised that you mentioned me in the Japanese way),

 I fixed the code to use the add_custom_command to generate the library. I'm not sure whether it is the best solution for your applications but it works without `$<BUILD_LOCAL_INTERFACE>`. I confirmed that this is compilable with Docker. 
 Is this fine for you?

Dockerfile
```diff
@@ -22,6 +22,8 @@ RUN apt-get update \
         make \
         wget \
         zlib1g-dev \
+        libreadline-dev \
+        libtinfo5 \
```
rgov, line 26: Can we build ABC with `ABC_USE_NO_READLINE=1` to avoid these dependencies?

Otherwise, please keep the alphabetical ordering of dependencies (though `libboost` is out of order).

include/stp/ToSat/BBNodeManagerAIG.h
```diff
@@ -30,9 +30,9 @@ THE SOFTWARE.
 #include "BBNodeAIG.h"
 #include "stp/ToSat/ToSATBase.h"
 
-#include "extlib-abc/aig.h"
-#include "extlib-abc/cnf_short.h"
-#include "extlib-abc/dar.h"
+#include "aig/aig/aig.h"
+#include "sat/cnf/cnf.h"
+#include "opt/dar/dar.h"
```
rgov, line 35: A downside here is that we lose the context that these headers are under `abc/`, however the directory structure does not allow us to do this. We could make a symlink `include/abc -> lib/abc/src` but I don't think that will work on Windows.

We should probably ask the ABC project to restructure its headers into an `include/` directory.

.gitmodules
```diff
@@ -0,0 +1,3 @@
+[submodule "lib/abc"]
+	path = lib/abc
```
rgov, line 2: Should we keep this as `extlib-abc`?

lib/CMakeLists.txt
```diff
@@ -122,7 +120,16 @@ endif()
...
 # in manually.
 # -----------------------------------------------------------------------------
-set(stp_link_libs ${MINISAT_LIBRARIES})
+# https://gitlab.kitware.com/cmake/cmake/-/issues/15415#note_633938
+
+file(COPY abc DESTINATION .)
+ADD_CUSTOM_COMMAND(
+    OUTPUT ${CMAKE_CURRENT_BINARY_DIR}/abc/libabc.a
+    COMMAND cd ${CMAKE_CURRENT_BINARY_DIR}/abc && make libabc.a ABC_USE_PIC=ON -j
+)
```
rgov, line 129: Has this been tested on macOS and Windows?

lib/Interface/cpp_interface.cpp
```diff
@@ -598,6 +598,13 @@ void Cpp_interface::setOption(std::string option, std::string value)
     else
       unsupported();
   }
+  else if (option == "diagnostic-output-channel")
+  {
+    if (value == "stdout")
+      success();
+    else
+      unsupported();
+  }
```
rgov, line 607: Before we merge we should factor this out and complete #495.

rgov (changes requested):

CMakeLists.txt
```diff
@@ -447,7 +447,9 @@ include_directories("${CMAKE_CURRENT_BINARY_DIR}/include")
 # FIXME: External library includes
 include_directories(# For extlib-abc and extlib-constbv
                     ${PROJECT_SOURCE_DIR}/lib
+                    ${PROJECT_SOURCE_DIR}/lib/abc/src
                    )
+add_definitions(-DLIN64 -DSIZEOF_VOID_P=8 -DSIZEOF_LONG=8 -DSIZEOF_INT=4)
```
rgov, line 452: Can we use [`CHECK_TYPE_SIZE()`](https://cmake.org/cmake/help/latest/module/CheckTypeSize.html) to avoid hardcoding these?

rgov (changes requested):

lib/CMakeLists.txt
nindanaoto: Unfortunately, ABC's CMakefile is including the make as an implicit dependency to determine some options. If we hope to build ABC from the source, it's inevitable. 
https://github.com/berkeley-abc/abc/blob/db245f59bdf55041e3308c8753963f1366effedc/CMakeLists.txt#L62

nindanaoto: However, of course, if you really hope that, we can use `camke --build` or something like that to achieve the same result in `add_custom_command`. But I don't think the ABC will likely remove the make dependency someday because the official README is describing only the building instruction using `make`.  I think we should use `make` to avoid the possible issue by using the second-tier support building way.

include/stp/ToSat/BBNodeManagerAIG.h
nindanaoto: I guess that just adding comments is enough. Because ABC is highly popular project, may be they will not hope to change the structure.

.gitmodules
nindanaoto: I agree.

CMakeLists.txt
nindanaoto: I implemented your idea.

.gitmodules
nindanaoto: Fixed.

include/stp/ToSat/BBNodeManagerAIG.h
nindanaoto: Added comments.

Dockerfile
nindanaoto: Removed `libreadline-dev` and fixed the order.

nindanaoto:
@rgov Can I resolve all the requested changes? I'm also wondering whether I should pass the static code analysis check.

lib/CMakeLists.txt
rgov: Interesting that they also call out to `make`.

Could you please share your thoughts on this part of my comment?

> I think this is the biggest obstacle to merging this. As ABC is also a CMake project it should work to include it within CMake's language, rather than calling out to another build tool. I read the described issue but it seems like there are other workarounds. And if ABC is not a private dependency, what's the issue—our library ends up with a handful of extra exported symbols? Doesn't seem like a big deal...

`ADD_CUSTOM_COMMAND` should be lowercase by the way. Please use the `WORKING_DIRECTORY` argument instead of the `cd` command. And yeah, using `cmake --build` does seem better if we absolutely have to call out this way.

nindanaoto: The reason we can't call `add_subdirectory`, is coming from the CMake issue I mentioned in the first comment. 
https://gitlab.kitware.com/cmake/cmake/-/issues/15415
If we do so, CMake will complain that the ABC will not be installed to the same destinations, and, as far as I know, this can be fixed only by using `$<BUILD_LOCAL_INTERFACE>`, CMake 3.26 feature. 

I strongly agree with that if we can use 'add_subdirectory()', it is better. However, I think you feel raising CMake version to 3.26 is not the fair deal for that. 
As a side note, I always prefer snap version of CMake for Ubuntu distirbutions because we can get the latest CMake by the canonical's official packaging system. Hence, if your concern is just to ease the installation on Ubuntu distributions, raising CMake to 3.26 or higher is not much problem, in my opinion. 

I guess that using 'cmake --build' seems to be the way to go although this will not remove the implicit dependency to 'make'.

nindanaoto: I agree with that using `WORKING_DIRECTORY` seems to be better.

lib/Simplifier/AIGSimplifyPropositionalCore.cpp
```diff
@@ -210,11 +213,11 @@ ASTNode AIGSimplifyPropositionalCore::topLevel(const ASTNode& top)
     assert((it->second).size() == 1); // should be a propositional variable.
     const int index =
         (it->second)[0].symbol_index; // This is the index of the pi.
-    Aig_Obj_t* pi = Aig_ManPi(mgr.aigMgr, index);
+    Aig_Obj_t* pi = Aig_ManLi(mgr.aigMgr, index);
     ptrToOrig.insert(make_pair(pi, result));
   }
 
-  Aig_Obj_t* pObj = (Aig_Obj_t*)Vec_PtrEntry(mgr.aigMgr->vPos, 0);
+  Aig_Obj_t* pObj = (Aig_Obj_t*)Vec_PtrEntry(mgr.aigMgr->vCos, 0);
```
rgov, line 220: The static analyzer complained about this, but it isn't a regression so I don't think you need to care.

lib/CMakeLists.txt
nindanaoto: Fixed to use `cmake --build`.

rgov:
I see now for other deps we use `scripts/deps/setup-cms.sh` etc., which do something similar to this. I don't mind it being part of CMakeLists.txt.

I made a quick change to use two `COMMAND` arguments rather than `&&` since I don't know if that works on Windows.

Thanks for contributing and working through the review process with me. Since other maintainers did not chime in, I'll go ahead and merge this. If there's a regression I'm sure someone will file an issue and we can revisit it.

rurban:
You cannot include the external abc as submodule. You really need a patched version, because abc is not C++ ready yet.

I had to patch a bunch of illegal conversions like
```C
#ifdef __cplusplus
static inline Aig_Obj_t* Aig_Regular(Aig_Obj_t* p)
{
  return reinterpret_cast<Aig_Obj_t*>(reinterpret_cast<intptr_t>(p) & ~01);
}
static inline Aig_Obj_t* Aig_Not(Aig_Obj_t* p)
{
  return reinterpret_cast<Aig_Obj_t*>(reinterpret_cast<ABC_PTRUINT_T>(p) ^ 01);
}
static inline Aig_Obj_t* Aig_NotCond(Aig_Obj_t* p, int c)
{
  return reinterpret_cast<Aig_Obj_t*>(reinterpret_cast<ABC_PTRUINT_T>(p) ^ (c));
}
static inline int Aig_IsComplement(Aig_Obj_t* p)
{
  return static_cast<int>(reinterpret_cast<ABC_PTRUINT_T>(p) & 01);
}
#else
```
and 
```C
#ifdef __cplusplus
static inline int Abc_Ptr2Int(void* p)
{
    return static_cast<int>(reinterpret_cast<ABC_PTRINT_T>(p));
}
static inline void* Abc_Int2Ptr(int i)
{
    return reinterpret_cast<void*>(static_cast<ABC_PTRINT_T>(i));
}
static inline word Abc_Ptr2Wrd(void* p)
{
    return static_cast<word>(reinterpret_cast<ABC_PTRUINT_T>(p));
}
static inline void* Abc_Wrd2Ptr(word i)
{
    return reinterpret_cast<void*>(static_cast<ABC_PTRUINT_T>(i));
}
#else
```

jirislaby:
This breaks build of https://github.com/klee/klee:
```
ninja: error: '/home/abuild/rpmbuild/BUILD/stp-2.3.4+20240927-build/stp-2.3.4+20240927/build/lib/extlib-abc/build/libabc-pic.a', needed by 'bin/kleaver', missing and no known rule to make it
```

aytey:
What archives do you have in `/home/abuild/rpmbuild/BUILD/stp-2.3.4+20240927-build/stp-2.3.4+20240927/build/lib/extlib-abc/build/`?

You're looking for the pic variant, but I wonder what variants did get built.

jirislaby:
When building klee, the build directory of stp does not exist anymore, because klee is built completely separately (this is in openSUSE build system).

`.a` cannot be added to the list of required libs -- it is linked into `libstp.so` and no longer available when `libstp.so` is used.

aytey:
... okay, so then this sounds like a problem on your end that however you're building KLEE depends on a static archive that doesn't exist.

Are you able to tell us how you built STP and then how you're building KLEE? That might tell us why KLEE thinks it wants a static archive from STP and not shared object.

jirislaby:
Given
```cmake
set(stp_link_libs ${MINISAT_LIBRARIES} ${CMAKE_CURRENT_BINARY_DIR}/extlib-abc/build/libabc-pic.a tinfo)
```
and
```cmake
target_link_libraries(stp
    LINK_PUBLIC ${stp_link_libs}
)
```
It's barely on my end. Again `.a` cannot be in `LINK_PUBLIC` of `stp`.

jirislaby:
> Would you be willing/able to submit a PR that fixes this to your liking? I don't know why these changes were added in this PR (I guess there was a reason), but given you're experiencing the issue, you might be best placed to try to fix it.

No, sorry, ATM, I am only able to test patches ;).

rgov:
This PR was merged with the understanding we could undo it later if it was a problem. It seems like the abc project needs to make changes upstream. If necessary, go ahead and revert the change, or maybe we should have a fork of abc within the stp namespace (with a PR upstream).

aytey:
I haven't looked yet, but that might be too hasty.

It might (should?) be possible to change our CMakelists such that we link ABC, but not require upstream users to also link ABC.

PR #497, opened 2024-10-22
Added experimental Kissat support

nindanaoto:
Dear maintainers,

 First of all, this PR is not intended for the immediate merge because this includes modification on kissat which is currently handled by my forked version. 
  As you may know, Kissat is the winner of this year's SAT competition, and its latest version is [supported in Bitwuzla](https://github.com/bitwuzla/bitwuzla/pull/124). Since both STP and Bitwuzla are state-of-the-art SMT solvers for the QF_BV problem, some people like me may want to compare them using the state-of-the-art SAT solver, kissat. 
  Hence, I wrote this PR. This PR can include some code not obeying STP's coding rules like explicit use of 'make' for building Kissat because Kissat does not support CMake at all. 
 I hope that some people who have the same need, comparing the state-of-the-art SMT solvers, can find some insights in this PR.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -25,3 +25,4 @@ todo*
 utils
 windows
 Dockerfile
+!scripts/deps/setup-kissat.sh
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -45,6 +45,7 @@ jobs:
       run: |
         ./scripts/deps/setup-minisat.sh
         ./scripts/deps/setup-cms.sh
+        ./scripts/deps/setup-kissat.sh
         ./scripts/deps/setup-gtest.sh
         ./scripts/deps/setup-outputcheck.sh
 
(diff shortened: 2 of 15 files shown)


nindanaoto:
I also wrote the [PR](https://github.com/arminbiere/kissat/pull/57) for Kissat to request merging my fork. 
FYI: [Kissat will not support CMake anytime soon](https://github.com/arminbiere/kissat/pull/17), so if you need CMake to merge, we may have to add that at the STP side.

TrevorHansen:
Thanks for all the work on this. It's a nice effort.

Unfortunately though, I don't think we need Kissat. 

Kissat doesn't support incremental solving, which is essential for how we solve array problems. Also, we've recently added Cadical support, so already have a solver in the same family.

Sorry too for how long this has all taken.

nindanaoto:
It is a reasonable choice. Thank you for clarifying that.

PR #505, opened 2025-07-10
CMake updates

MartinNowack:
A couple of build fixes/enhancements


nindanaoto:
Well, I found that tinfo seems to be not necessary.  As far as I remember,  this was required by ABC, but removing that still goes well. I'm not sure about why I added this. May be this was my fault, and removing that is better.

rgov:
@ccadar Nice to hear from you. Sorry for the delay. I've tried to adapt the "Don't link with tinfo" commit according to the latest CMake configuration. CI builds passed, so it seems to work... let me know if you have any trouble.

ccadar:
@rgov many thanks for the quick response! 

I think there is another issue here with installing `libabc-pic.so`, and I will propose a PR shortly.  Thanks again!

PR #508, opened 2025-08-29
Fix lib/CMkaeLists.txt to use add_subdirectory and absolute path

nindanaoto:
In #496 I used `add_custom_command` to build abc somehow, but I noticed we may be able to use `add_subdirectory` instead since we are using cmake only in `add_custom_command`, so I fixed the lib/CMakeLists.txt. 
I'm not sure whether this will resolve the issue reported as https://github.com/stp/stp/pull/496#issuecomment-2760491937. 
FYI: The solution is basically proposed by Claude Code, if you care.


nindanaoto:
I fixed a typo in the commit message by force push......

PR #513, opened 2026-07-12
Modernize CI: update actions, re-enable gcc, add Windows build

TrevorHansen:
- **actions/checkout v2 → v4** in both workflows (v2 is end-of-life).
- **CodeQL v1 → v3**: the workflow was disabled after GitHub shut down the v1 CodeQL actions. Also checks out submodules (extlib-abc is required to build now), uses `scripts/deps/setup-minisat.sh` instead of a stale inline build, and adds a `workflow_dispatch` trigger. The workflow still needs re-enabling in the Actions UI after this merges.
- **Re-enables the gcc job** alongside clang.
- **Adds a Windows (MSVC) build job** on `windows-2025`: builds minisat and STP statically, using vcpkg zlib and chocolatey winflexbison3, and smoke-tests the binary on a small QF_BV query. This replaces the AppVeyor build, which hasn't triggered since June 2024 (dead integration) and last failed on a hardcoded MSVC toolset path that no longer exists.

If the Windows job proves stable, a follow-up can remove `appveyor.yml` and the README badge.


TrevorHansen:
First version was done building with minisat. Will try getting it working with CMS instead.

PR #515, opened 2026-07-12
Improve unsigned interval propagators, fix leak and dead code

TrevorHansen:
A careful pass over the transfer functions in `UnsignedIntervalAnalysis`. The first commit adds unit tests for the transfer functions — nine of the thirteen fail against master. The second commit makes them pass.

## Precision improvements

- **BVEXTRACT**: the transfer function was completely dead — a stray `break;` sat between the case label and its body. The reinstated version bounds both ends (the old code only bounded the maximum): when no set bits are discarded from the top of the shifted maximum, the extract is exactly a right shift, so the minimum shifts down too. It also no longer mutates the shared cached zero bitvector, which would have tripped the cache's emptiness assertion.
- **BVDIV**: when the divisor's interval includes zero, the old code gave up entirely (the `bottomChanged` recovery code after the early `break` was unreachable). Since division by zero gives all ones, the lower bound — smallest dividend over largest divisor — always holds and is now kept.
- **BVMOD**:
  - remainder by the constant zero is the identity, so the dividend's interval carries over;
  - when the dividend is always below the divisor, the remainder *is* the dividend, so the minimum carries over too (previously only the maximum was kept);
  - when the divisor might be zero, the result is bounded by the larger of the dividend's maximum and the divisor's maximum minus one (previously nothing).
- **BVOR** (new): the minimum is the largest of the children's minimums, and no result bit can be above the highest possible bit of any child.
- **BVXOR** (new): same upper bound as BVOR.

## Cleanups

- Every interval computed by `topLevel()` leaked: nothing deleted the values of the `visited` map, and the `toDeleteLater`/`likeAutoPtr` members that presumably once handled collection were unused. The intervals are now deleted after strength reduction reads them, and the vestigial members are gone.
- SYMBOL/READ/WRITE nodes now cache their (null) result instead of being re-examined through every parent.
- ITE: the `children[1] != NULL` guards could never fail because null children are replaced with complete intervals before dispatch; they now test `knownC1`/`knownC2` as intended (same behaviour, but the guard is real).
- BVMULT: removed the equivalent never-null child check.
- The children-size assertion now runs before the accesses it guards.

## Testing

- New `UnsignedIntervalPropagators_Test` suite: 13 tests, 9 failing before the fix commit, all passing after.
- Full ctest suite (unit tests, API tests, query-files lit suite): all 47 pass.

### DIFF
--- a/include/stp/Simplifier/UnsignedIntervalAnalysis.h
+++ b/include/stp/Simplifier/UnsignedIntervalAnalysis.h
@@ -43,8 +43,6 @@ using std::make_pair;
 
 class UnsignedIntervalAnalysis
 {
-  vector<UnsignedInterval*> toDeleteLater;
-  vector<CBV> likeAutoPtr;
   STPMgr& bm;
   CBV littleOne;
   CBV littleZero;
@@ -55,7 +53,7 @@ class UnsignedIntervalAnalysis
   UnsignedInterval* freshUnsignedInterval(unsigned width);
   UnsignedInterval* getEmptyInterval(const ASTNode& n);
 
-  // We create all intervals through here. Handles collection
+  // The caller owns the returned interval.
   UnsignedInterval* createInterval(CBV min, CBV max);
 
   CBV getEmptyCBV(unsigned width);
--- a/lib/Simplifier/UnsignedIntervalAnalysis.cpp
+++ b/lib/Simplifier/UnsignedIntervalAnalysis.cpp
@@ -666,7 +666,7 @@ namespace stp
         }
         break;
 
-      case BVLEFTSHIFT:
+      case BVLEFTSHIFT: // OVER-APPROXIMATION
         if (knownC0 || knownC1)
         {
           const UnsignedInterval* c0 = children[0];
--- a/lib/Simplifier/UnsignedIntervalAnalysis.cpp
+++ b/lib/Simplifier/UnsignedIntervalAnalysis.cpp
@@ -899,10 +899,12 @@ namespace stp
         break;
       }
 
-      case BVEXTRACT: // OVER-APPROXIMATION
+      case BVEXTRACT:
       {
-        // The index children are always constants, so they're always known.
-        if (knownC0 && knownC2)
+        // The value is (child >> low) mod 2^width. This transfer function
+        // is exact. The index children are always constants; the guard
+        // matters because the shift amount must be the real one.
+        if (knownC2)
         {
           // The lowest bit of the extract is how far the child shifts right.
           unsigned shift_amount = *(children[2]->minV);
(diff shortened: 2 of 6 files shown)


TrevorHansen:
Added a third commit: BVLEFTSHIFT and BVSRSHIFT propagators, plus an exhaustive low-bitwidth soundness suite (`UnsignedIntervalExhaustive_Test`) that checks every transfer function against STP's constant evaluator over all intervals at widths 1-4. Runs in under a second.

TrevorHansen:
Fourth commit classifies every transfer function as perfect or over-approximating, and the exhaustive suite now verifies the classification: the perfect ones (NOT, AND, OR, XOR, EQ, BVGT, BVSGT, ITE, BVNOT, BVUMINUS, BVSX, BVCONCAT, BVPLUS, BVDIV, BVRIGHTSHIFT, BVSRSHIFT) must return exactly the brute-force hull at widths 1-4, and may only return nothing when the hull is complete. The over-approximating ones (BVMOD, BVMULT, BVAND, BVOR, BVXOR, BVEXTRACT, BVLEFTSHIFT) are checked for soundness only. Unknown (null) children are now exercised too.

TrevorHansen:
Fifth commit: BVAND/BVOR/BVXOR are now exact, using the interval bounds from Hacker's Delight section 4-3 implemented over constant bitvectors. The exhaustive suite verifies all three return precisely the brute-force hull at widths 1-4. Remaining over-approximating transfer functions: BVMOD, BVMULT, BVEXTRACT, BVLEFTSHIFT.

TrevorHansen:
Sixth commit: BVEXTRACT is now exact. An extract is (child >> low) mod 2^width and the shifted child covers every value between its shifted bounds, so the result is exactly the bounds' low bits when they agree above the extract's width, and complete (nothing propagated) when the result wraps. Exhaustively verified exact at widths 1-4 for every extract position. Remaining over-approximating transfer functions: BVMOD, BVMULT, BVLEFTSHIFT.

TrevorHansen:
Seventh commit: BVLEFTSHIFT is now exact. (x << s) mod 2^width keeps x's low bits, which run contiguously for each fixed shift (the BVEXTRACT argument), and there are at most width+1 effective shift amounts, so the result is the exact union of per-shift hulls. Exhaustively verified at widths 1-4. Only BVMOD and BVMULT remain over-approximating.

TrevorHansen:
Eighth commit tightens BVMOD: exact for constant divisors (same-quotient test), and the maybe-zero-divisor branch is now exact too (max = dividend's max, reached at divisor zero). At width 4 the inexact fraction drops from 12.4% to 7.9%; the remaining gap is confined to non-constant never-zero divisor intervals, for which no polynomial-in-width exact algorithm appears to be known (LLVM's ConstantRange::urem uses the same approximation).

TrevorHansen:
Ninth commit adds a 100-bit suite: every transfer function checked against brute-force hulls of narrow (at most four value) intervals placed at the boundaries — around zero, across the 2^64 word boundary, across the sign bit, and at the top. This exercises the multi-word CBV paths that the width 1-4 exhaustive tests can't reach. The perfect transfer functions are held to exactness there too; all pass.

TrevorHansen:
Tenth commit rewrites BVMULT. It turned out the old code multiplied interval bounds with BitVector_Multiply, which is signed - any bound in the upper half of the range made the double-width product look negative/overflowing, so the propagator silently gave up (e.g. [0,1]*[0,8] propagated nothing). Sound, but only by accident. Now: unsigned bound products at double width, exact when they land in the same 2^width block, and exact for a constant operand via a floor_sum-based binary search for the extremes of the wrapped arithmetic progression (gated to width <= 64). At width 4 the inexact fraction drops from 39.6% to 19.2%, and every returned interval is exact. Validated by the exhaustive suite, a 300-case randomised width-16 agreement test against brute force, and the 100-bit boundary suite.

PR #542, opened 2026-07-20
Add a bounded disjoint-interval-set domain to NodeDomainAnalysis

TrevorHansen:
Adds an unsigned **interval-set** abstract domain and wires it into
`NodeDomainAnalysis` alongside the existing FixedBits, single interval, and the
recently-added `ValueSet` (#532). Where `ValueSet` enumerates up to a dozen
concrete values, this domain keeps a **bounded list of disjoint intervals** —
better for ranges with a few gaps (sign-split results, `ITE` branches, modulo
holes) that neither a single interval nor a small value set captures well.

## Design

- `UnsignedIntervalSet` (`include/stp/Simplifier/UnsignedIntervalSet.h`): a
  sorted, disjoint, non-adjacent list of `UnsignedInterval` with a cap. When the
  list exceeds the cap, the neighbouring pair separated by the **smallest gap**
  is joined (absorbing the gap), so the closest intervals merge first. Joining
  only widens, so the result is always a sound over-approximation; the
  single-interval domain is exactly the `cap == 1` case.
- `UnsignedIntervalSetAnalysis` (`lib/Simplifier/UnsignedIntervalSetAnalysis.cpp`):
  reuses the existing ~80 single-interval transfer functions unchanged by
  driving `dispatchToTransferFunctions` over the **cross-product** of the
  children's member intervals and unioning the results. `ITE` is handled
  natively (branch union); the signed operands are split at the sign boundary so
  each per-piece transfer stays tight. Unknown/complete children are passed as
  `NULL`, matching the dispatch contract.
- Integration: `buildMap` folds the set's hull into the node's interval (a
  refinement of the plain single-interval transfer) before the existing
  `harmonise`, then clamps the stored set to the harmonised interval. Exposed via
  `getIntervalSetMap()`.

## Testing

- `UnsignedIntervalSet_Test`: container invariants, union, smallest-gap collapse,
  cap, hull, and randomised soundness.
- `UnsignedIntervalSetExhaustive_Test`: at small widths, for every input
  interval-set combination (including genuinely disjoint multi-piece sets),
  every reachable concrete result is a member (soundness), and the set is never
  looser than the single-interval domain (refinement); exact kinds agree.
- Full suite green in both Debug (asserts on) and Release: 57/57.


aytey:
@TrevorHansen : this isn't marked as draft, do you want it merged?

TrevorHansen:
> @TrevorHansen : this isn't marked as draft, do you want it merged?

Nah. Not ready yet. Need to evaluate it a bit more.

PR #559, opened 2026-07-21
Trim allocation and probe overhead in the parsing hot paths

TrevorHansen:
Profiling `--parse-only` runs (after #556/#557) showed node creation and teardown machinery dominating. This PR takes the non-map-implementation wins:

- **Cache each ASTInterior's hash in the node** (lazy, 0 = uncomputed). Children never change once the node is set up, so insertion into and erasure from the unique table no longer rewalk the children. *Note: the separate map-library evaluation also carries a version of this — whichever lands first, the other side's diff shrinks; the two efforts are otherwise disjoint.*
- **Probe the unique table with a stack ASTInterior** and only heap-allocate on a miss. `HashingNodeFactory` previously always allocated a candidate that `LookupOrCreateInterior` deleted whenever an equivalent node existed. The heap node steals the probe's children, node number and cached hash via a new move constructor.
- **Clear the functions map wholesale in `Cpp_interface::cleanUp`** instead of erasing one probe at a time from `~SolverFrame`; teardown of a file with 1.2M define-funs was ~2% of the whole run.
- **Check the define-fun map before the symbol table in the SMT-LIB2 lexer** (guarded by `hasFunctions()`, so files without functions are unaffected). Function-heavy files previously paid a wasted symbol-table probe per reference; a name can't legally be both, so the order isn't observable on valid input.

Interleaved A/B against master (`--parse-only`, release, 4 rounds each):

| file | master | this PR |
|---|---|---|
| AND-NESTED-32-32.smt2 (85MB, 1.2M define-funs) | 6.36s | **5.93s (−7%)** |
| sage/app7/bench_1270.smt2 | 0.57s | 0.56s (noise) |
| asp/Labyrinth laby_18_18_13 (36MB) | 4.01s | 3.97s (noise) |

The remaining parse-time leaders (interior/function-map `find` and rehash, ~35% combined) are hash-map-implementation-bound and left to the separate map evaluation.

Checked: all 59 ctest suites pass; sat/unsat matches the `:status` annotation on 87 benchmarks across five families (sage/app7, Sage2, Sydr, float, SchurNumbers).

### DIFF
--- a/include/stp/AST/ASTInterior.h
+++ b/include/stp/AST/ASTInterior.h
@@ -52,6 +52,12 @@ class ASTInterior : public ASTInternal
   // The vector of children
   ASTVec _children;
 
+  // Lazily computed by ASTInteriorHasher; 0 means not yet computed.
+  // Stays valid because the kind and children never change once set up,
+  // so the hash isn't recomputed on insertion or removal from the
+  // unique table.
+  mutable size_t _hash_cache = 0;
+
   /******************************************************************
    * Hasher for ASTInterior pointer nodes                           *
    ******************************************************************/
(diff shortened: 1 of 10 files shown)


TrevorHansen:
Rebased onto master now that #560 has merged, resolving the overlap:

- The two interior-hash caches are unified: kept this PR's `_hash_cache` (its 0→1 remap means a genuine zero hash still caches, and the move constructor is what the stack-probe `LookupOrCreateInterior` needs), dropped the `_cached_hash` copy that came in with #560. Field name and hasher are as this PR had them, with the merged comment.
- `cleanUp` no longer sets `last_found_function` — #560 removed the memo entirely (with the flat `unordered_dense` map the extra probe it avoided is cheaper than the memo's bookkeeping).
- The stack-probe lookup and bulk `functions.clear()` work unchanged against the `unordered_dense` containers.

On top of merged #560 this PR still gains ~5% on AND-NESTED-32-32 (stack-probe + pooled allocation + lexer ordering; interleaved minimum-of-5, heavily loaded box). 60/60 test suites pass; 2500-file differential sat/unsat sweep vs master: no mismatches.

PR #589, opened 2026-07-23
Store ASTInterior children inline (tail-allocated)

TrevorHansen:
An interior node kept its children in a `std::vector` member: 24 bytes of
header inline plus a **second heap allocation** for the backing array, and
**two cache misses** to reach a child (node → vector data → child). Because
nodes are immutable and hash-consed, that growable container is never grown.

This stores the children in a flexible array immediately after the object,
so a node and its children are a **single allocation**:

- `ASTInterior::create()` sizes one block
  (`sizeof(ASTInterior) + n * sizeof(ASTNode)`), placement-constructs the
  node, then placement-copies the children into the tail; the destructor
  destroys them (releasing their refs) and `operator delete` frees the
  block. Ordinary `new` is deleted so nodes can only be built, correctly
  sized, via `create()`.
- The old lookup built a full **stack "probe" node** and moved its vector
  onto the heap on a miss — impossible for a variable-sized object. It is
  replaced with a **transparent (heterogeneous) unique-table lookup**: a
  non-owning `ASTInterior::Probe{kind, children}` key hashes/compares against
  stored nodes, so a cache hit builds nothing and a miss constructs the node
  directly via `create()`.
- `GetChildren()` returns a `Span` over the tail array (the interface from
  the earlier Span migration), so call sites are unaffected.

**Behaviour-preserving**, verified:
- **byte-identical CNF** (`SRAI-SAFE-8-1024` 194,583 lines; `square` 57,279
  lines);
- 60/60 on a QF_BV sample against declared `:status`;
- **clean valgrind** (0 errors, 0 bytes definitely lost) exercising the
  placement new/delete and inline refcounting.

**Effect:** removes one allocation and one indirection per interior node and
shrinks each node (~30%). Measured a modest CPU improvement on AST-heavy
inputs — `SRAI-SAFE-8-1024 --exit-after-CNF` ~8.44s → ~8.13s user CPU (min
of 3, clean same-window A/B, ~3.7%). Peak RSS is largely unchanged, since
the high-water mark is reached during CNF/bit-blasting rather than while the
AST is live.


aytey:
@TrevorHansen : this isn't marked as draft, do you want it merged?

PR #596, opened 2026-07-23
Give the SAT budgets one meaning across every backend

aytey:
STP's `--max-num-confl` and `--max-time` budgets meant slightly different things depending on which SAT solver was underneath, and a budget of zero was read as "unlimited" by some backends and "give up now" by others. This settles on one contract and enforces it in one place.

### The contract

* **`-1` (or nothing) means no limit**, and is the only negative value accepted. Anything more negative is rejected by the CLI and by `vc_query_with_timeout()` rather than being run as unlimited, which is the dangerous direction for a caller that computed a budget and got the sign wrong.
* **`0` means a budget of zero**: give up without searching. Previously CryptoMiniSat and SimplifyingMinisat quietly turned the smallest budget into no budget at all.
* **The time budget covers the whole query** rather than each call into the SAT solver, of which a query makes one per refinement iteration. `SATSolver` computes the deadline once and every backend measures against it; CaDiCaL's `Terminator` and CryptoMiniSat's own limit now both read from it instead of re-arming per call.

### How

`SATSolver::solve()` is no longer virtual: it enforces the part of the budget that does not depend on the backend, then delegates to a new protected `solveInternal()`. That also gives Minisat, SimplifyingMinisat and Riss — none of which can abandon a search that is already running — a time limit enforced *between* calls, where before `--max-time` was accepted and ignored. The warning they print now describes what actually happens rather than claiming no support.

### Also included

* `vc_useCadical()` / `vc_supportsCadical()` / `vc_isUsingCadical()` and their Python equivalents. The C API previously had no way to select CaDiCaL, so a CaDiCaL-enabled build could switch away from it but not back.
* `RissCore`'s literal accessors are made `const`. They were missed when the `SATSolver` interface was made const-correct in #547, which leaves `RissCore` abstract — `-DUSE_RISS=ON` does not build on master today (`error: invalid new-expression of abstract class type 'stp::RissCore'`).
* `rewrite_rule_gen` no longer passes its `-1` default straight through to `setMaxConflicts()`.

### Known limitation

CaDiCaL's *conflict* budget stays per-call: CaDiCaL exposes no count of the conflicts it has used, so there is nothing to subtract. This is commented in `Cadical::solveInternal`. It is only observable on a query that makes more than one call into the SAT solver; every query I could construct made exactly one. A `conflicts()` accessor from a newer CaDiCaL would close it in a few lines.

### Testing

`tests/api/C/timeout-budget.cpp` checks the semantics against every compiled-in backend, and the Python suite covers the same ground through `check_with_timeout()`. Both include an unlimited query that has to be answered by the SAT solver rather than by simplification — that is the case which catches a backend mistaking "no budget configured" for a budget of zero, which is exactly the bug an earlier draft of this change introduced in CryptoMiniSat.

Built and run in four configurations, covering all five backends:

| configuration | result |
| --- | --- |
| CaDiCaL + Minisat + SimplifyingMinisat | 59/59 non-lit tests pass |
| the above + CryptoMiniSat | 9/9 API tests, 38 Python tests |
| Riss + Minisat (GCC 7) | 9/9 API tests, 38 Python tests |
| pristine `master`, same host | baseline for comparison |

Two suites fail identically on this branch and on a pristine `master` build of the same commit, both for environmental reasons: `query-files` (124/164 — the 40 failures all need `--exit-after-CNF`, i.e. the full CLI, and `libboost_program_options` is not installed on this host, so only `stp_simple` builds) and `test_ast` in the Python suite (`stp.py`'s AST visitor predates `ast.Constant`, so it breaks on Python 3.14). Neither is touched by this change.


aytey:
This was instigated because of the original description on https://github.com/stp/stp/pull/547, which then unearthed that these timeouts were handled inconsistently by the different SAT solver back-ends.

PR #599, opened 2026-07-23
Add --search-bias to tune the SAT search towards sat or unsat

TrevorHansen:
## What

Adds an opt-in `--search-bias {sat,unsat,none}` option that tunes the SAT search
towards one answer. It is a property of the query rather than of one solver:
`SATSolver::setSearchBias()` takes a `SearchBias`, `STP::get_new_sat_solver()`
applies it to whichever back end is in use before the first clause is added, and
each back end maps it onto whatever it happens to offer.

| back end | `unsat` | `sat` |
|---|---|---|
| cadical | `configure("unsat")` — `stabilize=0`, `walk=0` | `configure("sat")` |
| cryptominisat | `set_sls(0)` | unsupported, reported as such |
| minisat / simplifying-minisat / riss | nothing to offer, reported as such | same |

A back end that cannot honour a bias returns `false` rather than silently
dropping the request, and STP warns. `none` is the default, so every back end is
left at its own settings unless the option is passed.

## Why

Modern CDCL solvers alternate between a mode that is good at finding models and
one that is good at closing off the search space. On a workload known to lean one
way — verification conditions are usually unsatisfiable — the effort spent in the
other mode is wasted.

Measured on the `QF_BV/20230221-oisc-gurtner` family (RISC-V OISC correctness
proofs, all UNSAT), where SAT solving is ~99% of wall-clock.

**cadical**, isolated serial, `--cadical` vs `--cadical --search-bias unsat`:

| file              | default | unsat | speedup |
|-------------------|--------:|------:|--------:|
| SRLI-SAFE-48-48   |   76.2s |  5.9s | 12.9x |
| SRAI-SAFE-48-48   |  127.1s | 11.7s | 10.9x |
| SRL-SAFE-128-128  |  114.6s | 48.2s |  2.4x |
| SRL-SAFE-64-64    |   26.0s | 17.4s |  1.5x |
| (8-file frontier) |  503.2s |190.2s | **2.65x** |

It also solves instances that otherwise time out, e.g. SRAI-SAFE-128-128
(timeout → 43s) and SRAI-SAFE-256-256 (timeout → 246s).

**cryptominisat**, interleaved A/B over SRAI-SAFE-{16,20,24,28}: ahead on all 18
pairs, by 12% of wall clock at the median (0.9%–19.4%). Those figures include
STP's own preprocessing, so they understate the effect on the solve itself.

## What was tried and rejected

CryptoMiniSat has no named configurations — `set_up_for_scalmc()` and friends are
model-counting presets, not a sat/unsat bias — so the pieces had to be picked out
by hand, and only one of the two candidates survived measurement.

It rotates its polarity strategy over `{best, stable, best_inv, saved}` as the
search restarts, switching to geometric restarts in the `best` phase. That looks
like the stabilising mode cadical's `unsat` profile turns off. It isn't: pinning
the rotation to `polarmode_saved` was **slower** on 8 of 9 of the same pairs, by
10–49%, and `polarmode_stable` / `polarmode_neg` were worse still. The rotation
earns its keep whatever the answer turns out to be, so the stabilising side of
the bias is left alone for cryptominisat. Its restart strategy is not reachable
through the public API in any case.

## Soundness

A bias only selects heuristics, so it cannot change a correct SAT solver's
verdict. `tests/unit-tests/SearchBias_Test.cpp` holds every back end compiled
into the build to that — verdict stability across `none`/`sat`/`unsat` on both a
satisfiable and an unsatisfiable instance — and checks that a back end which
cannot honour a bias reports it rather than silently dropping it.

A 2501-file differential over the local fuzz corpus (mixed sat/unsat) produced 0
verdict mismatches for the cadical `unsat` path, which is the same
`configure("unsat")` call this PR now reaches through `--search-bias`. The
cryptominisat path has not had that sweep.

## Notes

- Opt-in; default behaviour unchanged for every back end.
- Supersedes the earlier CaDiCaL-only `--cadical-config <name>` on this branch.
  Free-form profile names (`plain`, and anything CaDiCaL adds later) are no
  longer reachable; the option now carries the portable concept only.
- `SearchBias` lives in its own dependency-free header so both
  `UserDefinedFlags.h` and `SATSolver.h` can include it.
- No constructor plumbing was needed for CaDiCaL. `configure()` requires the
  configuring state, but `set("quiet", …)` is exempt from that check
  (`solver.cpp:452-458`) and only `add()` leaves the state, so applying the bias
  after construction is safe.

### DIFF
--- a/include/stp/STPManager/UserDefinedFlags.h
+++ b/include/stp/STPManager/UserDefinedFlags.h
@@ -24,6 +24,8 @@ THE SOFTWARE.
 #ifndef UDEFFLAGS_H
 #define UDEFFLAGS_H
 
+#include "stp/Sat/SearchBias.h"
+
 namespace stp
 {
 
@@ -187,6 +189,10 @@ public:
   };
 
   enum SATSolvers solver_to_use;
+
+  // Which answer to tune the SAT search towards. NONE, the default, leaves
+  // every backend at its own settings, so the option is opt-in.
+  SearchBias search_bias = SearchBias::NONE;
 
   bool get_print_output_at_all() const
   {
--- a/include/stp/Sat/Cadical.h
+++ b/include/stp/Sat/Cadical.h
@@ -77,6 +77,8 @@ public:
 
   uint32_t newVar() override;
 
+  bool setSearchBias(SearchBias bias) override;
+
   void setVerbosity(int v) override;
 
   unsigned long nVars() const override;
--- a/include/stp/Sat/CryptoMinisat5.h
+++ b/include/stp/Sat/CryptoMinisat5.h
@@ -62,6 +62,8 @@ public:
 
   uint32_t newVar() override;
 
+  bool setSearchBias(SearchBias bias) override;
+
   void setVerbosity(int v) override;
 
   unsigned long nVars() const override;
--- a/include/stp/Sat/SATSolver.h
+++ b/include/stp/Sat/SATSolver.h
@@ -25,6 +25,7 @@ THE SOFTWARE.
 #ifndef SATSOLVER_H_
 #define SATSOLVER_H_
 
+#include "SearchBias.h"
 #include "minisat/core/SolverTypes.h"
 #include "minisat/mtl/Vec.h"
 #include <cassert>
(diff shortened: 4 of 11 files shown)


include/stp/STPManager/UserDefinedFlags.h
```diff
@@ -169,6 +171,10 @@ struct UserDefinedFlags
 
   enum SATSolvers solver_to_use;
 
+  // Optional CaDiCaL configuration profile (e.g. "unsat", "sat"). Empty means
+  // leave CaDiCaL at its default configuration.
+  std::string cadical_config = "";
```
aytey: I wonder if this should default to `sat` for most of STP's workloads (where typically you're after a satisfying model)?

TrevorHansen: It's an interesting one. I think before we change the setting on the SAT solvers for everyone, we'd need to measure the difference. Something I need to setup for other changes that I'd like to make where it's not obvious if they're going to help (or not).

Issue #12, opened 2013-09-05
Import fuzzer from my repo


msoos:
I wrote a fuzzer system that does some fun stuff, it's in my repo. It checks unsat with another smt solver, and also fuzzes options. It also checks timings, can use muliple fuzzers, runs multi-threaded... etc. It should also be able to check SAT solutions. Shouldn't be that hard. I need to re-licence it to MIT and move it over.

Issue #11, opened 2013-09-05
Create tool for reducing failing inputs


rgov:
In the past I've hit crashes in STP that required some effort to manually reduce down to the smallest failure-inducing input.

Tools like [C-Reduce](http://embed.cs.utah.edu/creduce/) can automatically reduce C files that trigger compiler bugs, using [delta debugging](http://www.st.cs.uni-saarland.de/dd/).

C-Reduce relies on [Delta](http://delta.tigris.org) which could be used to reduce CVC and SMT-LIB2 inputs also. It probably wouldn't be very complicated to write such a script.

TrevorHansen:
ddsmt is good.  It's not perfect, its results will include expressions that can be trivially simplified like (xor true V). It's a minor complaint though. If we supported printing smtlib2 from STP we could post process the results from ddsmt.

Issue #10, opened 2013-09-05
Remove "using namespace std"-s


msoos:
Code has a couple of "using namespace std". As you might imagine ts makes the code littered with "max" and other stuff that is really hard to spot (is it  define? C func? etc)  Takes 15min to remove, I did that for my fork, but can't merge. I would like to do it, but elbow broken so probably not a good moment to type a lot :)

I am personally fine with e.g. "using std::vector" but importing the whole namespace is just asking for trouble I think.

PR #609, opened 2026-07-25
Bump vendored ABC to upstream 3ed7e4179 (blocked: see description)

TrevorHansen:
> **Do not merge.** The bump is behaviourally neutral for solving and passes `gcc`, `clang` and `gcc (i386)`, but it breaks `gcc (cadical)` and `windows`, and the obvious fix breaks `STATICCOMPILE` builds instead. Opened as a draft so the investigation is on the record and reviewable; see [Blockers](#blockers).

Moves `lib/extlib-abc` from `95393064` (2024-09-22) to `3ed7e4179` (2026-07-23), 633 upstream commits.

## The slice of ABC we use is essentially unchanged

STP uses three headers (`aig/aig/aig.h`, `opt/dar/dar.h`, `sat/cnf/cnf.h`) and ~30 entry points around AIG construction, `Aig_ManDupDfs`/`Aig_ManCleanup`, `Dar_ManRewrite` and `Cnf_Derive`/`Cnf_DeriveSimple`.

* **All three headers we include are byte-identical between the two pins** (same blob hashes). Their transitive include closure is 18 headers (`gcc -MM -DLIN64 -DSIZEOF_VOID_P=8 -I lib/extlib-abc/src`); only 9 changed:
  * `misc/util/abc_global.h`: new `ABC_INT_MAX`; `sprintf`→`snprintf` in `Abc_UtilStrsav{Two,Num}`; `#define inline __inline` now skipped under `_MSC_VER`, a small improvement for our MSVC job.
  * `misc/vec/vec{Bit,Flt,Int,Ptr,Str,Wrd}.h`: `Vec_*` growth-overflow hardening — an `ABC_INT_MAX` cap on the `2 * p->nCap` doubling in `Vec_*Grow`/`Push`/`PushFirst`/`PushOrder`/`FillExtra`, plus `assert(p->nCap < ABC_INT_MAX)` in each `Vec_*Grow`.
  * `misc/vec/vec{Mem,Wec,Wrd}.h`: new inline helpers only (`Vec_MemAllocWithTTs`, `Vec_WecDupSize`, `Vec_WecMaxEntry`, `Vec_WrdStartTruthTables6`, `Vec_PtrPrintPointers`).
  * **No struct or typedef definition anywhere in the closure changed**, so nothing STP passes across the boundary changed layout.
* `src/aig/aig` is **untouched**. In `src/opt/dar` only `darScript.c` changed, purely by adding new scripts — `Dar_ManRewrite` (`darCore.c`) is unchanged. In `src/sat/cnf`, `cnfMan.c` was reformatted and `cnfUtil.c` got a MinGW pthread guard; `Cnf_Derive` and `Cnf_DeriveSimple` are unchanged.
* The one behavioural change reachable from STP is in `Cnf_DataWriteIntoFile`, which now appends unit clauses pinning DIMACS variables that are declared in the header but never referenced in a clause, so the dumped file has the same model count as the formula (a dangling variable otherwise doubles the count, breaking external #SAT / sampling). **This affects `--output-CNF` only** — STP builds its clauses from `Cnf_Dat_t` directly (`ToSATAIG.cpp`) and never reads the file back.

## What was verified good

| check | result |
|---|---|
| `ctest`, default Release config | 63/63 pass (incl. 164-case `query-files` lit suite) |
| `ctest`, `-DWERROR=ON` (CI-equivalent) | builds clean, no warnings outside `extlib-abc` |
| solver verdicts, 96 `.smt2` × 3 modes | 288 runs, **zero** differences vs the old pin |
| dumped CNF clause bodies | byte-identical vs the old pin |
| `gcc`, `clang`, `gcc (i386)` CI | pass |

The three verdict modes were default, `--disable-simplifications` (pushes 60/96 files all the way through AIG → CNF instead of 22), and `--aig-rewrite-passes 4 --aig-core-simplification 1 --bit-blast-simplification 1` (exercises `Dar_ManRewrite` and `AIGSimplifyPropositionalCore`).

Every dumped CNF's clause body is byte-identical; the only file-level difference is the appended dangling-variable unit clauses, e.g. for `square.smt2` the whole diff is one extra `-1 0` line plus the clause count in the header going 57276 → 57277.

**32-bit is fine.** The bump includes `5100825c5` "Only use `__int128` on architectures where it is present", which fixes `misc/util/utilAigSim.c` — added upstream in `0ff43a13c` (2025-12-25), after our old pin, originally using `__int128` unguarded and breaking every 32-bit build. Nothing has touched that file since the fix. Compiling it with `__SIZEOF_INT128__` undefined — the exact preprocessor state a 32-bit build sees, since the guard is `#if defined(__SIZEOF_INT128__)` → `HAVE_INT_128` — is warning-free under `-Wall`, and `-E | grep -c __int128` returns 0. The `gcc (i386)` job passes.

## Blockers

The bump drags in nine new upstream modules STP never reaches — `sat/cadical`, `sat/kissat`, `map/emap`, `misc/btor`, `opt/rar`, `opt/eslim`, `opt/ufar`, `opt/untk`, `opt/util` — and two of them break builds we care about. STP compiles *all* of ABC into `libabc-pic`, so there is no avoiding them.

### 1. ABC now vendors its own CaDiCaL, which collides with ours

`src/sat/cadical` is **CaDiCaL 2.2.0** in `namespace CaDiCaL` — the same namespace as the real CaDiCaL **rel-2.1.3** that STP links under `USE_CADICAL`.

* Against a **static `libcadical.a`** — what `build-cadical` CI uses, since `scripts/deps/setup-cadical.sh` builds only `build/libcadical.a` — the link fails outright with 20+ distinct `multiple definition of 'CaDiCaL::Solver::*'` (`~Solver`, `vars`, `add`, `clause`, `set`, `configure`, `limit`, …), first defined in `libcadical.a(solver.o)`, redefined in `libabc-pic.a(cadical_solver.cpp.o)`.
* Against a **shared `libcadical.so`** it is worse than a build failure: it builds, then crashes. A definition from a static archive overrides one already taken from a shared library, so `Solver::Solver()`, `add(int)`, `solve()` and `~Solver()` come from ABC's 2.2.0, while `Solver::val(int)` — 2.2.0 spells it `val(int, bool)`, so that mangled name is absent from the archive — stays undefined and binds to 2.1.3 at load time. A `Solver` built and driven by one implementation, then read by the other:

  ```
  Fatal error: glibc detected an invalid stdio handle
  #7  _IO_vtable_check ()
  #9  __GI__IO_fflush (fp=0x555555679d60)
  #10 CaDiCaL::Solver::val(int) () from .../libcadical.so
  #11 stp::Cadical::modelValue () at lib/Sat/Cadical.cpp:138
  #12 stp::AbsRefine_CounterExample::ConstructCounterExample () at CounterExample.cpp:72
  ```

  6 of 63 tests crash, nondeterministically (a first `ctest` run showed only 2).

Note that `nm -D` will *not* show this: `--exclude-libs,ALL` makes the archive's symbols local, so a dynamic-symbol clash check between `libstp.so` and `libcadical.so` comes back clean while the process is in fact running two CaDiCaLs. `nm -C` (full symtab) shows 1061 `CaDiCaL::` entries where the old pin had 2.

### 2. `opt/{eslim,ufar,util}` do not compile with MSVC

* `src/opt/eslim/utils.hpp:78` passes an `int` where `_BitScanForward64` wants an `unsigned long *`. It is the `#elif defined(_WIN64)` branch immediately below a commented-out `// #elif defined(_MSC_VER)` — it has evidently never been compiled.
* `src/opt/ufar/UfarMgr.h:22` and `src/opt/util/util.h:23` each define their own `struct timeval` behind `#ifndef _TIMEVAL_DEFINED`, a guard MSVC's headers do not set, so it collides with theirs *and* with ABC's own `lib/pthread.h` `struct timespec`. Cascades into ~20 `C2027 use of undefined type 'timeval'`.

These look like plain upstream portability bugs and are probably worth reporting to berkeley-abc.

### 3. …and excluding those modules breaks `STATICCOMPILE`

The natural fix — filter the nine unused modules out of the `libabc`/`libabc-pic` source lists in `lib/CMakeLists.txt`, in the spirit of #608 — does fix both blockers and most of the build-time cost. It was implemented and measured (see below), and with it:

* `-DUSE_CADICAL=ON` passes 63/63 against both a static `libcadical.a` and a shared `libcadical.so`, with `nm -C | grep -c 'CaDiCaL::'` back to 2.
* the MSVC compile errors are gone.
* the dumped CNF is byte-identical with and without the exclusion, confirming it is behaviour-neutral.

But it then fails to link anything static, because ABC's own command layer references the excluded modules. Reproduced locally with `-DSTATICCOMPILE=ON` on Linux, and on the `windows` job (which passes `-DSTATICCOMPILE=ON`):

```
abc.c:     undefined reference to `Abc_ManRewire' / `seteSLIMParams' / `applyeSLIM'
           / `Abc_BtorCat' / `kissat_solve_cnf' / `Gia_ManRewire'
abcTopo.c: undefined reference to `kissat_init' / `kissat_add' / `kissat_solve' ...
bmcBmc3.c: undefined reference to `cadical_solver_new' / `..._addclause' / ...
```

Chasing that means also excluding `base/abci/abc.c`, `base/abc/abcTopo.c`, `sat/bmc/bmcBmc3.c`, `sat/bmc/bmcMaj7.c` and whatever they in turn drag along — i.e. cutting ABC's command/frame layer out of the build. STP does not use any of it, so that is probably the right end state, but it is a separate piece of work with its own risk, not part of a submodule bump. Stopping here rather than forcing it.

## Cost, measured

| | old pin | new pin (this PR) | new pin + the exclusion in 3 |
|---|---|---|---|
| ABC objects | 1024 | 1256 | 1049 |
| `libabc-pic.a` | 30.0 MB | 41.2 MB | 34.1 MB |
| clean ABC build, user CPU | 452 s | 663 s | 484 s |
| `libstp.so` | 3.421 MB | 3.470 MB | 3.422 MB |
| STP's own build, user CPU | 263 s | 264 s | 263 s |

`sat/{cadical,kissat}` alone are 185 of the 232 new objects. `libstp.so` barely moves because `--exclude-libs,ALL` + `--gc-sections` from #608 discards everything unreachable from STP's ~30 entry points; without those flags the bump would have grown the shipped library by ~11 MB.

Wall clock is not reported: the box was shared during these runs (load 7–11), so user CPU is the comparable metric. Object counts and sizes are exact.

## Other notes

* Our Windows job is MSVC, not MinGW, so the MinGW pthread commits (`f2ae80823`, `a1f5e4541`) do not apply to us, though they are harmless. The relevant upstream Windows work is `f1d6c364c` (CMake Windows), `b73fcb78e` (missing includes, adds `misc/util/winsock_compat.h`), `b7ee7a5f7`, `b413eb90d`, `7a28b20d8` and `ad267aca8` (revived ABC's own Windows CI), plus the `_MSC_VER` guard in `abc_global.h`.
* ABC's `CMakeLists.txt` changed only by adding `CMAKE_EXPORT_COMPILE_COMMANDS`. The `cmake_info` / `SEPARATOR_*` flag extraction STP depends on is unaffected by the new `ABC_MAKE_VERBOSE` gating in ABC's `Makefile`, which only silences `$(info)` lines that path never reads.
* ABC's `Makefile` now auto-prepends `ccache` to `CC`/`CXX` if it is on `PATH`. Harmless for us — STP's CMake passes `CC=`/`CXX=` explicitly on the `make cmake_info` command line, which overrides it, and compilation is done by CMake's own rules anyway.


TrevorHansen:
Abandoning this. The bump is behaviourally neutral for solving but costs +47% ABC build CPU for 232 more objects in modules we don't use, and only three of the 633 upstream commits touch our hot path (none of them performance work). Against that it breaks `gcc (cadical)` and `windows`, and the module-exclusion workaround breaks `STATICCOMPILE` instead.

The root blocker is worth recording for whoever revisits this: upstream now vendors a complete CaDiCaL 2.2.0 under `src/sat/cadical/` in `namespace CaDiCaL`, which collides with the CaDiCaL STP links itself. C++ namespacing does not prevent it and neither does `--exclude-libs`. Resolving that class of collision properly would need `ABC_USE_NAMESPACE`, i.e. compiling ABC into its own namespace — a prerequisite project, not part of a pointer bump.

Leaving the branch in place; the analysis above stands on its own if someone picks this up later.

PR #612, opened 2026-07-25
Add --cnf-generation-effort to trade CNF size against generation time

TrevorHansen:
## Why

CNF conversion is the largest pre-SAT cost on the hard set — **22% of total time (3,887 s)**, about 3× the next stage. Inside it, **74–82% is the cut enumeration** in `Dar_ManComputeCuts`, whose parameters `Cnf_Derive` hardcodes:

| file | cuts | map | save | cuts as % of ABC CNF |
|---|---|---|---|---|
| `mcm/186.smt2` | 44.9 s | 4.0 | 6.1 | 82% |
| `mcm/185.smt2` | 43.2 s | 4.1 | 5.8 | 81% |
| `float/sin.c.175.smt2` | 35.2 s | 3.6 | 4.2 | 82% |
| `oisc/AND-NESTED-12-32.smt2` | 17.0 s | 1.9 | 3.9 | 74% |

On `mcm/186.smt2` that is 44.9 s of a 101.3 s run. Whether the effort is worthwhile is problem-dependent — a smaller CNF can pay for itself in solving, or generation can dominate — and there was no way to choose. `Cnf_DeriveSimple` was reachable only as a side effect of `--disable-simplifications`, which changes ~15 other settings at once.

(These phase numbers come from `Cnf_Man_t::time{Cuts,Map,Save}`, which ABC fills in but only prints from commented-out `ABC_PRT` lines. `Cnf_Man_t` is public in `cnf.h` and `Cnf_ManRead()` is exported, so they are readable without touching the submodule.)

## What

`--cnf-generation-effort {very-low,low,medium,high,very-high}`, measured on `20220315-ecrw/rw_rule_candidate_vmcai_2022_bw512_12.smt2`:

| level | generator | derive | vars | clauses |
|---|---|---|---|---|
| `very-low` | `Cnf_DeriveFast` | 3.3 s | 7,055,392 | 27,563,385 |
| `low` | `Mf_ManGenerateCnf`, LUT 3 | 15.4 s | 5,468,475 | 23,757,984 |
| **`medium`** (default) | `Cnf_Derive` | 20.3 s | 3,816,256 | 21,708,410 |
| `high` | `Mf_ManGenerateCnf`, LUT 6 | 32.5 s | 2,664,270 | 19,682,981 |
| `very-high` | `Mf_ManGenerateCnf`, LUT 8 | 68.5 s | 2,597,687 | 19,511,319 |

`medium` is `Cnf_Derive`, so **the default is unchanged** — byte-identical CNF to the previous code on all 29 test inputs that generate one.

`Mf_ManGenerateCnf` (`src/aig/gia/giaMf.c:1877`) is ABC's newer generator, already in our pinned submodule and used by ABC's own CEC and BMC. It works on `Gia_Man_t`, so the AIG is converted with `Gia_ManFromAig()` first. Worth stating plainly since the name invites the opposite assumption: **it is not a faster `Cnf_Derive`** — at ABC's own default LUT size 8 it is 3.4× slower for 10% fewer clauses, and past LUT 6 the returns are negligible.

## Options the measurements ruled out

Recorded so they are not retried:

* **`Cnf_Man_t::nMergeLimit`** — byte-identical CNF at 4, 6, 8, 10 and 12 (8,999,996 vars / 45,853,244 clauses / 142,898,869 literals at every setting). It gates `Cnf_CutCompose` (`cnfCut.c:312`), which `Cnf_DeriveWithMan`'s flow never reaches.
* **`fSkipTtMin`** — 20.6 s vs 20.3 s, CNF essentially unchanged. Within noise.
* **`Dar_ManComputeCuts` `nCutsMax`** — at 4 it halves derivation (10.0 s) but adds 47% clauses (31.8 M). `Cnf_DeriveFast` is both faster *and* smaller, so this is dominated on both axes.
* **`Cnf_DeriveSimple`** — 12% off generation for **1.9× the clauses** (51.7 M vs 27.6 M) and 2.4× the variables. Plain Tseitin does not earn a level, which is why `very-low` is `Cnf_DeriveFast`.

## Two integration details

Both non-obvious from the headers, so they are commented in the source:

1. **`fAddOrCla` must be 1.** `Cnf_Derive(pAig, 0)` emits the clauses asserting the COs itself; `Mf_ManGenerateCnf` only does so on request. It adds one clause OR-ing all COs, and `toCNF` has already asserted there is exactly one, so that clause is the unit asserting the top-level formula. Without it the CNF is trivially satisfiable.
2. **The `pVarNums` remap is sized with `Aig_ManObjNumMax()`, not `Aig_ManObjNum()`.** The latter subtracts `nDeleted`, and the `Aig_ManCleanup()` in `toCNF` deletes nodes, so object ids run past the count of live objects. `Mf_ManGenerateCnf` returns `pVarNums` indexed by Gia object id while `addVariables`/`fill_node_to_var` index by Aig CI object id, so it is rebuilt over the Aig id space and downstream callers are untouched.

## Testing

* Every level agrees with the default on **all 96 `tests/query-files`** under `--check-sanity`, which constructs *and checks* the counterexample. That is the necessary check here: these paths differ in variable numbering, so verdict agreement alone would not catch a bad mapping.
* `--disable-simplifications` (the `simple_cnf` path) verified unchanged against master, 96/96.
* `ctest` 63/63, including the 164-case `query-files` lit suite.
* `medium` CNF byte-identical to pre-change master on all 29 inputs that produce one (22 of 96 query-files, 7 of 25 real QF_BV benchmarks; the rest are solved during simplification and never reach CNF generation).

## Scope

Only `medium` has end-to-end solve-time evidence. The other levels are backed by CNF-generation measurements; establishing which wins on total time needs a benchmark set selected from runs timed to completion, which is separate work. The flag is what makes that measurable.

`simple_cnf` still short-circuits ahead of the scale, so `--disable-simplifications` behaviour is untouched. Folding it in — deleting `simple_cnf` and having `disableSimplifications()` set `very-low` — would remove a redundant mechanism and halve that path's clause count, but it changes what that flag means, so it is left for a separate discussion. Likewise `traditional_cnf` appears to be dead (read at `STP.cpp:702`, never written) and is left alone here.

### DIFF
--- a/lib/ToSat/ToCNFAIG.cpp
+++ b/lib/ToSat/ToCNFAIG.cpp
@@ -125,7 +125,7 @@ void ToCNFAIG::toCNF(const BBNodeAIG& top, Cnf_Dat_t*& cnfData,
 
   if (!uf.simple_cnf)
   {
-    cnfData = Cnf_Derive(mgr.aigMgr, 0);
+    cnfData = derive_cnf(mgr);
     if (uf.stats_flag)
       cerr << "advanced CNF" << endl;
   }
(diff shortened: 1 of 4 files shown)


lib/ToSat/ToCNFAIG.cpp
```diff
@@ -140,6 +140,86 @@ void ToCNFAIG::toCNF(const BBNodeAIG& top, Cnf_Dat_t*& cnfData,
...
+{
+  Gia_Man_t* pGia = Gia_ManFromAig(mgr.aigMgr);
+
+  // fAddOrCla must be set. Cnf_Derive(pAig, 0) emits the clauses asserting the
+  // COs itself, but this generator only does so on request: it adds a single
+  // clause OR-ing all COs, and toCNF() has already established there is exactly
+  // one, so that clause is the unit which asserts the top-level formula.
+  // Without it the CNF is trivially satisfiable.
+  Cnf_Dat_t* cnfData =
+      (Cnf_Dat_t*)Mf_ManGenerateCnf(pGia, nLutSize, 0, 1, 0, 0);
```
aytey, line 156: I'm not worried about the SCA warning here; we have lots of C-style casts, one more doesn't hurt.

aytey (approved):

PR #614, opened 2026-07-25
Vendor mimalloc and make a fast allocator the default

TrevorHansen:
STP is allocation-heavy at every stage: bit-blasting and CNF generation churn
the AIG and CNF data structures, and the SAT solvers churn clause memory.
Against the C library allocator, a modern replacement is **~15% faster to CNF,
~14% faster for CaDiCaL solving and ~7% for CryptoMiniSat**, with identical
results.

`USE_TCMALLOC` (#607) made that opt-in, and mostly did not fire:

* `find_library()` only looked for the unversioned `libtcmalloc_minimal.so`
  symlink from the `-dev` package, so a machine with only the runtime package
  failed to configure rather than using the library it already had.
* `STATICCOMPILE=ON` could not be satisfied at all without a static archive.

## What this does

Vendor mimalloc as a submodule the same way ABC is vendored, and select the
allocator with `STP_ALLOCATOR=mimalloc|tcmalloc|system`, defaulting to
`mimalloc`. Because it is in-tree, a fast allocator is always available —
including for static builds — and no detection can silently fail. Opting out is
now the deliberate act (`-DSTP_ALLOCATOR=system`).

`USE_TCMALLOC=ON/OFF` still works, with a deprecation warning.

### The allocator goes on the executables, not on libstp

An allocator has to interpose `malloc` process-wide, and `malloc`/`free` are two
halves of one data structure — whoever frees must be the allocator that
allocated. STP's C API hands buffers to callers who free them themselves
(`c_interface.cpp` uses `strdup`/`malloc`, and the header says *"It is the
caller's responsibility to free the buffer's memory"*), so a library that
quietly switched allocators would corrupt already-compiled callers. It is also
the practice [WG21 N1976](https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2006/n1976.html)
warns against.

So `stp` and `stp_simple` link a static mimalloc with `MI_OVERRIDE`; `libstp`
stays allocator-neutral and an application embedding it keeps whatever allocator
it chose. A side benefit is that the exported `STPTargets.cmake` no longer
carries an absolute path to a system allocator.

## Why mimalloc

Benchmarked tcmalloc, mimalloc and jemalloc against each other first. They are
within 1.5% of one another on every workload, so the choice came down to
packaging and memory:

| | tcmalloc | mimalloc | jemalloc |
|---|---|---|---|
| Speed | baseline | ±0% | ±1% |
| Peak RSS, CaDiCaL solve | 172 MB | **155 MB** | 158 MB |
| Vendorable | painful (autotools, ~50k LoC) | **easy (plain CMake)** | painful |
| Windows | no (minimal form) | **yes** | limited |

mimalloc is a plain CMake project that drops into the ABC pattern, and uses ~10%
less peak RSS than tcmalloc during solving.

## Verification

* Every shared object in the process — `libstp`, `libabc-pic`, `libminisat`,
  `libstdc++`, boost and `libc` itself — binds `malloc`/`free`/`operator new` to
  the executable's mimalloc (checked with `LD_DEBUG=bindings`).
* **64/64 tests pass** with assertions on, from a clean build.
* `STATICCOMPILE=ON` produces a fully static binary with mimalloc in it.
* All configurations exercised: default / `system` / `tcmalloc` / an invalid
  value (hard-errors) / `USE_TCMALLOC=ON` / `USE_TCMALLOC=OFF` / static.
* Vendored build measured at **0.861×** the time of the system allocator on a
  19-file interleaved A/B, CI [0.809, 0.914], faster on 17/19 — matching the
  earlier `LD_PRELOAD` study, so vendoring costs nothing.

## New test

`python-allocator-tests` pins the allocator contract of the ctypes interface,
which cannot use the executable's allocator and so has its own rules:

* `libstp` must not embed an allocator (else a ctypes process gets two heaps);
* a preloaded allocator must actually reach `libstp`, i.e. nothing in the build
  defeats symbol interposition;
* plus a plain solve, to catch the allocator change breaking ctypes at all.

Both assertions were checked to be load-bearing: the symbol scan flags the `stp`
binary, which does embed an allocator, and the interposition pattern does not
match when the preload is removed.

The shared `libmimalloc.so` is built so there is something to preload. It is not
installed.

## Notes for review

* **The 32-bit job needs CI to confirm.** `scripts/ci-32bit.sh` builds with
  `WERROR=ON` and will now pull mimalloc in on i386; I have no local i386
  environment. The `-Wno-error` guard mirrors ABC's. If it breaks, the cheap fix
  is defaulting to `system` on 32-bit.
* **Pinned v2.4.1**, the head of mimalloc's v2 line and the version benchmarked.
  v3.4.3 exists and may well be better, but is unevaluated here.
* The shared mimalloc is deliberately not installed, to avoid colliding with a
  system copy. If Python users should be able to preload it after
  `cmake --install`, that is a follow-up worth deciding on.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,3 +1,6 @@
 [submodule "lib/extlib-abc"]
 	path = lib/extlib-abc
 	url = https://github.com/berkeley-abc/abc.git
+[submodule "lib/extlib-mimalloc"]
+	path = lib/extlib-mimalloc
+	url = https://github.com/microsoft/mimalloc.git
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -564,9 +564,12 @@ if (STP_ALLOCATOR STREQUAL "mimalloc")
             "Run: git submodule update --init lib/extlib-mimalloc "
             "(or configure with -DSTP_ALLOCATOR=system).")
     endif()
-    # Only the static library, with the malloc/free/new/delete overrides on.
-    # Static keeps it self-contained: nothing extra to install or ship.
-    set(MI_BUILD_SHARED OFF CACHE BOOL "" FORCE)
+    # The static library is what the executables link: it keeps them
+    # self-contained, with nothing extra to install or ship. The shared one is
+    # built too but not installed - it is there to be LD_PRELOADed by processes
+    # that embed libstp rather than run the stp binary, the Python (ctypes)
+    # interface being the one in tree. See tests/api/python/allocator_tests.py.
+    set(MI_BUILD_SHARED ON  CACHE BOOL "" FORCE)
     set(MI_BUILD_OBJECT OFF CACHE BOOL "" FORCE)
     set(MI_BUILD_TESTS  OFF CACHE BOOL "" FORCE)
     set(MI_BUILD_STATIC ON  CACHE BOOL "" FORCE)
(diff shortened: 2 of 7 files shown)


aytey:
Can we land https://github.com/stp/stp/pull/616 first, please?

Right now, the Windows build _does not work_ without #616.

TrevorHansen:
> Can we land #616 first, please?
> 
> Right now, the Windows build _does not work_ without #616.

Sure, no rush with this.

PR #619, opened 2026-07-25
Fix the extra tools' build and compile them in CI

TrevorHansen:
## Why

`BUILD_EXTRA_TOOLS` defaults to OFF and no CI job enabled it, so nothing compiled the five tools under `tools/` that it guards. **All of them had stopped building.**

This surfaced while checking whether `STP::tosat` was dead code: `rewrite_rule_gen` is its only solving user, and it turned out the tool had not compiled in some time — so "is this member still needed?" could not be answered by building anything.

## Three separate causes

**1. Headers that were never self-contained** — 153 errors. They worked only by accident of include order from one translation unit:

| header | uses | included |
|---|---|---|
| `VariableAssignment.h` | `ASTNode` | nothing at all |
| `misc.h` | `VariableAssignment` | no |
| `rewrite_system.h` | `Rewrite_rule`, `VariableAssignment` | commented out, under the stale path `stp/Util/find-rewrites/rewrite_rule.h` from when this tool lived elsewhere |

**2. `ASTNode::GetChildren()` API drift** — 3 errors. It now returns an `ASTChildren` view (`Span<const ASTNode>`) rather than an `ASTVec&`. Fixed using the codebase's own idioms rather than casting around it:

* `v != n.GetChildren()` → compare like with like, `ASTChildren(v) != n.GetChildren()`, via `Span::operator!=`
* `const ASTVec& f = …` → hold the view directly as `const ASTChildren`; only `.size()` and `operator[]` are used
* `ASTVec s = …` → `toASTVec(…)`, the helper provided for exactly this case, since `s` is sorted in place afterwards

**3. `BBAsProp.h` naming a deleted class.** It still referred to `stp::ToSAT`, removed along with the old non-AIG translator; every "`node_to_satvar_map` was not declared" error cascaded from that single failed member declaration. This was breaking `measure_constantbitprop` and `test_constantbitprop`, not `rewrite_rule_gen`.

Its consumer `test_cbitp.cpp` also called `unit_prop_with_assumps()` and `fixedCount()`, which have since merged into `fixed_count_unit_prop_with_assumps()`. That returns the count directly and **asserts** satisfiability (`CryptoMiniSat5::getFixedCountWithAssumptions`: `assert(CMSat::l_False != r); // always satisfiable`) rather than reporting a conflict, so the harness has no conflict left to observe. Ported with a comment saying so, instead of silently dropping the check.

## The CI change

`-DBUILD_EXTRA_TOOLS:BOOL=ON` on the existing `gcc`/`clang` matrix job — that job already sets `NOCRYPTOMINISAT:BOOL=OFF`, and `rewrite_rule_gen` is gated on `USE_CRYPTOMINISAT` (`tools/rewrite_rule_gen/CMakeLists.txt:21`). Plus a step running `rewrite_rule_gen unit-test` and `rewrite_rule_gen test`.

**Compiling them is most of the value**, since that is the rot this missed. The two modes chosen are the only ones worth running, and the reasoning is in the YAML comment so nobody has to rediscover it:

* they build their own data and `assert` their results, needing no input file;
* assertions are live because CI leaves `CMAKE_BUILD_TYPE` unset, so `CMakeLists.txt:68` defaults to `RelWithDebInfo` and `NDEBUG` is filtered out (`:85-91`) with `ENABLE_ASSERTIONS` defaulting ON;
* `verify`, `expand`, `rewrite` and `write-out` all want a `rules_new.smt2` that is **not in the repository**, and exit 0 in 0.00 s without one — passing vacuously;
* the default no-argument mode is an interactive, unbounded rule search (it prompts "press enter to skip") which **segfaults after ~8.6 s on a stack overflow, ~20 000 recursive `findRewrites` frames deep.**

That last point is worth knowing before anyone tries to make CI generate rules: the generation path does not currently terminate usefully.

## WERROR exemption

The tools emit ~78 warnings — 54 `-Wswitch` over `Kind`, 19 `-Wsign-compare`, plus a few `-Wunused-parameter`/`-Wunused-result`/`-Wparentheses` — so they cannot be gated on `WERROR` without a cleanup out of proportion to keeping them compiling. They are exempted with `-Wno-error`, the same way `lib/CMakeLists.txt` already exempts the vendored ABC, and the comment says the exemption should go once the warnings are fixed. **This is a deliberate deferral, not an oversight.**

## Testing

All five tools build (`rewrite_rule_gen`, `time_constantbitprop`, `measure`, `test_constantbitprop`, `cvc_to_c`), including with `-DWERROR:BOOL=ON`. `rewrite_rule_gen unit-test` and `test` both exit 0 when run from the build directory exactly as the CI step does.

One local caveat: I had to use `-DBUILD_SHARED_LIBS=OFF` to link, because my locally-built CryptoMiniSat exports static GMP (`cryptominisat5Targets.cmake` lists `libgmpxx.a;libgmp.a`) which cannot go into a shared `libstp.so`. That is an environment quirk, not a code issue — CI's `gcc` job already builds shared with CMS enabled and links fine, which this PR's own CI run will confirm.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -124,6 +124,9 @@ jobs:
       run: |
         ./tools/rewrite_rule_gen/rewrite_rule_gen unit-test
         ./tools/rewrite_rule_gen/rewrite_rule_gen test
+        ./tools/rewrite_rule_gen/rewrite_rule_gen verify \
+          ../tools/rewrite_rule_gen/test-rules.smt2
+        ./tools/rewrite_rule_gen/rewrite_rule_gen generate 5 3
       working-directory: build
 
   build-cadical:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -95,7 +95,7 @@ jobs:
       run: |
         mkdir build
         cd build
-        ${{ matrix.env }} cmake -DNOCRYPTOMINISAT:BOOL=OFF -DENABLE_TESTING:BOOL=ON -DWERROR:BOOL=ON -DPYTHON_EXECUTABLE:PATH="$(which python3)" -DCMAKE_C_COMPILER_LAUNCHER=ccache -DCMAKE_CXX_COMPILER_LAUNCHER=ccache -G Ninja ..
+        ${{ matrix.env }} cmake -DNOCRYPTOMINISAT:BOOL=OFF -DENABLE_TESTING:BOOL=ON -DWERROR:BOOL=ON -DBUILD_EXTRA_TOOLS:BOOL=ON -DPYTHON_EXECUTABLE:PATH="$(which python3)" -DCMAKE_C_COMPILER_LAUNCHER=ccache -DCMAKE_CXX_COMPILER_LAUNCHER=ccache -G Ninja ..
 
     - name: Build
       run: |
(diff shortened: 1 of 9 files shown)


TrevorHansen:
Superseded by #621, which does the same build fix but repairs the tools' 85 warnings instead of exempting them from WERROR, and is rebased on current master.

Worth recording why that mattered: one of the warnings was a real defect. test_cbitp.cpp had `assert(signature.maxInputWidth = 1)` — an assignment, not a comparison, so it asserted 1 and checked nothing. Since assert compiles away under NDEBUG the assignment vanished too, leaving maxInputWidth at its original value and changing the loop bound below it, so debug and release behaved differently. Exempting the tools would have preserved that.

PR #627, opened 2026-07-25
Modernize the CMake baseline and make the build CMake 4 clean

TrevorHansen:
The declared minimum of 3.10 stopped being true when mimalloc was vendored: `lib/extlib-mimalloc` requires 3.18 and is the default `STP_ALLOCATOR`, so a default build already needed it. This raises the minimum to 3.18 to match reality, and removes the four `if(POLICY ...)` blocks — every policy they set (`CMP0048`, `CMP0074`, `CMP0077`, `CMP0092`) is NEW by default at 3.18, the newest having arrived in 3.15.

That in turn makes the build work with CMake 4, which removed compatibility with pre-3.5 minimums. CMake 4 is already in our matrix — the `windows-2025` runner ships it, which is why `bae9ff70` had to add `-DCMAKE_POLICY_VERSION_MINIMUM=3.12` for minisat. Three things stood in the way:

**`examples/simple` declared `VERSION 2.8`** and failed to configure outright on CMake 4 (it also called `cmake_minimum_required` after `project()`). Nothing builds this directory — no `add_subdirectory` reaches it — which is how a 2.8 minimum survived this long. CI now builds it against an install tree so it cannot rot again.

**`deps/gtest` was an unpinned clone of the `stp/googletest` fork** on a 2011-era revision declaring `VERSION 2.6.2`, making `-DENABLE_TESTING=ON` a hard configure error on CMake 4 — that would have broken the gcc, clang, cadical and i386 jobs whenever their runners move. `setup-gtest.sh` now clones upstream googletest at a pinned `v1.17.0`. The tests only use `TEST` and `TEST_F`, so no test source needed changing. GTest is consumed from source via `add_subdirectory`, so the script no longer builds and installs a copy nothing reads, and `cache-key.sh` no longer resolves that repo's HEAD now it is pinned.

**`find_package(STP)` could not work for any external project.** `libstp` named its link dependencies `LINK_PUBLIC`, which put `$<TARGET_FILE:libabc-pic>` and an absolute build-tree path to minisat into the exported `STPTargets.cmake`. Consumers have no such target, so configuring against an installed STP died on the generator expression:

```
Error evaluating generator expression: $<TARGET_FILE:libabc-pic>
No target "libabc-pic"
```

These are implementation details of `libstp`, so they are now linked `PRIVATE`. A shared `libstp` propagates none of it: the ABC archive is linked in with its symbols localised, and minisat comes through `DT_NEEDED`. For a static `libstp` CMake still propagates them in-build as `$<LINK_ONLY:...>`, which the static build below confirms.

`Docker.ubuntu20` is deleted: 20.04 ships CMake 3.16.3, below mimalloc's 3.18, so it could no longer build the default configuration, and the release went EOL in April 2025. `Dockerfile` (22.04) is unaffected. The README's claimed requirement of "3.0.2 or newer" is corrected, with a note that `-DSTP_ALLOCATOR=system` opts out of the mimalloc floor for anyone on an older CMake.

## Verification

Locally with **CMake 4.1.2** and **3.22.1**:

- Configure output is identical to before the change apart from line-number shifts and timings; the only remaining warning is the `lib/extlib-abc` submodule's own 3.5.0 deprecation, which is upstream's to fix.
- `ctest` passes **66/66** against the new googletest.
- The static build (`STATICCOMPILE=ON`) links, is genuinely static, and answers sat/unsat correctly.
- `examples/simple` configures, builds and runs against an installed STP — which it could not do before this change.

The CryptoMiniSat configuration could not be built locally (the system CMS config drags in a static `libgmp.a` that cannot go into a shared `libstp.so`); I confirmed the unmodified tree fails identically, so it is pre-existing and left to CI.

## Rebase note

Rebased onto master at `3be43fad` on 2026-07-26. Two conflicts, both from
master moving underneath: `README.markdown`'s CMake version line (#623 had
changed it to 3.10; resolved to the 3.18 this PR argues for) and
`Docker.ubuntu20`, which #622 edited for Python 3 while this PR deletes it
(the deletion stands — 20.04 ships CMake 3.16.3, below mimalloc's floor, and
is EOL). Every claim above was re-verified on the new base; `ctest` is now
66/66 rather than 65/65 because #599 added a test.

## Follow-up: `rewrite_rule_gen` needed its own ABC link

Making libstp's dependencies `PRIVATE` broke `tools/rewrite_rule_gen`, which CI
started building in #621 after this PR was written. It calls ABC itself, not
only through libstp -- `BBNodeManagerAIG`'s inline constructor and destructor
reach `Aig_ManStart`/`Aig_ManStop`, and `getDifficulty()` calls `Cnf_DataFree`
-- and was picking those up from libstp's old public link interface. With the
interface private, and `libstp.so` already localising the ABC symbols it links
in (`--exclude-libs`, unchanged here), the link fails:

```
undefined reference to `Aig_ManStart'
undefined reference to `Cnf_DataFree'
```

It now links `$<TARGET_FILE:libabc-pic>` itself, which restores exactly the
link line it had before this change. The other five extra tools were checked:
none touches ABC, and `measure` / `test_constantbitprop` legitimately need
CryptoMiniSat, which is what their guards are for.

## Not addressed here

`STPConfig.cmake.in` sets `STP_EXECUTABLE`, `STP_STATIC_LIBRARY` and `STP_SHARED_LIBRARY` all to the bare string `stp`, while only the library is exported — so `${STP_EXECUTABLE}` hands consumers a library target. `examples/simple` no longer prints it, but fixing the variable changes the installed package contract and belongs in its own change.

Separately, `cmake --install --prefix` does not relocate headers: `lib/CMakeLists.txt` installs `PUBLIC_HEADER` to the absolute `CMAKE_INSTALL_FULL_INCLUDEDIR` while libraries use the relative `CMAKE_INSTALL_LIBDIR`. The new CI step works around it by setting `CMAKE_INSTALL_PREFIX` at configure time.


aytey:
Right now, KLEE isn't building STP's master, but if they wanted to, this might create an issue: https://github.com/klee/klee/blob/master/CMakeLists.txt#L13C1-L14C1

I am not saying _not_ to land this, I am just saying "it is something to be mindful of".

TrevorHansen:
> Right now, KLEE isn't building STP's master, but if they wanted to, this might create an issue: https://github.com/klee/klee/blob/master/CMakeLists.txt#L13C1-L14C1
> 
> I am not saying _not_ to land this, I am just saying "it is something to be mindful of".

Let's reconsider if it causes problems. We've set to 3.18 because that's the minimum `mimalloc` needs. 3.18 was released in 2020 so not forcing customers onto the bleeding edge.

aytey:
> > Right now, KLEE isn't building STP's master, but if they wanted to, this might create an issue: https://github.com/klee/klee/blob/master/CMakeLists.txt#L13C1-L14C1
> > I am not saying _not_ to land this, I am just saying "it is something to be mindful of".
> 
> Let's reconsider if it causes problems. We've set to 3.18 because that's the minimum `mimalloc` needs. 3.18 was released in 2020 so not forcing customers onto the bleeding edge.

Genuinely, fine with me, I just didn't want us to sleep walk into it.

PR #629, opened 2026-07-25
Find the interpreter with FindPython3 rather than FindPythonInterp

TrevorHansen:
Following the README's build steps on a machine that still has python2 installed fails at configure time:

```
AttributeError: 'module' object has no attribute 'base_prefix'
CMake Error at bindings/python/stp/CMakeLists.txt:69 (message):
  Failed to determine python site package directory
```

`find_package(PythonInterp)` selects `/usr/bin/python2`, and the bindings then ask that interpreter for `sys.base_prefix`, which is python3-only. Every CI job passes `-DPYTHON_EXECUTABLE:PATH=$(which python3)`, so CI has never seen it — the documented `mkdir build && cd build && cmake ..` is the path that breaks.

### Change

`FindPythonInterp` has been deprecated since CMake 3.12, and CMP0148 (3.27) makes using it an error under NEW policy settings, so this moves to `FindPython3`:

- `PYTHON_EXECUTABLE` is still honoured, as a hint to the search — that's what CI, `scripts/ci-32bit.sh` and the README all pass.
- It is still *set* from `Python3_EXECUTABLE` afterwards, because `bindings/python/stp/CMakeLists.txt`, both test suites, `lit.site.cfg.in` and the `GitSHA1` provenance string all read it. No consumer needed touching.
- The bindings decision now keys off `Python3_Interpreter_FOUND` instead of "some interpreter was found".
- Asking for python3 explicitly means a bad hint degrades instead of exploding: `-DPYTHON_EXECUTABLE=$(which python2)` now reports `Could NOT find Python3: Found unsuitable major version "2"` and configures with the interface disabled.

`FindPython3` requires CMake 3.12 (June 2018), so `cmake_minimum_required` goes 3.10 → 3.12 and the README is updated to match. For reference, Ubuntu 20.04 shipped 3.16 and Debian bullseye 3.18, so nothing supported is excluded.

### Verification

All three configurations built and tested with `-DENABLE_TESTING=ON -DNOCRYPTOMINISAT=ON`:

| Configuration | Result |
| --- | --- |
| no `PYTHON_EXECUTABLE` (fails on master) | `Found Python3: /usr/bin/python3.10`, 1278 targets built, **65/65 tests pass** |
| `-DPYTHON_EXECUTABLE=$(which python3)` (what CI passes) | unchanged, configures and builds as before |
| `-DPYTHON_EXECUTABLE=$(which python2)` | `Could NOT find Python3: Found unsuitable major version "2"`, configure succeeds, interface off |

One cosmetic note for reviewers: configure still prints `Found PythonInterp` once. That comes from vendored googletest (`deps/gtest/cmake/internal_utils.cmake:205`), not from STP, and it now resolves to python3 because `PYTHON_EXECUTABLE` is set by the time it runs.

The `-DPYTHON_EXECUTABLE` flags in CI are left in place: they are no longer needed to make the build work, but pinning the interpreter explicitly in CI is still worth keeping.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -18,7 +18,8 @@
 # OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
 # THE SOFTWARE.
 
-cmake_minimum_required(VERSION 3.10 FATAL_ERROR)
+# 3.12 is the floor for FindPython3, which is used to locate the interpreter.
+cmake_minimum_required(VERSION 3.12 FATAL_ERROR)
 
 if(POLICY CMP0048)
     #policy for VERSION in cmake 3.0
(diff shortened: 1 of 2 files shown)


TrevorHansen:
Closing as a duplicate: #622 already replaces `find_package(PythonInterp)` with `find_package(Python3 COMPONENTS Interpreter)`, keeps a `find_package(PythonInterp 3)` fallback so no CMake floor bump is needed, and covers the python3 migration of the scripts, tests and packaging metadata as well. #627 handles the CMake baseline (including `examples/simple`, which still sat at `cmake_minimum_required(VERSION 2.8)`).

PR #640, opened 2026-07-25
Drop two orphan tokens from the SMT-LIB1 parser

TrevorHansen:
`smt.lex` returned `BOOLEXTRACT_TOK` and `BOOL_TO_BV_TOK` for `boolextract` and `boolbv`, and `smt.y` declared both — but **no grammar rule referenced either**. Each keyword was lexed into a token the parser could not shift, making it a guaranteed syntax error.

Neither is standard SMT-LIB1; they are CVC/STP spellings that were never wired into this grammar. Bison doesn't warn about declared-but-unused tokens, which is how they went unnoticed.

### Behaviour is unchanged

The keywords were a syntax error before and remain one after — they now fall through to the identifier rule and come back as an undeclared symbol. Minor upside: the error names the offending token, which it previously did not.

```
$ stp --SMTLIB1 boolextract.smt
syntax error: line 4
  token: boolextract
```

### `nand` / `nor` deliberately left alone

Those two are orphaned in exactly the same way, but they are **not** dead weight: `smt.y` declares them directly under its `/*BV SPECIFIC TOKENS*/` comment, and they are the unprefixed SMT-LIB1 bitvector spellings — of a piece with `concat` and `extract`, which this grammar does implement. They represent an unimplemented part of the input language, so removing them would discard a marker rather than clean anything up.

### Not affected

- The `BOOLEXTRACT` node kind stays. It is still created by the CVC parser (`cvc.y:479-486`), reachable via `stp --CVC`, and handled by the bit-blaster, simplifier, consteval, ValueSetAnalysis, CounterExample, ArrayTransformer and all four printers.
- `bvnand` / `bvnor` / `bvxnor` are properly wired (`smt.y:906,914,805`) and untouched.

### Verification

Rebuilt so bison/flex regenerate, then confirmed (word-boundary match) that `BOOLEXTRACT_TOK` and `BOOL_TO_BV_TOK` occur zero times in `lexsmt.cpp` and `parsesmt.tab.h`, while `NAND_TOK`, `NOR_TOK`, `BVNAND_TOK` and `BVNOR_TOK` all survive.

- All 8 `.smt` files under `tests/` match their `CHECK-NEXT` expectations.
- All 63 `.cvc` files match theirs.


aytey:
Do we actually care about the smtlib (v1) interface? The same likely goes for the old cvc format.

TrevorHansen:
> Do we actually care about the smtlib (v1) interface? The same likely goes for the old cvc format.

Agreed that it's generally best to leave the smtlib1 and cvc parsing alone.

PR #644, opened 2026-07-26
Add SMT-LIB floating-point support (QF_FP, QF_BVFP, QF_ABVFP) via SymFPU

aytey:
This adds SMT-LIB floating-point support to STP: the `QF_FP`, `QF_BVFP` and `QF_ABVFP` logics end to end, from parsing through simplification and bit-blasting to model reporting and print-back. Floats are lowered to pure bitvector circuits during simplification using [SymFPU](https://github.com/martin-cs/symfpu), so everything downstream (AIG/CNF, all SAT backends, arrays, counterexamples) works unchanged. Opened as a **draft** for direction feedback first — it is a large change (130 commits, 112 files) and I am happy to slice it differently if reviewers prefer.

## What is included

- SMT-LIB2 front end: the `(_ FloatingPoint eb sb)` sorts and `Float16/32/64/128`, all `fp.*` operations and classifications, rounding modes, `(fp ...)` literals, the indexed `to_fp`/`to_fp_unsigned`/`fp.to_ubv`/`fp.to_sbv` forms, specials, `=` over floats, and nullary `define-sort` aliases for floating-point sorts (other `define-sort` shapes still answer `unsupported`, and the new smt2-command-tests keep passing).
- Semantics: SMT-LIB's unspecified results (`fp.min`/`fp.max` on ±0, out-of-range `fp.to_ubv`/`fp.to_sbv`) are totalised through shared uninterpreted arrays, so each unspecified result is a function of its operands rather than a free value per occurrence.
- A set of cheap structural rewrites before blasting (constant folding on interned floating-point constants, `fp.sub` lowering, identity/canonicalisation rules), matching what other bit-precise solvers do.
- APIs: ~25 `vc_fp*` functions in the C API (with `getExprKind`/`exprkind_t` kept in sync and compile-time asserted), a small header-only C++ wrapper (`stp/fp.hpp`), and Python bindings (`FloatExpr`, typed model reading).
- Printing: SMT-LIB2 print-back round-trips floating-point input (operators, indexed forms, rounding-mode names, float-typed constants, `set-logic`).
- Tests: gtest suites for the C/C++ API surface, a floating-point lit corpus, and print-back round-trip tests; all wired into CTest and gated on the feature.

## The dependency and the build option

SymFPU is header-only and dual-licensed GPL3/BSD-3-Clause; STP uses the BSD option (recorded in `LICENSE_COMPONENTS`). It is vendored as a pinned submodule at `lib/extlib-symfpu/symfpu` — the same arrangement as mimalloc — so a fresh recursive clone builds floating-point support with nothing to install, and `SYMFPU_INCLUDE_DIRS` can point at an external clone instead.

Support is opt-out: `ENABLE_FLOATING_POINT` defaults to ON, and `-DENABLE_FLOATING_POINT=OFF` builds without SymFPU entirely. The public headers and the `libstp` ABI are identical in both configurations — every `vc_fp*` symbol exists and links either way — and a build without support rejects floating-point input fail-closed with one consistent diagnostic: the parser refuses at the first floating-point construct with a line number (`set-logic QF_FP` fails on line 1), `vc_fpType` refuses at the C API entry point, and the `SetExpWidth`/`CreateFPConst` funnels plus stubbed blaster entry points back both up. A dedicated CI job (`gcc (no-fp)`) builds the opt-out configuration and asserts both that it still solves bitvector input and that the diagnostic appears.

## Validation

Every CI job builds with floating-point on (Linux gcc/clang, Release with assertions off, i386, CaDiCaL, static, MSVC), plus the opt-out job. Beyond CI: differential runs over the floating-point corpus across every refactor, minisat/CaDiCaL cross-solver agreement checks, ASAN-clean API test runs, and `make install` verified in both configurations with a consumer program compiled against the installed tree (identical installed file sets, the `fp.hpp` doc example solving on the ON install and producing the diagnostic on the OFF install).

## Known limitations (stated up front)

- SMT-LIB's format floor (`eb >= 2`, `sb >= 2`) is enforced at parse and API level, but formats with `sb` of 2 or 3 depend on an open SymFPU issue (martin-cs/symfpu#14); the submodule pin will be bumped when that lands.
- Real literals (`((_ to_fp 8 24) RNE 1.5)`) and `fp.to_real` are rejected with dedicated, actionable diagnostics: STP is bit-precise and has no theory of reals, and the SMT-LIB floating-point benchmark families do not use them.
- On a no-floating-point build the C++/Python wrappers abort via `FatalError` at the first floating-point call (as documented in the headers) rather than raising; a capability-query follow-up is planned.
- No SMT-COMP-scale benchmark sweep has been run yet; that is next on my list and I will attach numbers to this PR when I have them.

Provenance: the branch was started in 2021 and completed recently; the Windows/ABC fix it originally surfaced landed separately as #616.


aytey:
**FP benchmarking results: symfpu_rebased vs bitwuzla & cvc5**

Ran the full SMT-LIB floating-point benchmark suites (non-incremental + incremental) against the `symfpu_rebased` build, bitwuzla, and cvc5 with a 5-minute per-file timeout.

**Suites tested:** QF_FP (40,572 files), QF_ABVFP (36,558 files), QF_BVFP (17,250 files) — 94,381 files total.

**Correctness vs bitwuzla:** On the 94,078 files where both STP and bitwuzla returned sat/unsat, they agreed on every single one — **zero disagreements**.

STP had 119 errors (all array extensionality — a known unsupported feature) and 183 timeouts; bitwuzla had 0 errors and 93 timeouts. (cvc5 had 18,862 errors, likely unsupported logic for many ABVFP/BVFP files.)

**Performance vs bitwuzla** (on files where both solved it):

| Category | Files |
|---|---|
| STP 2x+ faster | 587 |
| Comparable (within 5x) | 54,545 |
| STP 5x+ slower | 38,946 |
| — of which 20x+ slower | 15,474 |
| STP solved, bitwuzla timed out | 0 |
| Bitwuzla solved, STP timed out | 138 |

**Summary:** Correctness looks solid — no wrong answers across 94k files. Performance-wise, STP is comparable on ~58% of files but significantly slower on ~41%, particularly on div/sqrt/rem operations. There are 587 files where STP is notably faster.

aytey:
> Summary: Correctness looks solid — no wrong answers across 94k files. Performance-wise, STP is comparable on ~58% of files but significantly slower on ~41%, particularly on div/sqrt/rem operations. There are 587 files where STP is notably faster.
>

Before we get too depressed over this: we don't know how STP compares to bitwuzla/cvc5 _on non-FP problems_ -- so this just might be how it compares for BVs anyway.

TrevorHansen:
> > Summary: Correctness looks solid — no wrong answers across 94k files. Performance-wise, STP is comparable on ~58% of files but significantly slower on ~41%, particularly on div/sqrt/rem operations. There are 587 files where STP is notably faster.
> 
> Before we get too depressed over this: we don't know how STP compares to bitwuzla/cvc5 _on non-FP problems_ -- so this just might be how it compares for BVs anyway.

Nice work. Sounds fine to me, too. I think it's best to get FP all working correctly first, then focus on speeding it up second.

aytey:
subsumed PRs: #739, #740, #741, #742, #743, #744

PR #645, opened 2026-07-26
Add --array-equality: the quantifier-free extensional theory of arrays

aytey:
STP has always rejected equality between whole arrays. With the new `--array-equality` option (C API: `vc_setFlag(vc, 'x')`; Python: `stp.Solver(array_equality=True)`), STP decides the quantifier-free *extensional* theory of arrays: `(= a b)` and `(distinct a b)` over array terms become first-class atoms, including over `store` chains and array-valued `ite`.

The implementation follows

> Robert Brummayer and Armin Biere, *Lemmas on Demand for the Extensional Theory of Arrays*, Journal on Satisfiability, Boolean Modeling and Computation 6 (2010), 165–201.

## Design in one paragraph

Array equalities are abstracted to fresh Boolean variables at node-creation time in the shared node factory, so no downstream pass ever meets an array-sorted `EQ`. Each equality carries the paper's preprocessing-step-1 witness — a fresh index λ and two "virtual reads", constrained by `a = b ∨ a[λ] ≠ b[λ]` — whose defining equations double as anchors from which the equality operands are recovered after STP's own simplification has rewritten them. When the SAT solver produces a candidate assignment, a pure consistency checker runs the paper's propagation rules (seeding, down/up over writes, across true equalities, congruence on insertion; writes participate as reads of themselves per §11.4, with one representative access per concrete index per §11.2) to a fixed point. A congruence conflict yields the §8 lemma — built over shortest propagation paths, a property the FIFO work list provides by construction (§11.1) — encoded as clauses over already-bit-blasted names and added to the incremental SAT solver before re-solving. The loop interleaves with STP's existing lazy read refinement: arrays connected to an abstracted equality are owned by the new checker, all other arrays are refined exactly as before. One shape bypasses the machinery entirely: a chain of writes equated with its own base (the ubiquitous frame condition) rewrites directly into guarded read equalities.

## Compatibility

Default-off, byte for byte: without the option, output, accepted language, and model APIs are unchanged. An opt-in test (`default-off-capi-baseline-differential`) builds the pre-feature baseline from git history and compares the observable C API behaviour byte for byte; it still passes with this branch sitting 37 upstream commits past that baseline.

## Model surfaces

`(get-model)` prints each array as a nullary `define-fun` whose body is a constant-zero array with the observed writes stored on top in ascending index order, replayable in any conforming SMT-LIB2 solver. `vc_getCounterExampleArray` returns deterministic entries sorted by index. Array-valued `(get-value ...)` is rejected as unsupported.

## Testing

- **17 unit tests**: the consistency checker against the paper's worked examples with exact event traces, rule-application counts, and lemma premises; and the solve-time preparation layer (anchor recovery, cone computation, array-ITE elimination with its replacement cache, scalar naming) driven directly, including a death test pinning the fail-loudly contract.
- **15 C API tests**: end-to-end refinement, repeated and incremental solves with exact record counts, model determinism, dormancy when the option is on but unused, and default-off behaviour.
- **22 SMT-LIB2 lit tests**, two of them default-off boundary tests.
- **Python API tests**, which also exercise the procedure under `vc_push`/`vc_pop`.
- Externally, the branch has been validated against a curated 88-file extensionality corpus (44 sat / 44 unsat) and several thousand differential fuzz cases cross-checked against z3, with every printed model replayed through z3.

## Limitations

Arrays of arrays are unsupported (STP's sort system has no nested array sorts); there is no constant-array (`as const`) input syntax; array-valued `get-value` is rejected in favour of `get-model`.

Documentation: `docs/array-extensionality.rst`.

Opening as a draft while large-benchmark hardening continues.


aytey:
replaced by: #745, #746, #747, #748, #749, #750

PR #648, opened 2026-07-26
Harden the fuzzing scripts and use z3 as the reference solver

TrevorHansen:
Follow-up to #646, from reviewing those scripts.

## Correctness

- **A fuzzsmt failure was silent.** If generation failed, `big.smt2` held only the `set-logic` header, both solvers printed nothing, `cmp` passed, and the loop spun at full speed testing nothing. Now checked.
- **`big.smt2` was appended to, not truncated**, so it accumulated across iterations whenever the end-of-loop cleanup wasn't reached.
- **Timeouts were indistinguishable from wrong answers.** `ulimit -t` leaves truncated output, which compares unequal. They're still saved — these problems are small, so running out of CPU is worth investigating — but they're labelled now, and each saved directory gets a `what-happened.txt` recording which solver failed, its exit code, and the options in use.
- **An empty option array** made `$RANDOM % 0` fail every iteration.

## Reference solver

`CHECKER` now defaults to `z3`. Measured on a 2500-query QF_ABV batch:

| solver | time | answers |
|---|---|---|
| z3 5.0.0 | 6.8 s | identical to bitwuzla |
| STP `-d` | 79 s | identical to bitwuzla |
| bitwuzla 0.9.1 | 251 s | baseline |

Confirmed on a second batch with a different seed (6.4 s vs 210 s, identical answers). The checker drops from ~76% of an iteration's wall clock to ~8%. bitwuzla's build was already `buildtype=release -O3`, so this isn't a debug-build artifact — though LTO and alternative SAT backends were not tried.

The checker is now **probed at startup**. FuzzSMT emits the bit-vector overflow predicates, and a solver that doesn't know them is worse than one that fails outright: z3 4.8.12 answers anyway *and* prints an `(error ...)` line, and since only stdout is compared, every iteration would land in `FAIL_DIR` looking like an STP soundness bug. The probe caught that boolector 3.0.1 fails on `bvnego`, so it's dropped from the suggested list.

⚠️ On distros shipping an older z3 (Ubuntu 22.04 has 4.8.12) the probe will now refuse to start until a modern z3 is installed, or `CHECKER=bitwuzla` is set.

## fuzz.sh

- **Cleanup moved into a trap.** Ctrl-C is the documented way to stop, and bash dies on the signal rather than running the last line, so `FUZZ_DIR` was never removed in the normal case. Only a directory the script names itself is removed outright; a `FUZZ_DIR` from the environment may be a scratch area holding other things, so it just gets `rmdir`.
- **Check that `parallel` is GNU parallel** — moreutils ships an unrelated program by that name with no `--ungroup` and no `{}`.
- Dropped the no-op trailing `wait`; restored the trailing newline.

## Testing

`bash -n` on both. Live runs against both z3 5.0.0 and bitwuzla as checker (14 and 24+ iterations), zero spurious failures, zero leftover files after SIGINT and SIGTERM. Verified the probe rejects z3 4.8.12 and boolector 3.0.1 and accepts z3 5.0.0 and bitwuzla 0.9.1. Also confirmed workers get independent `$RANDOM` streams, so parallel copies don't test the same option sequence.

### DIFF
--- a/scripts/fuzz/fuzz.sh
+++ b/scripts/fuzz/fuzz.sh
@@ -6,10 +6,12 @@
 # Usage: fuzz.sh [jobs]
 #
 # Environment:
-#   JOBS      Number of workers. Default: the [jobs] argument, else half the
-#             cores, so a worker and its reference solver each get one.
+#   JOBS      Number of workers. Default: the [jobs] argument, else 1/8th
+#             the number of cores.
 #   FUZZ_DIR  Parent of the per-worker working directories. Default:
-#             /dev/shm/stp-fuzz where tmpfs exists, else $TMPDIR/stp-fuzz.
+#             /dev/shm/stp-fuzz-$$ where tmpfs exists, else $TMPDIR/stp-fuzz-$$.
+#             A directory we generate ourselves is removed on exit; one given
+#             here is left alone beyond rmdir'ing the empty worker directories.
 #
 # Everything fuzz_single.sh reads (STP, CHECKER, FUZZSMT_JAR, FAIL_DIR, ...) is
 # passed straight through, so e.g.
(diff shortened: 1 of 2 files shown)


TrevorHansen:
Added a grouped warning when option settings are dropped. Previously each drop was one line on stderr at startup, which scrolls past among all the workers — and a dropped entry means a code path silently stops being fuzzed while the run looks healthy for hours.

Against a build configured without cadical/cryptominisat:

```
WARNING: 6 of 48 option settings dropped, this build
  does not support them:
    --cnf-generation-effort=very-low  (--cnf-generation-effort is the missing one)
    --cnf-generation-effort=very-high  (--cnf-generation-effort is the missing one)
    --cadical
    --cryptominisat
    --cryptominisat --threads=4  (--cryptominisat is the missing one)
    --interactive=1  (--interactive is the missing one)
  Those code paths are NOT being fuzzed. Rebuild with them enabled if
  that is not deliberate.

42 option settings
```

TrevorHansen:
Two more commits.

**Dropped the standalone `--bb.mult-v2=1`.** It's non-default as far as boost is concerned, so the `--help` filter kept it, but it emits CNF byte-identical to the baseline:

| entry | files with different CNF |
|---|---|
| `--bb.mult-v2=1` alone | **0 of 23** (QF_BV), **0 of 26** (QF_ABV) |
| `--bb.mult-variant=5` | 12 of 23 |
| `--bb.mult-variant=5 --bb.mult-v2=1` *vs* `--bb.mult-variant=5` | 1 of 23 |

Every site reading `upper_multiplication_bound` is gated behind constant-bit propagation having produced `MultiplicationStats`, which doesn't happen under the default multiplication variant — so the entry was quietly re-running the baseline. Paired with variant 5 it does bite (rarely — `multWithBounds()` is genuinely hard to reach), so that entry stays. The header now documents the `--output-CNF --exit-after-CNF` check that catches this class of mistake.

**Options are now drawn from groups so flags combine.** One entry per group — simplify, mult, bitblast, cnf, solver, misc — concatenated. Every group carries an empty entry so it can sit out; an all-empty draw still reproduces the default.

```
  simplify   22 entries
  mult       10 entries
  bitblast    8 entries
  cnf         3 entries
  solver      6 entries
  misc        3 entries
95040 combinations
```

Grouping is what keeps mutually exclusive options apart — the SAT solvers share a group, as do the multiplication variants, so `--minisat` and `--cadical` can never be handed over together.

Testing: 120 randomly drawn combinations run against a fuzz file, all exiting 0 with nothing but sat/unsat on stdout; then 26 harness iterations (~1560 queries) against bitwuzla with zero mismatches.

PR #656, opened 2026-07-27
Add a batch of syntactic simplification rules to the node factory

TrevorHansen:
A batch of self-contained, purely syntactic rules for the SimplifyingNodeFactory. None of them need domain information or sharing counts; they all act on the shape of the children at creation time.

**Comparisons**
- 1-bit `BVGT`/`BVSGT` collapse to their single satisfying assignment, turning word-level comparisons into boolean structure.
- `BVGT(~a, ~b) --> BVGT(b, a)` (complement reverses the unsigned order).
- `x > (x + c) --> x > ~c` and `(x + c) > x --> NOT(x > ~c)`: the sum wraps past `x` exactly when `x` exceeds `~c`.
- 1-bit equalities normalise `(x = 1)` to `NOT(x = 0)` so both polarities of a test hash to the same node.

**Division / remainder**
- `BVDIV(x, x) --> ITE(x = 0, all-ones, 1)`.
- `SBVDIV` by a zero constant and of a zero dividend fold to ITEs matching the SMT-LIB division-by-zero semantics.
- `SBVDIV`/`SBVREM` normalise a negative constant operand to positive (truncating semantics commute with negation). The most negative constant is excluded, since its negation is itself.

**Structure**
- Extracts entirely inside a sign-extension rebase onto the sign bit, so different slices of the extension share one node; extracts entirely inside the original term drop the extension.
- Concats that repeat a 1-bit term (directly, or against its own sign-extension) become a single `BVSX`.
- `(c << s)` with a negative constant base becomes `-((-c) << s)`, normalising constants to positive.
- AND/OR creation now also annihilates a literal against its complement one level down a same-kind child, e.g. `AND(p, AND(q, NOT p)) --> false`. (The scan only flattens same-kind children, so an AND nested in an OR is untouched.)

**Testing**
- Each rule has a new case in `SimplifyingNodeFactory_Exhaustive_Test.cpp` proving it fires and that both factories agree on every assignment at small widths (34 tests pass).
- Full `ctest` suite passes (67/67, including the query-files lit corpus).
- Differential sat/unsat sweep vs the previous build over a 2501-file corpus, new side with assertions on: no mismatches, no assertion failures.

### DIFF
--- a/lib/NodeFactory/SimplifyingNodeFactory.cpp
+++ b/lib/NodeFactory/SimplifyingNodeFactory.cpp
@@ -137,6 +137,40 @@ ASTNode SimplifyingNodeFactory::create_gt_node(const ASTVec& children)
     return ASTFalse;
   }
 
+  // A 1-bit unsigned comparison has a single satisfying assignment: 1 > 0.
+  if (children[0].GetValueWidth() == 1)
+  {
+    const ASTNode a = NodeFactory::CreateNode(
+        EQ, children[0], bm.CreateOneConst(1));
+    const ASTNode b = NodeFactory::CreateNode(
+        EQ, children[1], bm.CreateZeroConst(1));
+    return NodeFactory::CreateNode(stp::AND, a, b);
+  }
+
+  // Bitwise complement reverses the unsigned order: ~a > ~b <=> b > a.
+  if (children[0].GetKind() == BVNOT && children[1].GetKind() == BVNOT)
+  {
+    return NodeFactory::CreateNode(stp::BVGT, children[1][0], children[0][0]);
+  }
+
+  // x > (x + c) <=> x > ~c, and (x + c) > x <=> NOT(x > ~c): the sum wraps
+  // past x exactly when x exceeds ~c.
+  for (unsigned side = 0; side < 2; side++)
+  {
+    const ASTNode& plus = children[side];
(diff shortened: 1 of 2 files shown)


aytey:
> BVDIV(x, x) --> ITE(x = 0, all-ones, 1).
>

Do you know what STP does here today when`x` is 0? I think (and don't quote me here!) z3 treats division as uninterpreted if it depends on division by zero.

TrevorHansen:
> > BVDIV(x, x) --> ITE(x = 0, all-ones, 1).
> 
> Do you know what STP does here today when`x` is 0? I think (and don't quote me here!) z3 treats division as uninterpreted if it depends on division by zero.

Yes, that's something to be careful about. It used to be like that - smt-lib changed so unsigned division is total and evaluates to all-ones.  In STP we've produced all-ones for a while.

TrevorHansen:
Added a follow-up commit: `SBVDIV(x, x)` folds to the same `ITE(x = 0, all-ones, 1)` as the unsigned rule — `sdiv(0,0)` is all-ones just like `udiv(0,0)` (the dividend 0 is non-negative), so no side condition is needed. The exhaustive test checks both divisions produce the identical node. Full suite and the 2501-file differential sweep re-run clean.

PR #667, opened 2026-07-27
Evaluate value-set combinations directly on bit-vectors

TrevorHansen:
The value-set analysis built a hash-consed constant node for every member of every child's set, then ran `NonMemberBVConstEvaluator` over each of the up-to-144 combinations — interning every result into the unique table and cloning it back out. Profiling (`propagator_bench`, Release) showed about half the time in that node machinery (`malloc`/`free` ~26%, `BitVector_Hash` 9%, `LookupOrCreateBVConst`, `ASTBVConst::CleanUp`) and only ~4% in the actual arithmetic; the transfer functions ran at 1.3–16 µs/call against ~20–40 ns for the constant-bit propagators.

This evaluates each combination straight on the CBVs with a file-local `evalOnCBVs()` that mirrors `NonMemberBVConstEvaluator` case for case (booleans as width-1 vectors), and stops evaluating as soon as the result set holds every value of its width, since a complete set can only be discarded at the end anyway — a comparison over two 8-member sets otherwise ground through all 144 combinations just to learn {true, false}.

**Speed** (propagator_bench, ns/call, geometric mean over the 378 valueset rows): **8.6x**.

| ops | speedup |
| --- | --- |
| signed/unsigned comparisons | 17–27x (max 86x) |
| extract / concat / zero+sign extend | 19–23x |
| bvadd / bvsub / bitwise / eq | 7–11x |
| shifts, udiv / urem / sdiv / srem | 3.1–3.6x |
| bvmul | 3.0x |

Spot rows at width 32, 8-member sets: bvand 7114→1185 ns, bvadd 7087→1125 ns, bvult 10674→557 ns, eq 10058→2790 ns, bvmul 15568→8969 ns.

**Verification**
- Full `--domains valueset` propagator_bench run: every `exhaustive_precise`, `exhaustive_unsound` (all zero), `missed_conflicts`, `deducible_bits`, `gained_bits` value is byte-identical to the pre-change run; the exhaustive phase enumerates all 65536 width-3 cases per operation deterministically.
- 500-file differential sweep from the fuzz corpus, assertions build vs master Release: 0 sat/unsat mismatches, 0 assertion failures (the `NodeDomainAnalysis` asserts cross-check the value sets against the interval and fixed-bit domains on every node).
- ValueSet, ValueSet_Exhaustive, NodeDomainAnalysis, StrengthReduction(+Exhaustive) unit tests: 85/85. lit suite: 65/65.


TrevorHansen:
Second commit removes the duplication the first one introduced: the bit-vector-level evaluation now lives in consteval.cpp as the shared core — a CBV-returning `NonMemberBVConstEvaluator` overload for the term kinds and `NonMemberBVConstPredicateEvaluator` for the two-argument predicates (comparisons, equality, boolean extraction, overflow tests). The node-level evaluator keeps only the node glue (child evaluation, interning, TRUE/FALSE, ITE, boolean connectives) and delegates the rest, and `ValueSetAnalysis` calls the same two functions, so each kind's semantics is in one place. Net −454 lines versus the first commit.

Neutrality of the consteval restructuring: `cnf_neutral.sh` emits byte-identical CNF across the hard corpus (39/39, 1 pre-existing no-CNF skip), the value-set exhaustive precision output is byte-identical, a fresh 500-file differential sweep with assertions on found 0 mismatches / 0 assertion failures, and the microbenchmark speeds are unchanged (bvand w32/8-values 1168 vs 1185 ns). Unit and lit suites pass on the -Werror build.

TrevorHansen:
Third commit adds genuinely exhaustive testing. Reviewing coverage honestly: the existing exhaustive checks (`ValueSet_Exhaustive_Test`, propagator_bench's precision phase) compute their ground truth with `NonMemberBVConstEvaluator` itself, so while they exhaustively prove the value-set analysis agrees with the evaluator, they could not have caught the evaluator's own semantics changing in the second commit — truth and analysis would have drifted together. Only the corpus-based CNF byte-compare and answer sweeps tested the evaluator itself, and those aren't exhaustive.

`ConstantEvaluator_Exhaustive_Test` closes that: every kind the evaluator handles, over every input at widths up to 6 (all operand pairs for the binary ops, three-child cases for the n-ary ones, every extend/extract/concat shape, every bit for BOOLEXTRACT), compared against an oracle written in plain int64/uint64 arithmetic that never touches constantbv — including the SMT-LIB division-by-zero definitions, bvsmod's sign handling, bvashr's sign fill, and the six overflow predicates (which the bench doesn't cover at all). Both entry points are checked on every input: the node-level evaluator and the CBV-level functions the value-set analysis uses. ~350k evaluator calls, runs in ~0.6 s, passes.

PR #681, opened 2026-07-28
Distribute predicates over an ITE on the ground path

TrevorHansen:
## Summary

- On the bench-hard set, 99.6% of the operator kinds that stop the ground-path climb (34,631 events, 636 files) are a single kind: **ITE** — the unconstrained variable's chain enters one branch of a select whose condition involves other variables (flag tests, division-by-zero guards).
- Predicates distribute over ITE: `P(g(ite(c, f(x), t))) ≡ ite(c, P(g(f(x))), P(g(t)))`. When x occurs only in one branch, the x-side is an ordinary ground-path instance and collapses to a fresh boolean `v`; the rewrite leaves `ite(c, v, P(g(t)))`.
- x's witness definition is sound regardless of the condition: x only influences the formula when c selects its branch, and then the branch evaluates to exactly `v` by the collapse's forward-validated guarantee.
- One ITE frame per path in this version; the ITE must be single-use like every interior node; the ground suffix above the ITE is re-applied to the other branch. The splice reuses the comparison rule's `MutableASTNode::build` + `replaceWithAnotherNode` mechanics.

## Testing

- Five new tests: distribution through then/else branches, with a ground suffix above the ITE, a shared ITE that must refuse, nested frames that decline — each checks x's elimination plus back-substitution equivalence over every assignment (67/67 suite total).
- Full ctest and a 2,500-file differential sat/unsat sweep vs master to follow in the PR checks; corpus firing numbers measured on the bench-hard set will be posted as a comment.

### DIFF
--- a/lib/Simplifier/RemoveUnconstrained.cpp
+++ b/lib/Simplifier/RemoveUnconstrained.cpp
@@ -110,6 +110,24 @@ void RemoveUnconstrained::replace(const ASTNode& from, const ASTNode to)
   simplifier->UpdateSubstitutionMapFewChecks(from, to);
 }
 
+// Rebuild one collected step as an ASTNode around `in`. Used when a
+// distributed ITE's other branch gets the suffix steps re-applied.
+static ASTNode applyStepToNode(NodeFactory* nf, STPMgr& bm,
+                               const GroundStep& s, const ASTNode& in)
+{
+  if (s.kind == BVSX || s.kind == BVZX)
+    return nf->CreateTerm(s.kind, s.outWidth, in,
+                          bm.CreateBVConst(32, s.outWidth));
+  if (s.kind == BVEXTRACT)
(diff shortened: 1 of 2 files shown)


TrevorHansen:
Measured on the bench-hard set (4,661 hard QF_BV files, same instrumented sweep as the taxonomy):

- Ground-path fires go from **2,564 events / 96 files** (master) to **5,676 events / 475 files** with this PR — 3,111 collapses go through a distributed frame.
- `term_kind_unhandled` (previously 34,631 events / 636 files, 99.6% ITE) drops to **zero**. What remains of the ITE population: 18,666 events / 529 files of *nested* selects blocked by the one-frame limit (the obvious follow-up), and 9,659 events / 414 files of *shared* selects where distributing would be unsound.
- Declines are unchanged at the 9 constant-true predicates (6 files) — the correct refusals.
- Soundness: 2,500-file differential sat/unsat sweep vs master — 0 mismatches, 0 skips; full ctest 69/69.

PR #684, opened 2026-07-28
Distribute predicates over stacked ITEs on the ground path

TrevorHansen:
## Summary

- The one-frame limit of #681 left the largest tractable population on the bench-hard set: 18,666 events / 529 files of *stacked* selects (spear/Sage2 saturation ladders).
- The climb now collects a vector of ITE frames (innermost first) and folds the rewrite inside-out: `P(...ite(c_i, path_i, t_i)...) ⇒ ite(c_k, … ite(c_1, v, P(above_1(t_1))) …, P(above_k(t_k)))`.
- Growth is one rebuilt predicate per frame — linear, capped at 4 frames. Frames must be single-use like every interior node; x's witness definition stays sound because x only matters when every frame selects its branch.

## Testing

- Nested-two-frames flips from declining to distributing; three frames with ground suffixes between them and mixed then/else positions; five-frame stack declines at the cap — all with back-substitution equivalence over every assignment (69/69 suite).
- Corpus firing numbers and the differential sweep to follow as a comment.

### DIFF
--- a/lib/Simplifier/RemoveUnconstrained.cpp
+++ b/lib/Simplifier/RemoveUnconstrained.cpp
@@ -139,12 +139,14 @@ static ASTNode applyStepToNode(NodeFactory* nf, STPMgr& bm,
  * boolean v with var := ITE(v, w_true, w_false) recorded, exactly like the
  * direct EQ rule.
  *
- * One term-level ITE may sit on the path with a non-ground condition and
- * other branch: the predicate distributes over it,
+ * Term-level ITEs may sit on the path with non-ground conditions and
+ * other branches: the predicate distributes over each,
  *   P(g(ite(c, f(x), t)))  ==>  ite(c, v, P(g(t))),
- * where the x-branch P(g(f(x))) collapses to the fresh boolean v as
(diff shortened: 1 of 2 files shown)


TrevorHansen:
Measured on the bench-hard set (4,661 hard QF_BV files), including the cap-8 commit:

| | one frame (#681, master) | recursive, cap 4 | recursive, cap 8 |
|---|---|---|---|
| Fires | 5,676 ev / 475 files | 6,659 / 502 | **7,328 / 515** |

Frame depth at fires (cap 8): 1,772 at one, 1,577 at two, 341+290 at three/four, and a real tail the original cap was hiding — 538 at five, 182 at seven, 17 at exactly eight.

Remaining ITE-related misses: 1,215 events / 257 files still stack deeper than eight (steeply diminishing — not worth chasing further), and 18,081 events / 472 files hit a *shared* select, which is the multi-parent soundness boundary rather than a depth limit.

Soundness: 2,500-file differential sat/unsat sweep vs master — 0 mismatches, 0 skips; full ctest passes; the five-frame ladder that previously declined at the cap now distributes and the nine-frame stack declines.

PR #685, opened 2026-07-28
Retry and verify the winflexbison3 install on Windows CI

TrevorHansen:
The Windows job on master failed at Configure with `Could NOT find BISON` (run 30353379324). The real cause was upstream: community.chocolatey.org returned 503 during `choco install winflexbison3`, and choco exited 0 anyway after "Chocolatey installed 0/0 packages", so the install step passed silently.

Retry the install up to five times with backoff, and fail the step outright if `win_bison`/`win_flex` never appear, so a feed outage either recovers or fails at the step that caused it.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -473,7 +473,17 @@ jobs:
           submodules: 'true'
 
     - name: Install flex and bison
-      run: choco install winflexbison3 --no-progress
+      # choco exits 0 even when the package feed is down ("installed 0/0
+      # packages"), which surfaces later as a confusing missing-BISON
+      # configure error. Retry the install and verify the tools landed.
+      run: |
+        foreach ($attempt in 1..5) {
+          choco install winflexbison3 --no-progress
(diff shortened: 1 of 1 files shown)


aytey:
I actually had a branch in progress on this (not an issue that you've done it), but I was going to solve it via *not* pulling from Choco and using the package for winflexbison3 directly from github.com (also with retries).

Thoughts?

aytey:
Alternative approach is here: https://github.com/stp/stp/pull/689 — it drops Chocolatey entirely and fetches the pinned release archive from upstream instead, so the job no longer depends on `community.chocolatey.org` at all. Same diagnosis as yours, just a different tradeoff (bigger diff and a pinned version, against removing the flaky dependency rather than retrying it). Only one of the two should land — happy for it to be yours.

TrevorHansen:
> Alternative approach is here: #689 — it drops Chocolatey entirely and fetches the pinned release archive from upstream instead, so the job no longer depends on `community.chocolatey.org` at all. Same diagnosis as yours, just a different tradeoff (bigger diff and a pinned version, against removing the flaky dependency rather than retrying it). Only one of the two should land — happy for it to be yours.

Yes, happy to reduce our dependencies.

PR #687, opened 2026-07-28
Price the ITE distribution instead of capping its frames

TrevorHansen:
Replaces `MAX_ITE_FRAMES` in `RemoveUnconstrained::tryGroundPathCollapse` with a measurement of what the rewrite actually costs.

This PR previously raised that cap from four to eight. Measuring it showed the raise bought nothing (geomean 0.9996, −0.02% summed difficulty over the spear family, with more files getting larger than smaller), so the change has been replaced with one that addresses the underlying problem instead.

### What the cap was standing in for

Ground-path collapse climbs from an unconstrained variable `x` towards the root while every node on the way is single-use and every sibling is a constant, and replaces the predicate it reaches with a fresh boolean. Term-level ITEs are allowed to sit on that path because the predicate distributes over them:

```
P(g(ite(c, f(x), t)))  ==>  ite(c, v, P(g(t)))
```

applied per frame from the innermost out. Each frame re-applies the ground steps recorded above it to that frame's other branch, so the rewritten form holds one copy of the suffix chain per frame where the original held one in total.

Whether that costs anything is decided by the other branches, not by the frame count, because the copies are hash-consed against each other:

- other branches `(bvadd y i)` under a constant multiply — `(bvmul (bvadd y i) c)` simplifies to `(bvadd (bvmul y c) (i*c))`, so every frame shares the one multiplier and the rewrite is free at any depth.
- other branches `(bvand y m_i)` under a division — nothing merges, and each frame keeps its own divider.

A frame count cannot separate those. At four it refused deep ladders that cost nothing and admitted shallow stacks that tripled the formula.

### The change

Drop the cap. After the distributed term is built and before it is spliced in, price it against the term it replaces with `DifficultyScore` — STP's existing estimate of the AIG AND-nodes a formula would blast to, summed over distinct nodes so it prices sharing correctly — and decline when it came out bigger. The frame count is still bounded, by the climb's own `MAX_PATH`, so the speculative build is bounded too.

### Measurements

A four-deep ITE ladder over a 64-bit division, CNF clauses:

| | k=1 | k=2 | k=3 | k=4 |
|---|---|---|---|---|
| rewrite declined outright | 3,110 | 3,220 | 3,438 | 3,563 |
| cap 4 (before this PR) | 13,636 | 28,119 | 39,361 | 53,392 |
| this PR | 13,636 | 3,220 | 3,438 | 3,563 |

The guard recovers the declined baseline exactly from two frames on.

Corpus effect, measured as effective difficulty score (the pre-simplification score where difficulty reversion fires, since that discards the round), engagement filter computed per comparison:

| vs cap 4 | engaged | geomean | summed | smaller / larger |
|---|---|---|---|---|
| rewrite off entirely | 1,633 | 1.0657 | +8.15% | 277 / 1,323 |
| cap 8 | 972 | 0.9996 | −0.02% | 342 / 468 |
| **this PR** | 1,030 | 0.9995 | −0.03% | **569 / 394** |

That is over the spear family (1,695 files), which is where essentially all the engagement is: across a 40-file sample of every QF_BV family (1,265 files) only 47 files are touched by the rewrite at all and only 27 by this change, at +0.05% summed. So the headline is neutrality, not a speedup — what the change buys is that the pathological case is no longer reachable, and an unexplained constant is gone.

Cost is below the noise floor: −0.16% instructions over twelve of the engaged files, because the rewrites it declines save more downstream work than the scoring costs.

### Tests and verification

The nine-frame stack that previously had to decline at the cap now distributes — its branches are constants, so each re-applied suffix folds away — and a new four-frame stack with symbolic branches and a remainder above them must decline. Both check back-substitution equivalence over every assignment of their free variables. The pass's suite is 94/94 and `ctest` is 183/183. A differential sweep over a 2,501-file fuzz corpus against the capped behaviour agrees on all 2,501 answers (2,213 sat, 288 unsat, no timeouts).

### DIFF
--- a/tests/unit-tests/RemoveUnconstrained_Exhaustive_Test.cpp
+++ b/tests/unit-tests/RemoveUnconstrained_Exhaustive_Test.cpp
@@ -972,14 +972,14 @@ TEST(RemoveUnconstrained_GroundPath, ite_three_frames_with_suffixes)
 
 TEST(RemoveUnconstrained_GroundPath, ite_frame_cap_declines)
 {
-  // Five stacked frames exceed MAX_ITE_FRAMES: x must survive. One
+  // Nine stacked frames exceed MAX_ITE_FRAMES: x must survive. One
   // shared condition variable keeps the equivalence check enumerable.
   Context c;
   ASTNode x = c.bv();
   ASTNode y = c.bv();
   ASTNode t = c.hf->CreateTerm(BVMOD, W, x, c.konst(4));
-  for (int i = 0; i < 5; i++)
-    t = c.hf->CreateTerm(ITE, W, c.hf->CreateNode(EQ, y, c.konst(i)), t,
+  for (int i = 0; i < 9; i++)
+    t = c.hf->CreateTerm(ITE, W, c.hf->CreateNode(EQ, y, c.konst(i % 8)), t,
                          c.konst(7));
   ASTNode top = c.hf->CreateNode(EQ, t, c.konst(2));
 
--- a/tests/unit-tests/RemoveUnconstrained_Exhaustive_Test.cpp
+++ b/tests/unit-tests/RemoveUnconstrained_Exhaustive_Test.cpp
@@ -1207,22 +1207,49 @@ TEST(RemoveUnconstrained_GroundPath, ite_three_frames_with_suffixes)
   c.checkEquivalent(back, result);
 }
 
-TEST(RemoveUnconstrained_GroundPath, ite_frame_cap_declines)
+TEST(RemoveUnconstrained_GroundPath, ite_deep_stack_distributes)
 {
-  // Five stacked frames exceed MAX_ITE_FRAMES: x must survive. One
-  // shared condition variable keeps the equivalence check enumerable.
+  // Nine frames, deeper than any fixed cap would allow. Every other branch
+  // is a constant, so each re-applied suffix folds away and the distributed
+  // form is smaller than the predicate it replaces: it must be taken. One
+  // shared condition variable keeps the equivalence check enumerable, its
+  // constant wrapped to the W-bit range.
   Context c;
   ASTNode x = c.bv();
   ASTNode y = c.bv();
   ASTNode t = c.hf->CreateTerm(BVMOD, W, x, c.konst(4));
-  for (int i = 0; i < 5; i++)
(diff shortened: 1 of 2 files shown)


tests/unit-tests/RemoveUnconstrained_Exhaustive_Test.cpp
```diff
@@ -972,14 +972,14 @@ TEST(RemoveUnconstrained_GroundPath, ite_three_frames_with_suffixes)
...
-  // Five stacked frames exceed MAX_ITE_FRAMES: x must survive. One
+  // Nine stacked frames exceed MAX_ITE_FRAMES: x must survive. One
   // shared condition variable keeps the equivalence check enumerable.
   Context c;
   ASTNode x = c.bv();
   ASTNode y = c.bv();
   ASTNode t = c.hf->CreateTerm(BVMOD, W, x, c.konst(4));
-  for (int i = 0; i < 5; i++)
-    t = c.hf->CreateTerm(ITE, W, c.hf->CreateNode(EQ, y, c.konst(i)), t,
+  for (int i = 0; i < 9; i++)
```
aytey, line 981: Worth adding `const size_t MAX_ITE_FRAMES = 8;` to this file and adding a comment to both to "keep them in sync"?

(I'm hesitant to say "put it in header").

TrevorHansen: Done in a7964ab9 — the test now declares `const unsigned MAX_ITE_FRAMES = 8;` locally and builds the stack as `MAX_ITE_FRAMES + 1`, with a "keep in sync" comment at both sites. Left it out of a header, as you suggested. Checked it is genuinely boundary-sensitive: setting the cap in the pass to nine makes `ite_frame_cap_declines` fail.

The `% 8` on the condition constant was unrelated to the cap — it wraps into the 3-bit test width — so it now reads `% (1u << W)`.

TrevorHansen: This is moot now — the PR has been rewritten. `MAX_ITE_FRAMES` is deleted rather than raised: the frame count turned out to be the wrong thing to bound, so the rewrite now prices the term it builds against the one it replaces and declines when it grew. With no constant left there is nothing for the test to stay in sync with, and `ite_frame_cap_declines` has been replaced by two tests that exercise the two sides of that decision.

TrevorHansen:
The benefit isn't clear enough to justify this change.

PR #688, opened 2026-07-28
Replace shared non-surjective terms with image-constrained fresh variables

TrevorHansen:
## Summary

- The largest single shape among the remaining ground-path misses is a **shared derived term**: `zx(x)`, sign-extensions, checksum chunks of a single-use variable consumed by several parents (16k zero-extension events alone on bench-hard). Predicate-level collapse is impossible there, and a plain fresh variable is unsound (the image is smaller than the domain).
- When the variable's single-step chain ends at a shared `t = op(x, c)` whose image is exactly and cheaply expressible, replace **t itself** with a fresh `v`, conjoin `v ∈ Image(t)`, and define `x := projection(v)`: `∃x. φ(t(x)) ⟺ ∃v. v ∈ Image(t) ∧ φ(v)`.
- Shape table: concat with constant either side, zx (high bits zero), sx (fixed point of re-extension), even multiplies (low bits zero + odd-part inverse), constant shifts, low masks, constant modulus/division (interval bounds).
- Consumers then see a *variable*, unlocking the symbol-keyed passes (equality propagation, BVSolver, the disjoint-extract splitter, this pass itself).
- This rewrite is equisatisfiable **with model mapping**, not pointwise-identical after back-substitution — the new `checkEquisat` harness verifies both properties over every assignment.

## Testing

- Seven new tests: the five rewritten shapes, a kind outside the table (must decline), and a multi-step shared chain (must decline); suite 75/75.
- Full ctest and a 2,500-file differential sat/unsat sweep vs master to follow in checks; bench-hard impact numbers to be posted as a comment. Sized beforehand: 1,073 of 1,074 shared-node files on the hard set have ≥1 convertible node (833 fully convertible).

### DIFF
--- a/include/stp/Simplifier/RemoveUnconstrained.h
+++ b/include/stp/Simplifier/RemoveUnconstrained.h
@@ -33,6 +33,7 @@ THE SOFTWARE.
 #include "stp/AST/AST.h"
 #include "stp/AST/MutableASTNode.h"
 #include "stp/STPManager/STPManager.h"
+#include "stp/Simplifier/AchievableImage.h"
 #include "stp/Simplifier/Simplifier.h"
 
 namespace stp
@@ -51,6 +52,14 @@ class RemoveUnconstrained
 
   bool tryGroundPathCollapse(MutableASTNode& muteNode,
                              vector<MutableASTNode*>& variables);
+
+  bool tryImageConstrainShared(const ASTNode& var, MutableASTNode& sharedNode,
+                               const GroundStep& step,
+                               vector<MutableASTNode*>& variables);
+
+  // Membership constraints produced by tryImageConstrainShared during a
+  // topLevel_other() run; conjoined onto the result before returning.
+  ASTVec imageConstraints;
 
   void replace(const ASTNode& from, const ASTNode to);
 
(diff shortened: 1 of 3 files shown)


TrevorHansen:
Converting to draft: the repo's own test suite caught a regression that invalidates the design premise.

`simplification-tests/mod2.smt2` previously solved entirely in preprocessing: the shared `(concat (_ bv0 5) ((_ extract 9 0) x))` structurally tells constant-bit propagation that the high bits are zero, which folds `(bvurem · 12000)` to the identity and collapses the asserts. After this rewrite the term is a bare variable whose image lives in a separate conjunct — and the over-approximating analyses read *structure*, not side constraints, so the simplification is lost and the instance reaches the SAT solver (verdict unchanged; equisatisfiability holds).

The general lesson: every shape in the table (zx high-zeros, mod bounds, even-multiply strides) carries its image information structurally, exactly where cbitp/intervals can use it. Replacing structure with a relational constraint deletes information from those passes, and the hoped-for symbol-keyed enablement doesn't compensate. Landing this would require teaching the domain analyses to mine top-level constraints first — a different and larger feature.

Leaving as draft for the record; fine to close.

PR #691, opened 2026-07-28
Slice equalities over concatenations of an unconstrained variable

TrevorHansen:
## Summary

- The factory already splits `(= (concat a b) K)` against a **constant**; equalities against symbolic terms never split, so `(= (concat x y) t)` with x single-use survives the pipeline (the log-slicing family is built almost entirely from such equalities).
- The climb now steps through `BVCONCAT` nodes with symbolic siblings (pure concatenation only, from the bare variable), and at an equality against **any** term slices by bit range: `x := extract_hi(t)`, one residual equality per sibling slice.
- **No fresh boolean is needed**: x's conjunct becomes identically true under its definition, and falsity survives through the sibling residuals. The rewrite is pointwise, so it holds in any polarity and under shared predicates, and when a sibling slice is itself free, its residual re-enters the worklist and the EQ rule introduces the boolean exactly where falsity would otherwise be lost.

## Testing

- Six new tests: basic split, nested frames with a constant slice, cascading free sibling, constant other side via the hashing factory, and refusals for comparisons and shared concats — back-substitution equivalence over every assignment (72/72 suite).
- Full ctest and the 2,500-file differential sweep vs master to follow; per-file impact on the log-slicing family to be posted as a comment.

### DIFF
--- a/lib/Simplifier/RemoveUnconstrained.cpp
+++ b/lib/Simplifier/RemoveUnconstrained.cpp
@@ -182,6 +182,17 @@ bool RemoveUnconstrained::tryGroundPathCollapse(
   const size_t MAX_ITE_FRAMES = 4;
   std::vector<IteFrame> frames;
 
+  // Concat frames: BVCONCAT nodes with a possibly-symbolic sibling.
+  // Equality distributes over concatenation slice by slice, so these
+  // need no image -- but only pure concatenation may then sit between
+  // the variable and the predicate, and the predicate must be EQ.
+  struct ConcatFrame
+  {
+    MutableASTNode* sibling;
+    bool pathHigh; // the path is child 0 (the high part)
+  };
+  std::vector<ConcatFrame> cframes;
+
   MutableASTNode* cur = &muteNode;
   for (unsigned depth = 0; depth < AchievableImage::MAX_PATH; depth++)
   {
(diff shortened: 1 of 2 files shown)


TrevorHansen:
Honest impact numbers, measured before review:

- **Fires once on the entire 4,661-file bench-hard set** (one slice, in 2018-Mann/arbiter_data_integrity_n4q8w8d8b30.smt2). Soundness verified as usual: 72/72 suite, full ctest, 2,500-file differential sweep with zero discrepancies.
- Both motivating residual instances turn out to be structurally out of scope: the log-slicing concats sit **under extracts** before their equalities (the pure-concatenation requirement bails), and the Sage2 concat words have **shared spines** — the same node feeds hundreds of consumers, and defining the free slice from one consumer's other side would constrain the rest, so the single-use refusal is mandatory, not a limitation.
- Everything else in this shape family was already covered by the factory's constant-side splitting.

So this is a correctness-completing rule with essentially no corpus yield — the mechanism is right and the tests document the falsity-preservation argument, but I'd understand either merging it as completeness or closing it as dead weight. Extending through extract wrappers could reach the log-slicing family, but I wouldn't build that without first measuring whether the extract-wrapped population survives the factory's extract-of-concat folding at all.

TrevorHansen:
Rarely useful.

PR #692, opened 2026-07-28
Extensionality + FPs

aytey:
This stacks https://github.com/stp/stp/pull/644 and https://github.com/stp/stp/pull/645, resolves the merge conflicts, and then fixes any issues that come out from the combination of both (mostly, fixes in this branch *do not* land obviously in either child branch).


aytey:
replaced by: #739, #740, #741, #742, #743, #744, #745, #746, #747, #748, #749, #750

PR #693, opened 2026-07-29
Collapse predicates one-sidedly when their polarity is pure

TrevorHansen:
Stacked on #691 (base is `concat-equality-slicing`; only the top commit is new).

The ground-path collapse required every sibling on the path to be a constant, so `(= (bvand x t) 0)` with symbolic `t` could not eliminate an unconstrained `x` — even though `x := 0` forces the bvand, and with it the equality, regardless of `t`.

**Singleton mode.** A symbolic sibling now switches the climb into singleton mode when the path can be forced to the operator's absorbing element: 0 through `bvand`/`bvmul`, all-ones through `bvor`, a zero value operand through the shifts, a zero dividend through the remainders (STP defines remainder-by-zero as the dividend). The absorbing element may be reached through a ground prefix (the image supplies a validated witness), and the known value flows through later constant steps by evaluation. Only that single outcome of the predicate is then known, so the collapse is one-sided: it fires exactly when the predicate's global polarity is pure and prefers the forced outcome. Making a positive-only predicate true (or a negative-only one false) cannot falsify a model of the original (monotonicity), and any model of the rewrite extends with `x := witness`, which realises the value — so satisfiability is preserved both ways. The predicate becomes a constant and the witness lands in the substitution map, so models reconstruct as usual. ITE frames above the absorbing step still distribute; frames below it decline (their rebuilt other-branch would need the symbolic step re-applied).

**Image fallback.** The same rule rescues all-constant chains where the image can achieve only one polarity (`AchievableImage::decideOneSided`), e.g. `(not (= (zero_extend x) 63))`: the two-sided decide fails for lack of a true witness, but under pure negative polarity forcing false suffices. This converts the previously deliberate `decide_declined` refusals when the polarity allows.

**Polarity** is computed by walking the mutable tree upward from the predicate through AND/OR (same), NOT/NAND/NOR (flip), IMPLIES (antecedent flips), and boolean ITE branches; ITE conditions, XOR/IFF operands and anything unrecognised count as both polarities and decline. Shared predicates are fine when all occurrences agree.

**Testing.** 17 new exhaustive tests including the must-decline cases (wrong polarity, mixed polarity, shared mixed-polarity predicate, frame below the absorbing step). The suite gains a satisfiability-equality check: the pointwise back-substitution identity still holds for one-sided rewrites (the witness decides the predicate independently of the other variables) but cannot by itself catch a wrong-polarity rewrite — a deliberately inverted gate passes pointwise while flipping sat to unsat. Full suite 89/89, ctest 69/69 (Debug + assertions), 2500-file differential sat/unsat sweep vs an old-master baseline: 0 disagreements, 0 skips. Model construction verified end-to-end (`x = 0` comes back out of the solver on the motivating formula).

**Measured corpus yield: small but real.** Temporary instrumentation (reverted): 0 fires on a 1000-file sample (400 random QF_BV + 600 from the families where ground-path fires cluster), but the full 4661-file hard set finds **7 firing files (all Sage2), 10 fires** -- 9 via the image fallback (exactly the old `decide_declined` constant-predicate population, now converted under pure polarity) and 1 genuine singleton-mode fire (Sage2/bench_15975). A differential check on the firing files at 60 s agrees everywhere both sides solve (one file times out on both binaries). So this is a capability-class extension (the alignment-check idiom `(x & mask) == 0` with a symbolic mask) whose corpus rate is below random-sample resolution; it is CNF-behaviour-neutral wherever it does not fire.

### DIFF
--- a/lib/Simplifier/RemoveUnconstrained.cpp
+++ b/lib/Simplifier/RemoveUnconstrained.cpp
@@ -127,6 +127,111 @@ static ASTNode applyStepToNode(NodeFactory* nf, STPMgr& bm,
   return nf->CreateTerm(s.kind, s.outWidth, s.constants[0], in);
 }
 
+// Constant-evaluate one collected step on a known input value.
+static ASTNode evalStepOnConstant(STPMgr& bm, const GroundStep& s,
+                                  const ASTNode& in)
+{
+  std::vector<CBV> args;
+  if (s.samePathAllOperands)
+  {
+    args.push_back(in.GetBVConst());
+    args.push_back(in.GetBVConst());
+  }
+  else if (s.kind == BVSX || s.kind == BVZX)
+  {
+    // The evaluator takes the new width from outWidth.
+    args.push_back(in.GetBVConst());
+  }
+  else
+  {
+    const size_t arity = s.constants.size() + 1;
+    size_t ci = 0;
+    for (size_t i = 0; i < arity; i++)
+    {
+      if (i == s.pathIndex)
+        args.push_back(in.GetBVConst());
+      else
+        args.push_back(s.constants[ci++].GetBVConst());
+    }
+  }
+  return bm.CreateBVConst(NonMemberBVConstEvaluator(s.kind, args, s.outWidth),
+                          s.outWidth);
+}
+
+/* Global polarity of a boolean node within the live mutable tree: the
+ * union, over every occurrence and every path to the root, of the
+ * negation parity along it. POL_POS alone means the formula is monotone
+ * non-decreasing in the node (making it true can only help), POL_NEG
+ * the dual. ITE conditions, XOR/IFF operands and anything unrecognised
+ * get both. Only pure single-polarity predicates admit the one-sided
+ * collapse. */
+static const unsigned POL_POS = 1;
+static const unsigned POL_NEG = 2;
+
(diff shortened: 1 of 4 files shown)


lib/Simplifier/RemoveUnconstrained.cpp
```diff
@@ -127,6 +127,111 @@ static ASTNode applyStepToNode(NodeFactory* nf, STPMgr& bm,
...
+
+/* Global polarity of a boolean node within the live mutable tree: the
+ * union, over every occurrence and every path to the root, of the
+ * negation parity along it. POL_POS alone means the formula is monotone
+ * non-decreasing in the node (making it true can only help), POL_NEG
+ * the dual. ITE conditions, XOR/IFF operands and anything unrecognised
+ * get both. Only pure single-polarity predicates admit the one-sided
+ * collapse. */
+static const unsigned POL_POS = 1;
+static const unsigned POL_NEG = 2;
```
aytey, line 169: Are we worried about having multiple "types" of polarity; my array_ex branch also adds one: https://github.com/aytey/stp/blob/5cd0c6fe560bb77a134adc5e5bfb73cdccc4e8fc/lib/AbsRefineCounterExample/AbstractionRefinement.cpp#L40-L45

TrevorHansen:
Rarely useful.

PR #694, opened 2026-07-29
README: build in parallel, make the python example copy-pasteable

TrevorHansen:
Two small documentation fixes:

* The build instructions now use `cmake --build . -j$(nproc)` (and `make -j$(nproc)` in the testing section). A serial build spends most of its time compiling ABC's ~1000 files, and the plain `cmake --build .` is what most people copy-paste from the quick install.

* The python interface example is now a plain script with the outputs shown as comments, instead of an IPython transcript whose `In [n]:` prompts have to be stripped before it can be run. It also gets a `python` fence for syntax highlighting. Verified the snippet runs as written against an installed build; the model comment says "e.g." since any satisfying assignment may be returned.

### DIFF
--- a/README.markdown
+++ b/README.markdown
@@ -65,8 +65,8 @@ Overflowing a 32-bit integer using the python interface:
 import stp
 
 s = stp.Solver()
-x = s.bitvec('x')
-y = s.bitvec('y')
+x = s.bitvec('x', width=32)
+y = s.bitvec('y', width=32)
 s.add(x + y < 20)
 s.add(x > 10)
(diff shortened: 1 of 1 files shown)


README.markdown
```diff
@@ -230,7 +229,7 @@ pip install lit
 mkdir build
 cd build
 cmake -DENABLE_TESTING=ON ..
-make
```
aytey, line 233: This should actually be `cmake --build . -j$(nproc)` in case the generator _isn't_ Makefiles

PR #696, opened 2026-07-29
Collapse predicates over unconstrained chains against symbolic terms

TrevorHansen:
## Summary

- The ground-path collapse required the predicate at the top of an unconstrained chain to face a **constant**; a census across the 4,661-file hard set counted 1,755 bails against symbolic terms (`(= (bvand x c) t)`, `(bvugt (zext x) t)`, `(bvsgt (bvmul x c) t)`, ...).
- The predicate is now rewritten into its **invertibility condition** — a predicate over the other side `t` alone — joined with a fresh boolean, with `x` defined to realise either truth value:
  - `(= (bvand x c) t)  ==>  (and (= (bvand t (bvnot c)) 0) b)`, `x := ite(newP, t, fallback)`
  - `(bvugt (f x) t)  ==>  (and (bvugt M t) (or b (bvugt m t)))`, `x := ite(newP, x_M, x_m)`
- **Equalities** use a pseudo-inverse walk whose conditions state exactly that `t` lies in the chain's image. Exactness forces a shape rule: steps with non-unique preimages (and/or/extract/urem/udiv/shifts) are admitted only at the bottom of the chain, where their input really is the free variable — composed higher, their picked preimage can exit the lower chain's image while another preimage stays inside, silently losing satisfying assignments. Bijective steps (xor/plus/sub/neg/not) and injective-under-condition steps (concat-with-constant, sign/zero extension) compose anywhere.
- **Comparisons** are monotone in the chain value, so only the image's true extremes in the predicate's order matter; they come from exhaustive enumeration of variables up to 12 bits (the population is dominated by 8-bit variables). The under-approximating sample mode cannot drive this: an under-approximated extreme would force the rewritten predicate false where the original was satisfiable.
- If the other side holds an unconstrained variable of its own, the rewrite **defers**: an elimination from that side collapses the predicate to a bare boolean, strictly better than copying `t` into a condition (which would destroy that side's unconstrainedness — caught by `simplification-tests/propagate.smt2`).
- The rewrite is a pointwise function equality — `P(f(x_def), t) == newP` for every valuation of `t`'s variables and `b` — so it holds in any polarity and under shared predicates, exactly like the fresh-boolean EQ rule, and composes with the ITE distribution frames.

## Impact, measured before review

- **Fires 526 times across 63 of 4,661 hard-set files** (EQ over one step 460 — almost all `(= (bvand x c) t)`; comparisons 44; EQ under ITE frames 12; a 5-step signed-comparison chain 10).
- Wall-clock on the 63 firing files: zero answer disagreements; performance-neutral within the sat-instance noise floor. Two apparent slowdowns (bench_9668 4.7s→timeout, bench_16157 1.8→12.2s) are solver lottery, not structure: on both, the new CNF is *smaller* (e.g. 2,228,926 vs 2,228,991 clauses), bench_9668 times out for **both** binaries under `--simplifying-minisat`, and symmetric wins of the same size appear on other files (3.0→1.8, 3.4→1.9).

## Testing

- Twelve new `RemoveUnconstrained_SymbolicSide` tests with exhaustive back-substitution equivalence over every assignment: the and-mask equality, a bijective+concat chain, urem/lshr/extract bottoms, term-on-the-left, unsigned/signed comparisons over enumerated extremes, ITE-frame distribution, and refusals for lossy-above-lossy, a degenerate mask, and a shared chain. 78/78 suite, full ctest 69/69 (Debug + assertions).
- 2,500-file differential sweep vs master: 2500 agree, 0 disagree, 0 skipped; the rule fires in 191 of those files and all 191 pass `-d` counterexample checking in the assertions build.


TrevorHansen:
Pushed the BVMULT extension to the walk (c33fa0bf), with honest impact numbers measured before review:

- **Mechanism**: an odd constant multiplication is a bijection (modular inverse via `Simplifier::MultiplicativeInverse`) and may sit anywhere on the chain; an even constant `c = odd * 2^k` images exactly the multiples of `2^k`, so like the shifts it is bottom-only, with condition `t[k-1:0] = 0` and witness `x := inv(odd) * (t >> k)`.
- **Yield: zero.** Fires on neither the 4,661-file hard set (526 fires in 63 files, shape-for-shape unchanged from the previous census) nor the 2,500-file fuzz corpus. Equalities over constant multiplications are solved earlier by the wordlevel solver, which owns the invertible case; what reaches this pass are the comparison shapes, already served by the enumerated-extremes path.
- Soundness unchanged: 82/82 suite (four new tests, including refusal of a non-bottom even multiplication and multiplication by zero), full ctest, and a fresh 2,500-file differential sweep vs master with zero disagreements.

So this commit is completeness (the walk no longer has an arbitrary gap at BVMULT, and the tests document the even-multiplication image argument) rather than yield; dropping it from the PR is a defensible call.

PR #698, opened 2026-07-29
cbitp: write derived constants back into the formula

TrevorHansen:
Constant bit propagation regularly proves interior nodes constant under the assumption that the top is true, but the write-back rejected any fact whose defining proposition already appeared inside the top — and the proposition usually folds straight back to the very conjunct the information came from (`EQ(node, const)` simplifies to the asserted equality), so almost nothing was ever written back.

Dropping that rejection alone turns out to be a no-op: every top-level conjunct is itself a boolean node fixed to true, so each conjunct is erased by the replacement (`conjunct → TRUE`) and then restored verbatim by the conjoined fact, masking the term-level substitutions underneath it.

Instead, this collects the boolean facts first and rewrites them with the discovered term-level constants before conjoining:

- **Term facts** are conjoined verbatim. They are built from the original (pre-replacement) nodes and conjoined after the replacement runs over the top, so `top ⟺ top[n:=c] ∧ (n=c)` holds whether or not the equality already appears inside the top.
- **Boolean facts** are rewritten with the term constants before being conjoined. This is justified because the term facts are conjoined verbatim, and a term fact is never written into its own defining equality. A boolean fact the constants don't change, and that the top already contains, is left in place — same as the old guard, avoiding erase-and-restore churn.
- The conjoined facts are sorted by expression number and deduplicated, so the output doesn't depend on the fixed-bits map's iteration order.

### Example

```smt2
(assert (not (bvsgt #x00000000 x)))
(assert (= ((_ extract 31 31) x) #b0))
(assert (not (= x (bvneg (bvadd x (bvmul x (concat ((_ extract 31 1)
   (bvadd #xFFFFFFFF (concat (_ bv0 31) ((_ extract 31 31) x)))) #b0)))))))
(assert (= #b1111111111111111111111111111111 ((_ extract 31 1)
   (bvadd #xFFFFFFFF (concat (_ bv0 31) ((_ extract 31 31) x))))))
```

cbitp fixes bit 31 of x from the second assert and derives that the whole `bvadd/extract/concat` chain is constant, making the multiplier `#xFFFFFFFE`. Previously all of that was discarded and the problem went to the SAT solver. Now the constants are written in, and the word-level simplifications finish the job: `x + x·#xFFFFFFFE = -x`, the double negation cancels, and the query collapses to false before bit-blasting.

### Validation

- 2500-file fuzzed-corpus differential sweep against the previous build: all answers agree, no skips.
- Assertion-enabled build (with counterexample construction) clean over all 2500 corpus files plus the motivating instance.
- Coarse timing over 300 corpus files within noise.
- `answer_diff` over the 40-file real-benchmark corpus is still running; I'll post the result as a comment.

Node counts can grow slightly on some inputs (~5% post-pipeline on one Sydr benchmark, neutral elsewhere sampled) because the conjoined facts substitute only term constants while the top substitutes everything, so rewritten copies share less structure. The existing difficulty-score reversion still guards the net-harm case.

### DIFF
--- a/lib/Simplifier/constantBitP/ConstantBitPropagation.cpp
+++ b/lib/Simplifier/constantBitP/ConstantBitPropagation.cpp
@@ -33,8 +33,10 @@ THE SOFTWARE.
 #include "stp/Simplifier/constantBitP/ConstantBitP_MaxPrecision.h"
 #include "stp/Simplifier/constantBitP/ConstantBitP_TransferFunctions.h"
 #include "stp/Simplifier/constantBitP/ConstantBitP_Utility.h"
+#include <algorithm>
 #include <fstream>
 #include <iostream>
+#include <vector>
 
 using std::endl;
 using std::cout;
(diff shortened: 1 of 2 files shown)


TrevorHansen:
Pushed a second commit that restructures the fact rewriting per review discussion: instead of rewriting the boolean facts with a separate term-constants map, everything — the top and every fact — is rewritten with the full replacement map, sharing one cache. A fact's own map entry can't erase the fact because its rewrite starts at the node's **children**, and a node is never its own descendant. This gives "full map minus self" semantics with a single shared memo, restores DAG sharing (the ~5% node growth is gone), and lets boolean facts discharge into each other as well.

Numbers: the motivating example now leaves 15 nodes after the first cbitp pass (was 17 with the first commit, 23 unchanged before); on the Sydr `predicate_593` benchmark the first pass leaves 26242 nodes vs 48691 pristine (the first commit had regressed this to 51635).

Also added `tests/query-files/simplification-tests/cbitp_writeback.smt2`: run with `--exit-after-CNF`, so it only answers `unsat` if the simplifications alone collapse the query — the pristine binary prints nothing.

Revalidated on the new commit: 2500-file differential sweep all agree (0 skips), assertion+counterexample build clean over 800 corpus files, motivating instance and sat spot-checks give valid models.

PR #702, opened 2026-07-31
Seed the ground-path square root without overflowing

aytey:
#680 added `isqrt64` to seed the ground-path collapse's backward hint chain. Its Newton seed `(n + 1) / 2` overflows to zero at `n == UINT64_MAX`, and the next iteration then divides by it.

`hintBackStep` calls it for a `(bvmul t t)` step of width 64, so any chain whose back-propagated target is all-ones takes a SIGFPE inside `RemoveUnconstrained`. That needs nothing but plain QF_BV — no floating point, no arrays — and crashes master today, 2.4.0 included:

```smt2
(set-logic QF_BV)
(declare-fun x () (_ BitVec 64))
(assert (= (bvmul (bvlshr x #x0000000000000001) (bvlshr x #x0000000000000001)) #xffffffffffffffff))
(check-sat)
```

```
$ stp --SMTLIB2 crash.smt2 ; echo $?
136
```

(136 is a hardware trap, so there is no assertion text.)

Fixed here by seeding with `n / 2 + (n & 1)`: the same `ceil(n/2)` without the overflow. Every smaller `n` keeps the seed it had, so no existing result — or the CNF built from it — moves, and `UINT64_MAX` now returns the exact root `2^32 - 1`. Checked against the old function over ~11M inputs.

Two regression tests, both SIGFPE without the fix: a unit test that back-propagates all-ones through a width-64 square and checks the recovered root is really used, and `unconstrained-ground-path3.smt2`.

Found by a murxla campaign that was fuzzing a downstream floating-point branch; the trace was FP-heavy but minimises to this bit-vector core, and the crashing code is FP-free.

### DIFF
--- a/lib/Simplifier/AchievableImage.cpp
+++ b/lib/Simplifier/AchievableImage.cpp
@@ -183,11 +183,15 @@ uint64_t isqrt64(uint64_t n)
 {
   if (n == 0)
     return 0;
-  uint64_t x = n, y = (x + 1) / 2;
+  // The Newton seed is ceil(n/2), written so that it can't overflow:
+  // (n + 1) / 2 wraps to zero at n == UINT64_MAX, and the first
+  // iteration then divides by it. Every smaller n gets the same seed,
+  // so no other input's result moves.
+  uint64_t x = n, y = n / 2 + (n & 1);
   while (y < x)
   {
     x = y;
-    y = (x + n / x) / 2;
+    y = (x + n / x) / 2; // x >= isqrt(n) here, so n / x <= x: no overflow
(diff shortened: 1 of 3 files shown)


aytey:
Might be worth a 2.4.1 maybe, given this is ~~likely~~ a regression.

TrevorHansen (approved):

TrevorHansen:
> Might be worth a 2.4.1 maybe, given this is ~likely~ a regression.

Nice work finding this. Yes, agreed it's a good idea to do a new release.

PR #720, opened 2026-08-02
Move the bit-blasted-encoding comparison into propagator_bench

TrevorHansen:
Replaces `tools/measure_constantbitprop` with `--bcp-check` on `propagator_bench`, and extends it into an arc-consistency checker for the bit-blasted encoding. 16 files, +810/−568.

Five commits:

| | |
| --- | --- |
| `5e976df4` | move the comparison into `propagator_bench`, delete the old tool |
| `b27550f0` | `--cnf` — choose how the CNF being propagated over is generated |
| `1a593b04` | **bug fix**: variable elimination was hiding the bits being counted |
| `103268f3` | `--bcp-exhaustive` — arc consistency, contradictory cases included |
| `8cc4283f` | fix a dangling reference gcc 13 rejects |

## The question it preserves

`measure_constantbitprop` asked something no other harness asks. STP bit-blasts to CNF and calls a SAT solver anyway, so a word-level propagator only earns its keep by fixing bits that boolean constraint propagation over the same circuit would not have fixed on its own.

That is a *different* question from the one `propagator_bench` already answers. The precision verdict asks whether a **better** propagator could exist — the reference is the ideal. This asks whether the one we have is **worth running** — the reference is the practical alternative.

So it becomes a third oracle for the cbitp domain rather than a fourth domain. `--bcp-check N` encodes `op(children) = result` once per row, then per case asserts the known bits as unit clauses and counts what comes out fixed at decision level zero. `--bcp-budget` caps it; the `ns/call` column stays the propagator's.

## Why the old tool only ever measured bvsge

Its `Functions` table had every entry but `bvsge` commented out, and `BBAsProp.h` could only express two same-width children and one result. Three things had to change:

* **The ArrayTransformer runs before bit-blasting.** `sbvdiv`, `sbvrem` and `sbvmod` are rewritten first, so encoding them the old way would have hit an unhandled kind.
* **Bits that never reached the CNF** carry a sentinel rather than a variable number in the map `addVariables()` builds, and must be skipped rather than asserted.
* **Structural children** (extract bounds, extend widths) are constants in the encoded node, so they are neither asserted nor counted. That also lets `extract` and the extends be checked here, which `maxPrecision()` cannot do.

It also drops a coupling: `BBAsProp` dereferenced `GlobalSTP->arrayTransformer`, which is why #717 had to patch `measure.cpp`. `Bcp.cpp` builds its own `ArrayTransformer` like `maxPrecision()` does.

## The bug, and what it invalidated

`getFixedCountWithAssumptions()` called CryptoMiniSat's `simplify()` before adding the assumptions, and `simplify()` runs **bounded variable elimination**. An eliminated variable never appears in `get_zero_assigned_lits()`, so a bit that unit propagation *does* imply read as not deduced whenever BVE had removed its variable.

The test that exposes it: `bvand` at width 32 with every input bit known and the output unknown. Unit propagation over arc-consistent AND clauses **must** fix all 32 output bits. It reported 27 with plain Tseitin and 30 with the default CNF; with `set_no_bve()`, all six settings report 32.

The shortfall always lands on the BCP side, so it reads exactly like a propagator advantage. **`measure_constantbitprop` had the same bug through the same function**, so its published 17.17x for `bvsge` was wrong for as long as the tool existed.

## Results, width 16, 50% fixed, both-ways

| op | BCP bits | cbitp bits | ratio |
| --- | --- | --- | --- |
| `bvsrem` | 0.685 | 2.560 | 3.74x |
| `bvsmod` | 0.480 | 1.685 | 3.51x |
| `bvsdiv` | 2.460 | 6.490 | 2.64x |
| `bvurem` | 1.675 | 2.875 | 1.72x |
| `bvadd` | 3.235 | 4.340 | 1.34x |
| `bvsub` | 3.345 | 4.325 | 1.29x |
| `bvult` | 0.195 | 0.245 | 1.26x |
| `bvashr` | 8.115 | 10.165 | 1.25x |
| `bvshl` | 8.945 | 10.005 | 1.12x |
| `bvlshr` | 9.005 | 9.960 | 1.11x |
| `bvmul` | 1.470 | 1.565 | 1.06x |
| `bvudiv` | 6.380 | 6.435 | 1.01x |
| `bvand`, `bvor`, `bvxor`, `bvnot` | — | — | **1.00x** |
| `eq`, `ite`, `concat`, `extract`, both extends | — | — | **1.00x** |

And `bvsge` at width 65, the old tool's configuration: 0.17 vs 0.21 bits, **1.2x** at 50% fixed (not 17.2x).

**Most operations are at or near 1.0x** — for those the word-level propagator adds nothing the SAT solver would not have found; it earns its keep by letting word-level simplification see the constants earlier, before any CNF exists. The signed division family is the only large margin.

## Arc consistency

`--bcp-check` samples, and every case it draws is built from a solution, so it never asks whether propagation *detects a contradiction*. `--bcp-exhaustive W` does both halves: every combination of fixed and unfixed bits, contradictory ones included, against a brute-forced ideal. A row is `yes` only if nothing is left underived, no contradiction is missed, and nothing is fixed that the ideal does not fix.

`getFixedCountWithAssumptions()` now takes a `conflict` out-parameter to report this — `add_clause()` propagates a unit at level zero as it adds it, so its `false` return is exactly unit propagation deriving the empty clause.

For `bvand` at width 4 — all 531441 partial assignments, of which **336960 have no solution** — every one of the six `--cnf` settings is arc-consistent: 0 incomplete, 0 missed conflicts. About 45s per setting.

The negative control, at width 3 with the default CNF, is what makes that worth anything:

| op | verdict | incomplete | missed conflicts |
| --- | --- | --- | --- |
| `bvand`, `bvxor` | yes | 0 | 0 |
| `bvadd` | **NO** | 820 | 12 |
| `bvmul` | **NO** | 730 | 128 |
| `bvudiv` | **NO** | 3220 | 61 |
| `bvult` | **NO** | 162 | 0 |

Exactly the expected split: bit-independent operations are arc-consistent, carry chains and global relations are not.

## `--cnf`

Picks the generator — `simple` (plain Tseitin) or the five `cnf_effort` levels — and the row reports clause and variable counts so size and strength read together. For `bvand` the setting changes the encoding size by **2.1x** (959 clauses against 451 at width 64) and its propagation strength **not at all**: all six deduce identically at every width, which is what the arc-consistency of the Tseitin AND clauses predicts. It is a size/time knob, not a propagation knob.

## Also removed

`BBAsProp.h` and `StopWatch.h`, whose last user this was. `include/stp/Util/` no longer holds any test scaffolding.

## CryptoMiniSat

`getFixedCountWithAssumptions` is a `CryptoMiniSat5` method, so `Bcp.cpp` compiles to a stub without it and both options are refused at the command line rather than quietly reporting nothing:

```
$ propagator_bench --bcp-check 10     # in a -DNOCRYPTOMINISAT=ON build
propagator_bench: --bcp-check/--bcp-exhaustive need a build with CryptoMiniSat (configure with -DNOCRYPTOMINISAT=OFF)
$ echo $?
1
```

## Testing

Clean configure with the `Extra tools` CI flags: warning-free under `WERROR`, `ctest` 70/70. Also built with `-DNOCRYPTOMINISAT=ON` for the stub path. Text, CSV and HTML output all carry the new figures.

Full write-ups in `bench-hard/reports/2026-08-02-cbitp-vs-bitblasted-propagation.md` and `2026-08-02-cnf-effort-vs-propagation.md`.


TrevorHansen:
Two follow-up commits, one of which is a correction to the numbers in the PR description above.

**`--cnf` (b27550f0)** lets `--bcp-check` choose how the CNF it propagates over is generated — `simple` (plain Tseitin) or the five `cnf_effort` levels — and reports clause/variable counts so size and strength read together.

**`set_no_bve()` (1a593b04) is a bug fix, and it invalidates the ratios in the description.** `getFixedCountWithAssumptions()` called CryptoMiniSat's `simplify()` before adding the assumptions, and `simplify()` runs bounded variable elimination. An eliminated variable never appears in `get_zero_assigned_lits()`, so a bit that unit propagation *does* imply read as not deduced whenever BVE had removed its variable.

The test that exposes it: `bvand` at width 32 with every input bit known and the output unknown. Unit propagation over arc-consistent AND clauses must fix all 32 output bits. It reported 27 (Tseitin) and 30 (default); with BVE off, all six settings report 32.

The shortfall always lands on the BCP side, so it looks exactly like a propagator advantage. Corrected figures at width 16:

| op | was | is |
| --- | --- | --- |
| `bvsdiv` | 39.3x | 2.64x |
| `bvudiv` | 26.8x | 1.01x |
| `bvand` | 1.6x | 1.00x |
| `bvsge` (width 65) | 17.2x | 1.2x |

The real picture: most operations are at or near **1.0x** — `bvand`, `bvor`, `bvxor`, `bvnot`, `eq`, `ite`, `concat`, `extract` and both extends are identical to three decimals, and `bvudiv`/`bvmul` are within rounding of it. The signed division family is the only large margin (`bvsrem` 3.74x, `bvsmod` 3.51x, `bvsdiv` 2.64x); addition and the shifts sit at 1.1x–1.35x.

Note this means **`measure_constantbitprop` had the same bug**, through the same function — its published 17.17x for `bvsge` was wrong for as long as the tool existed. Reproducing that number during the port looked like evidence the measurement had survived; it was evidence the bug had.

It also erased the apparent differences between CNF settings, since each encoding gave BVE different opportunities. With it fixed, all six settings encode `bvand` equally well — byte-identical to the propagator at every width — which is what the arc-consistency of the Tseitin AND clauses says must happen.

Full write-ups in `bench-hard/reports/2026-08-02-cnf-effort-vs-propagation.md` and the corrected `2026-08-02-cbitp-vs-bitblasted-propagation.md`.

PR #727, opened 2026-08-02
rewrite_rule_gen: stop the search overflowing the stack, and say what it does

TrevorHansen:
The unbounded search — the tool’s headline mode, what it does with no arguments — **segfaulted**. It recursed to depth 23771 and ran out of stack. This fixes that, and makes the tool tell you what it is doing.

## The crash

`findRewrites()` recursed three ways. Two were tail calls — *narrow the list by one counterexample, start again, and return* — and the list shrinks by about one element each time, so those alone went tens of thousands of frames deep. The third splits the list into equivalence buckets and recurses into each, and a narrowed restart re-enters that split, so converting only the tail calls just moved the growth: it then died at **depth 38730** instead.

The search now carries its own stack (`pending`), so the depth of STP’s no longer depends on the size of the function list.

## What the crash was hiding

The search did not terminate. It cycled on the same nine expressions forever:

```
depth:6541292, size:9 values:0   pairwise pass finds a counterexample
depth:6541293, size:9 values:1   "Split into 1 pieces"   nothing separated
depth:6541294, size:9 values:0   the same nine again
```

A restart *replaced* the distinguishing assignments with the single newest one, so every split was as coarse as the first, and a group that one assignment cannot separate never got separated.

They are accumulated now, and a split yielding a single bucket keeps its assignments and goes on to the pairwise pass rather than performing the identical split again. The pairwise pass contributes another assignment, and only then is splitting retried.

**An unbounded run now proceeds instead of crashing.** Being straight about the limits: it still finds no rules on the default function list within nine minutes, with the assignment set grown to 11902 — the group is not being separated even so. That is a further defect in the search which this does **not** fix and which I have not traced.

## Regression check

Against a binary built from master:

| mode | result |
| --- | --- |
| `unit-test` | byte-identical output |
| `test` | byte-identical output |
| `verify test-rules.smt2` | byte-identical output, 4 rules verified |
| `generate 5 3` | same end state (0 rules, same summary); differs only in the progress counters, which the restructure necessarily changes |

## Saying what it does

| | was | now |
| --- | --- | --- |
| `--help` / `-h` / `help` | silently exited 0 | prints the modes; there was no usage text at all |
| unrecognised mode | fell through every branch, exited 0 — a typo looked like success | prints the usage, exits 1 |
| `rewrite`, `write-out` with no rules | exited 0 having truncated `rules_new.smt2` to nothing | fail with a message, and write nothing |
| `load_new_rules()` | fell back to stdin silently, which on a terminal looks like a hang | names the file it read, or says it is falling back |

```
$ rewrite_rule_gen genrate 5 3
rewrite_rule_gen: unrecognised mode genrate 5 3

usage: rewrite_rule_gen [mode [arguments]]
...
$ echo $?
1
```

## Testing

Builds warning-free under `WERROR` with `BUILD_EXTRA_TOOLS=ON`; `ctest` 70/70. The four invocations in the `Extra tools` CI job all still exit 0.

### DIFF
--- a/tools/rewrite_rule_gen/rewrite_rule_gen.cpp
+++ b/tools/rewrite_rule_gen/rewrite_rule_gen.cpp
@@ -2211,13 +2211,27 @@ void test()
 // Substitutes concrete values for the variables. Every leaf is a constant
 // afterwards, so the factory folds the result -- unless it is missing a fold,
 // which is what this mode hunts, so the caller checks rather than assumes.
-ASTNode evalAt(const ASTNode& n, const ASTNode& vVal, const ASTNode& wVal)
+// The variables the enumeration is over, and a disjoint copy of each. The
+// copies are what makes constancy decidable without guessing a value: n is
+// constant exactly when it agrees with itself over fresh variables.
+vector<ASTNode> mcVars;
+vector<ASTNode> mcFresh;
+
+// Substitutes concrete values for the variables. Every leaf is a constant
+// afterwards, so the factory folds the result -- unless it is missing a fold,
+// which is what this mode hunts, so the caller checks rather than assumes.
+ASTNode evalAt(const ASTNode& n, const vector<ASTNode>& values)
 {
   ASTNodeMap ft;
-  ft.insert(make_pair(v, vVal));
-  ft.insert(make_pair(w, wVal));
+  for (size_t i = 0; i < mcVars.size(); i++)
+    ft.insert(make_pair(mcVars[i], values[i]));
   ASTNodeMap cache;
   return SubstitutionMap::replace(n, ft, cache, nf);
+}
+
+bool isFolded(const ASTNode& n)
+{
+  return n.isConstant() || n == mgr->ASTTrue || n == mgr->ASTFalse;
 }
 
 // Distinct nodes in the DAG; used only to report the smallest findings first.
@@ -2237,33 +2251,32 @@ size_t nodeCount(const ASTNode& n)
   return nodeCount(n, seen);
 }
 
-bool isFolded(const ASTNode& n)
-{
-  return n.isConstant() || n == mgr->ASTTrue || n == mgr->ASTFalse;
-}
-
 // True when n takes the same value at every sample. Wrong only in the safe
 // direction: it can pass a node that is not constant, never reject one that
 // is.
-bool sameAtEverySample(const ASTNode& n,
-                       const vector<std::pair<ASTNode, ASTNode>>& samples,
+bool sameAtEverySample(const ASTNode& n, const vector<vector<ASTNode>>& samples,
                        ASTNode& value)
 {
-  value = evalAt(n, samples[0].first, samples[0].second);
+  value = evalAt(n, samples[0]);
   if (!isFolded(value))
     return false; // not folded even fully applied; not what this looks for
 
   for (size_t i = 1; i < samples.size(); i++)
-    if (evalAt(n, samples[i].first, samples[i].second) != value)
+    if (evalAt(n, samples[i]) != value)
       return false;
   return true;
 }
 
 // n is constant exactly when it agrees with a copy of itself over fresh
-// variables, whatever the two are assigned.
+// variables, whatever they are assigned.
 bool provablyConstant(const ASTNode& n)
 {
-  const ASTNode other = renameVars(n);
+  ASTNodeMap ft;
+  for (size_t i = 0; i < mcVars.size(); i++)
+    ft.insert(make_pair(mcVars[i], mcFresh[i]));
+  ASTNodeMap cache;
+  const ASTNode other = SubstitutionMap::replace(n, ft, cache, nf);
+
   const ASTNode agree = (n.GetType() == BOOLEAN_TYPE)
                             ? nf->CreateNode(IFF, n, other)
                             : nf->CreateNode(EQ, n, other);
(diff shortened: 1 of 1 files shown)


TrevorHansen:
Second commit: `missed-constants`, a mode that produces results today.

The search modes are unproductive on the default function list, so this asks a question the same machinery can already answer. It builds every two-level function and predicate over the two variables — **no constant leaves**, so nothing folds merely because a constant was handed in — and asks of each result the simplifying node factory did *not* reduce to a constant whether it is one anyway.

**105 distinct shapes at 6 bits, in about seven seconds.** The smallest:

```
(bvmul (bvsdiv v v) (bvsdiv v v))   is always #b000001
(bvsdiv (bvsdiv v v) (bvsdiv v v))  is always #b000001
(bvsmod v (bvsdiv v v))             is always #b000000
(bvsrem v (bvsdiv v v))             is always #b000000
(bvurem v (bvsdiv v v))             is always #b000000
(= #b000000 (bvsdiv v v))           is always false
(= (bvnot v) (bvmul v v))           is always false
(bvand (bvsdiv v v) (bvshl v v))    is always #b000000
(bvlshr (bvshl v v) (bvneg v))      is always #b000000
```

Division is total, so `(bvsdiv v v)` is 1 or -1 and never 0 — most of these turn on that. `(bvnot v) = (bvmul v v)` would need `v² + v + 1 ≡ 0`, and `v² + v` is always even, so it never holds at any width.

Operator spread across the 105: `=` 68, `bvsdiv` 63, `bvmul` 24, `bvshl` 22, `bvnot` 20, `bvurem`/`bvsrem`/`bvsmod` 16 each.

## How it decides

Constancy is *"n agrees with a copy of itself over fresh variables, for every assignment"* — no candidate value has to be guessed, and one formulation covers both terms and predicates. Concrete evaluation at thirteen assignments runs first and rejects all but a few percent, so the solver is called ~1400 times rather than 26500.

Reporting is keyed on the expression with the variables collapsed together, since `(op v v)` and `(op w w)` are one finding — that is 650 raw hits down to 105 distinct shapes — and sorted smallest first, since those are the ones worth teaching the factory. Output is SMT-LIB2; the internal node printing is unreadable at this size.

## Checked, not assumed

Five findings verified against the `stp` binary rather than the tool’s own solver call — all five negations unsat, and a non-constant control (`(bvmul v v) = 1`) correctly sat. Expressions the factory *already* folds, `(bvsub v v)` and `(bvxor v v)` among them, are correctly absent from the output.

`ctest` 70/70, warning-free under `WERROR`, and the four `Extra tools` CI invocations still exit 0.

TrevorHansen:
## Third commit: the first fix the tool found — and it is a solver change

**Flagging this explicitly because it is a `SimplifyingNodeFactory` semantic change sitting in a PR that is otherwise about a tool.** Happy to split it out if you would rather review it separately.

`SBVMOD` has had an `x smod x → 0` rule for a long time. `SBVDIV` never got the matching one, and *every* rule in its case is guarded on one of the children being constant, so `(bvsdiv v v)` went through untouched — and nothing downstream could tell that the quotient is never zero.

```
(bvsdiv x x)  ->  (ite (= x 0) ones 1)
```

Division is total: `x / x` is 1 wherever the division is a real one, and the remaining case is `0 / 0`, which takes the value the neighbouring rule already gives for `x / 0` — zero is not negative, so all ones. At width 1 it collapses again, both branches being `0b1`.

### Soundness

Checked against a binary built **without** the rule — necessarily, because with the rule in place both sides of the equivalence fold to the same node and the check proves nothing. That circularity caught me first time round.

| width | `(bvsdiv v v) = (ite (= v 0) ones 1)` negated |
| --- | --- |
| 1, 4, 6, 8, 16, 32 | unsat at every one |

Control: the deliberately wrong claim `(bvsdiv v v) = 1` everywhere comes back **sat**, as it should.

### Effect, regenerated

| | before | after |
| --- | --- | --- |
| distinct constant-but-unfolded shapes | 105 | **47** |
| shapes involving `bvsdiv` | 63 | **0** |
| expressions enumerated | 26500 | 23913 |
| runtime | 7.4s | 5.7s |

One rule accounts for more than half the findings. The ones it takes with it are all consequences — `(bvsmod v (bvsdiv v v))`, `(bvsrem v (bvsdiv v v))`, `(bvmul (bvsdiv v v) (bvsdiv v v))`, `(= 0 (bvsdiv v v))`.

### What is left, smallest first

```
(= (bvnot v) (bvmul v v))              is always false
(bvlshr (bvshl v v) (bvneg v))         is always #b000000
(bvshl  (bvshl v v) (bvnot v))         is always #b000000
(bvugt  (bvshl v v) (bvneg v))         is always false
(bvand (bvnot v) (bvand v w))          is always #b000000
(bvlshr (bvmul v v) (bvmul #b000010 v)) is always #b000000
```

Three families now: `v² + v + 1 ≡ 0` is unsatisfiable because `v² + v` is always even; a `(bvshl v v)` cluster where the shift is out of range; and `(bvand (bvnot v) (bvand v w))` — `x & ~x` is presumably folded directly but **not through a nested `bvand`**, which looks like the next easy one.

`ctest` 70/70, warning-free under `WERROR`.

TrevorHansen:
Third commit: `missed-constants [variables] [arity]`, defaulting to 4 and 3.

The n-ary kinds — `bvadd`, `bvand`, `bvor`, `bvxor` — are now built at every arity up to the limit, so `(bvand x y z)` is reached. They are commutative and associative, so only non-decreasing index tuples are generated; the other orderings are the same node and the factory would hash them together anyway.

Sampling scales with the variable count: every variable at each of zero, one and all-ones, then each variable alone at each of those with the rest zero, then twelve random assignments. The corners are where the shift and division edge cases live, so they are there explicitly rather than by hoping the random draws land on them.

**Behaviour at the old settings is unchanged** — `missed-constants 2 2` reports the same 105 shapes over the same 26,500 expressions, which is the check that the generalisation did not alter anything.

## What each dimension buys

| variables | arity | expressions | wall | shapes |
| --- | --- | --- | --- | --- |
| 2 | 2 | 26,500 | 7s | 105 |
| 2 | 3 | 90,126 | 21s | **155** |
| 3 | 2 | 146,210 | 23s | 105 |
| 4 | 3 | 6,451,913 | **1759s** | **155** |

**The arity is what pays; the variables buy nothing at all.** Three children adds 50 shapes. Four variables adds none — and not merely the same count: canonicalising the variable names and comparing the two sets, the 155 shapes at 4 variables are *exactly* the 155 at 2.

On reflection that is what one would expect. A two-level expression that is constant has little use for a variable it can only mention once. But it costs 72× and 1.2GB to find out, so a large run now says so before it starts:

```
note: 6451913 expressions. Measured: 4 variables at arity 3 takes about half an hour
and 1.2GB, and finds exactly what 2 variables at arity 3 finds in twenty seconds.
```

**The default is 4 3 as asked, but on this evidence I would make it 2 3** — same findings, twenty seconds instead of half an hour. It is a one-word change; say the word and I will flip it.

PR #729, opened 2026-08-02
SimplifyingNodeFactory: x & ~x is zero however the two are nested

TrevorHansen:
The second rule `rewrite_rule_gen`’s new `missed-constants` mode (#727) turned up.

`handle_bvand()` spots the complementary pair only among the immediate children of one `BVAND`, so these both survived as expressions even though the conjunction is the same one either way:

```
(bvand (bvnot x) (bvand x y))                   is always 0
(bvnot (bvand x (bvand (bvnot x) (bvnot y))))   is always ones
```

The local is called `flat_children`, but nothing flattens.

## The rule

The conjuncts are gathered through nested `BVAND`s before the existing scan, and a node appearing alongside its own negation anywhere in that set annihilates the whole conjunction.

Three things keep it cheap and safe:

- The walk runs **only when a child is itself a `BVAND`**, which is not the common shape — `BVOR` is lowered to `~(~x & ~y)`, so an OR contributes a `BVNOT` child rather than a `BVAND` one.
- It is bounded at depth 8. Giving up early costs a rewrite that does not fire, never a wrong answer.
- It can only ever return the annihilator, so **a node it does not fire on comes out exactly as it did before**.

The set-based test is also symmetric where the existing scan is not: it asks of each element whether it is a `BVNOT` whose operand is present, rather than relying on `x` having sorted ahead of `~x`.

## Verification

Soundness was checked against a binary built **without** the rule — necessary, because with it in place both sides of the equivalence fold to the same node and the check proves nothing.

| claim | baseline | with rule |
| --- | --- | --- |
| `(bvand ~x (bvand x y)) = 0` | unsat | unsat |
| `~(x & (~x & ~y)) = ones` | unsat | unsat |
| `(bvand ~x (bvand y (bvand x y))) = 0` — three levels | unsat | unsat |
| `(bvand ~x (bvand y y)) = 0` — **must not** be annihilated | sat | sat |

Re-running `missed-constants 2 2`, exactly those two shapes disappear and nothing else changes — I diffed the full 105-shape sets before and after, not just the counts. `ctest` 70/70, warning-free under `WERROR`.

## One thing it surfaced

With those two gone, the diff shows one shape newly reaching the report: `(bvsrem q q)` where `q` is `(ite (= 0 x) ones 1)`. `SBVREM` *does* have a `children[0] == children[1]` rule, so this is the same nesting-versus-structure gap in a different operator — the rule fires on a syntactic self-pair but not once the operand is a folded ITE. Not addressed here.

Companion to #728, which fixes `(bvsdiv x x)`; the two are independent and neither is stacked on the other.


TrevorHansen:
Reworked the bound after being asked whether descending 8 levels was too much. It was arbitrary, and measuring showed it was wrong — but in the *other* direction.

Instrumenting the factory over 400 files from `tests/query-files`:

| | |
| --- | --- |
| BVAND creations | 899 |
| with a BVAND child (walk triggered) | 189 — **21%** |
| conjuncts collected per walk | **6 on average** |
| deepest nesting reached | **32 levels** |
| times the rule fired on that corpus | 0 |

So nested BVANDs are neither rare nor shallow, and the walk is cheap: six nodes on average, because the deep cases are chains rather than wide trees. **A depth limit of 8 would have stopped a quarter of the way down the deepest chains** — giving up exactly where the deep nesting is, while saving nothing worth saving.

Depth was the wrong axis. The cost that needs containing is a wide tree, so the bound is now on how many conjuncts have been gathered (64, an order of magnitude above the observed average). Since every BVAND has at least two children, that budget bounds the recursion as well as the work, so there is no unbounded descent either.

The instrumentation is not in the commit. It also confirmed the rule *fires* rather than my inferring it from a solver result — `fired=1`, `conjuncts=3`, on `(bvand (bvnot x) (bvand x y))`.

Re-verified after the change, against a binary built without the rule:

| claim | baseline | with rule |
| --- | --- | --- |
| `(~x & (x & y)) = 0` | unsat | unsat |
| `~(x & (~x & ~y)) = ones` | unsat | unsat |
| three levels deep | unsat | unsat |
| **ten levels deep** — past the old cap | unsat | unsat |
| `(~x & (y & y)) = 0` — must **not** fire | sat | sat |

`ctest` 70/70.

TrevorHansen:
Measured this properly on the 4661 hard problems (the non-incremental QF_BV set), instrumenting `handle_bvand` to record, whenever a complementary pair exists, the shallowest depth at which it is reachable. 4660 of the 4661 reported; the runs used `--exit-after-CNF` so the factory is exercised through parsing, simplification and bit-blasting without paying for the solve.

| | |
| --- | --- |
| walks (a BVAND with a BVAND child) | 1,907,715 |
| walks finding a complementary pair | 65 |
| deepest nesting encountered | 60 |

**Where the pair actually is:**

| shallowest depth that catches it | count |
| --- | --- |
| 1 | 35 |
| 2 | 27 |
| 3 | 1 |
| 7 | 1 |
| 8 | 1 |

So the answer to "was there a closer pair" is **usually yes — 95% are within two levels — but not always**. Two of the 65 genuinely need seven and eight levels. The 60-deep nesting is real, and just once it carries an annihilator eight levels down.

That kills both of the positions I had held. My earlier claim that a depth limit "would give up exactly where the deep nesting is" was rhetoric — I had measured traversal depth, not pair depth, and the rule had fired zero times on the corpus I used. And the 3-level cap I was about to substitute would have dropped 3% of the firings.

It also means a depth limit is the wrong instrument: to lose nothing it would have to be set at **8**, and 8 has no meaning beyond being the deepest pair this particular corpus contains. So the bound stays on conjuncts gathered (64) — it bounds the cost directly, bounds the recursion too since every BVAND has at least two children, and sits comfortably past every pair observed.

Instrumentation is not in the commit. One trap worth recording: hard problems hit the timeout, and SIGTERM skips the static-destructor flush, so the counters have to be written from a signal handler or the data silently disappears for exactly the biggest files.

Re-verified after the change — `ctest` 70/70, and against a binary built without the rule: both target shapes unsat, a ten-level nesting unsat, and `(~x & (y & y))` correctly still sat.

TrevorHansen:
Answering the OR question, and adding the checks. Rebased onto master now that #727 and #728 have landed.

## Boolean AND / OR — already handled, and deliberately left alone

`CreateSimpleAndOr()` **already has** a nested-complement check: a `nested_same_kind` flag, then a pass that treats a same-kind child's children as contributing to the same conjunction. So there is no gap here of the sort `handle_bvand` had. But it descends exactly **one** level, so the same question applies — is one enough?

Instrumented it over the same 4661 hard problems:

| | |
| --- | --- |
| AND/OR creations | **165,026,767** |
| with a same-kind child | 11,631,286 |
| with a complementary pair anywhere in the conjunction | **223** |
| max nesting | 60 |

| shallowest depth that catches the pair | count |
| --- | --- |
| 1 | 222 |
| 2 | 1 |

**222 of 223 are reachable in one level — exactly what the existing check does.** One case in 4660 files sits at depth two. Deepening a path taken 165 million times to gain one fold is not a trade worth making, so the boolean side is unchanged and the tests below simply pin the behaviour that is there.

The contrast with `BVAND` is worth stating: that path is walked 1.9M times, not 165M, and its pair depths run to eight. Same rule, different answer, because the cost and the tail are different.

## BVOR — covered already, and now proved to be

`BVOR` is lowered to `~(~a & ~b)` at creation, so no `BVOR` node ever reaches the factory: every bitvector OR is a `BVAND` question by the time the rule sees it, and the double negations cancel so a nested `x | ~x` presents as a nested `x & ~x`. Nothing further was needed — but I checked rather than assumed, and the `bvor_*` tests below **fail without the rule and pass with it**, which is the proof.

## The checks

Nine tests in `SimplifyingNodeFactory_Test.cpp`, asserting the equality the rule should make true. The parser builds through the simplifying factory, so a node reaching `ASTTrue` is one the factory *folded* — a satisfiability check could not tell that apart from a claim that merely happens to hold, and every claim here holds either way.

Verified in both directions. With the rule removed, exactly the five bitvector cases fail:

```
[  FAILED  ] bvand_complement_nested_once
[  FAILED  ] bvand_complement_nested_other_order
[  FAILED  ] bvand_complement_nested_deeply        <- eight levels
[  FAILED  ] bvor_complement_nested_once
[  FAILED  ] bvor_complement_nested_deeply         <- eight levels
[  PASSED  ] 4 tests    <- the two guards and the two boolean cases
```

The guards (`*_no_complement_is_left_alone`) assert the equality does **not** fold, so a rule that annihilated too eagerly would fail them. The deep cases sit at eight levels, the depth of the deepest annihilating pair actually occurring in the hard problems.

`ctest` 70/70 on the rebased branch.

TrevorHansen:
Yes — and it did **not** work. Good catch; that was a real hole.

## The case

```
A & (v2 & ~A)      with  A = (v0 & v1)
```

`collectConjuncts()` *replaced* a nested `BVAND` with its children, so the set held `v0, v1, v2, ~A` — and `~A`’s operand `A` was never in it, because `A` had been decomposed rather than recorded. The conjunction survived as an expression despite being zero.

Confirmed as a failing test before fixing it, rather than reasoned about:

```
[  FAILED  ] SimplifyingNodeFactory_Test.bvand_complement_of_a_nested_and
```

## The fix

One line: the nested node goes into the set **as well as** being descended into. Both facts are wanted — `A` is a conjunct in its own right, and so is each of its children. Both insertions are sound, since a factor of the conjunction and the factors of that factor are all implied.

## Checks

Three more tests, now twelve in total:

| test | |
| --- | --- |
| `bvand_complement_of_a_nested_and` | the shape above |
| `bvor_complement_of_a_nested_or` | the same through the OR lowering |
| `bvand_nested_and_without_its_negation` | the guard |

Reverting just the one-line change fails exactly the first two and leaves the other ten passing.

**The guard deserves a note.** The obvious spelling, `(v0 & v1 & v2 & ~(v0 & v2))`, is useless: I tried it first and the solver says it is *always zero* — just not for a reason this rule can see — so it would pass whether or not the rule were sound. It is now `(v0 & v1 & v2 & ~(v3 & v4))` over disjoint variables, checked satisfiably non-zero, so a rule that fired on it would be **wrong** rather than merely eager.

`ctest` 70/70. PR is now +245/−0 across the factory and the tests.

TrevorHansen:
Done — and worth doing, though **not** for the reason one might hope.

The check was building an `ASTNodeSet` per call: 1.9M of them over the hard problems, each allocating to hold six elements on average. It is now a fixed stack array of node numbers packed with a polarity — `x` as `2n`, `~x` as `2n+1` — so a literal and its negation differ only in the low bit and sorting brings them adjacent. Finding a pair is one pass looking for neighbours that differ by one with the even one first; equal neighbours are duplicates rather than a pair.

That is the same shape as the duplicate removal a few lines below, which already sorts and inspects neighbours, so it reads as house style rather than a new idea. `MAX_CONJUNCTS` was already the bound on what the walk gathers, so 64 entries cannot overrun.

## Whether it is faster: essentially no

Measured with instruction counts, which are deterministic, over eight hard problems run to CNF generation:

| | instructions |
| --- | --- |
| hash set | 8,364,696,841 |
| stack array | 8,362,253,241 |
| | **−0.029%** |

Per file the deltas run −0.11% to +0.02%, **in both directions** — so at whole-program level this is a wash, and I am not going to dress 0.03% up as a win. The BVAND path is not where the time goes.

The reasons to prefer it are that it allocates nothing and has fewer moving parts. If you would rather keep the set for readability that is a defensible call; the measurement does not argue either way.

One methodology note: the first A/B I set up compared two `stp` binaries and reported them byte-identical, because with `BUILD_SHARED_LIBS=ON` the change lives in `libstp.so` and the executable never changes. The comparison above swaps the *library* via `LD_LIBRARY_PATH`, which is the thing that actually differs.

Behaviour unchanged: the twelve folding tests pass, and removing the rule still fails exactly the seven that test it. `ctest` 70/70.

PR #732, opened 2026-08-03
PropagateEqualities: dense set for tagNodes' shared-node detection

TrevorHansen:
Follow-up to #730, same validation. `tagNodes` inserts every node shared between candidate right-hand sides into an `std::unordered_set<ASTNode>`; the set is then copied out and sorted by expression number before use, so its iteration order never reaches any decision. Switching it to `ankerl::unordered_dense` (same container #730 used for the variable-id sets) removes the chained-bucket traffic from the shared-node detection pass.

Verified byte-identical CNF against master over the same 40-file corpus (39 identical, 1 baseline-NOCNF skip); full test suite passes.

### DIFF
--- a/include/stp/Simplifier/PropagateEqualities.h
+++ b/include/stp/Simplifier/PropagateEqualities.h
@@ -30,6 +30,7 @@ THE SOFTWARE.
 #include "stp/Simplifier/Simplifier.h"
 #include "stp/Simplifier/PropagateEqualities.h"
 #include "stp/Simplifier/NodeSimplifier.h"
+#include "extlib-unordered-dense/ankerl/unordered_dense.h"
 
 /* 
   Finds formulae asserted at the top level, and removes the variables, e.g:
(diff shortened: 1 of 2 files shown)


TrevorHansen:
Measured this in isolation against its merge base (three-arm interleaved A/B alongside #733, plus deterministic instruction counts):

- Variable Elimination is faster in all 9 interleaved pairings, but only by 1–2.5% (e.g. 12.46s → 12.32s median on laby_22_22_17).
- Whole-run wall time is within the noise floor (±1%, one file slightly worse).
- Retired instructions are +0.3% to +0.7% across five instances — expected for a chained-buckets → open-addressing swap (more ALU per probe, fewer memory stalls), which is why wall time is the honest metric.

So: a small, consistent stage-level win plus consistency with the other dense-set uses in this file, but nowhere near #730/#733's effect. No objection to closing instead of merging if it doesn't feel worth the churn.

PR #753, opened 2026-08-04
RemoveUnconstrained: eliminate fp.gt over an unconstrained float

TrevorHansen:
An unconstrained float under an `fp.gt` was being lowered to its full SymFPU circuit — around 160 word-level nodes and 2200 CNF clauses at binary64 for a single `(fp.gt x c)` — even though the query needs no SAT call at all. RemoveUnconstrained runs before FloatBlast, while the comparison is still one source node and the variable still occurs in it exactly once; afterwards the variable feeds its unpack circuit many times over and stops looking unconstrained. This adds the pass's first floating-point rule, so such comparisons are decided at the word level the way `BVGT` ones already are.

The rule mirrors the neighbouring `BVGT` case, with IEEE's extremes as witnesses: NaN makes any ordered comparison false, and an infinity of the right sign makes `fp.gt` true whenever any value can. Against a constant, the predicate is replaced by a fresh boolean `v` with `x := ite(v, +-oo, NaN)` recorded — `+oo` when the variable is on the left, `-oo` on the right — so model construction falls out of the recorded ite as usual. When both sides are unconstrained the witness pair is `(+oo, +0)` / `(NaN, NaN)`.

Two deliberate refusals: the always-false extremes (`x > +oo`, `x > NaN`, `-oo > x`, `NaN > x`) are left for the blasted circuit to collapse, exactly as the `BVGT` rule leaves its own always-false cases; and formats the blaster refuses are skipped, so that refusal stays loud instead of being quietly eliminated away.

One wrinkle: FP constant folding is deferred solver-wide, so a literal operand usually arrives as `to_fp`'s three-child reinterpret form over constant bits rather than as an interned constant — only `(fp ...)` literals intern at parse. The rule resolves that form locally through the `CreateFPConst` funnel (which also canonicalises NaN payloads, keeping the node-identity comparisons sound) without rewriting the operand or changing what the parser builds.

Tested with two new lit tests (the satisfiable shapes in both literal spellings plus the both-unconstrained form, and the four always-false extremes as separate queries), and exhaustively at the 8-bit format (3, 5): all 256 constants on either side of an unconstrained variable, elimination on versus off, with no disagreements; the emitted witness was also re-checked through the elimination-off path. The full query-files suite passes.

If this shape looks right, `fp.lt`/`fp.leq`/`fp.geq`, `fp.eq` and the classification predicates admit the same treatment with the same witness analysis.

### DIFF
--- a/lib/Simplifier/RemoveUnconstrained.cpp
+++ b/lib/Simplifier/RemoveUnconstrained.cpp
@@ -37,6 +37,7 @@ THE SOFTWARE.
 #include "stp/Simplifier/RemoveUnconstrained.h"
 #include "stp/AST/MutableASTNode.h"
 #include "stp/Extensionality/ExtensionalityContext.h"
+#include "stp/FloatBlaster/FloatBlaster.h"
 #include "stp/Simplifier/AchievableImage.h"
 #include "stp/Simplifier/constantBitP/Dependencies.h"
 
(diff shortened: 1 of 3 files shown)


TrevorHansen:
Hi @aytey, Here's a first go at implementing unconstrained variable elimination for a floating-point predicate. Is this how you are imaging us handling it? Cheers.

aytey:
> Hi @aytey, Here's a first go at implementing unconstrained variable elimination for a floating-point predicate. Is this how you are imaging us handling it? Cheers.

Yes, this looks great -- well done for working out the true/false witnesses; they seem to correct to me 😄

PR #754, opened 2026-08-04
SimplifyingNodeFactory: mirror the FP less-thans onto the greater-thans

TrevorHansen:
The factory already converts every bit-vector less-than into a swapped greater-than (`BVLT -> BVGT` and friends), which is why downstream passes only ever needed rules for the greater-than kinds. The floating-point comparisons had no such canonicalisation, so anything consuming FP comparisons faces four kinds where two suffice.

`fp.lt(a, b)` is `fp.gt(b, a)` exactly, NaN included — both mean "ordered and strictly ordered", so the swap is a pure mirror — and likewise for `fp.leq`/`fp.geq`. This rewrites the less-thans at node creation. Unlike the total bit-vector order, this is as far as FP comparisons collapse: `not(fp.lt)` is geq-*or-unordered*, so the four kinds reduce to two, not one.

The x-vs-x rules keep their meaning through the mirror (`x < x` reaches the `FP_GT` arm and still folds to false), and FloatBlast's `FP_LT`/`FP_LEQ` arms stay as ground truth, since the hashing factory (`--disable-simplifications`) still builds those kinds verbatim.

Testing: `test_fprewrites` gains four checks — the mirror fires structurally, and agrees with the un-normalised original on every operand pair of the exhaustive test format. A new lit test asserts the xor of each comparison with its mirror is unsat, twice over: once under `--exit-after-CNF` through the simplifying factory (the xor operands intern as one node and collapse before any circuit exists, so the verdict prints only while the rule fires), and once under `--disable-simplifications` (the SAT solver proves the semantic identity over the full circuits). The full query-files suite passes.

A follow-on benefit: once this lands, the unconstrained-variable elimination for `fp.gt` (#753) automatically covers `fp.lt` spellings too — `fp.lt x c` normalises to `fp.gt(c, x)`, which is the arm that rule already handles; a future `FP_GEQ` arm would pick up `fp.leq` the same way.

### DIFF
--- a/lib/NodeFactory/SimplifyingNodeFactory.cpp
+++ b/lib/NodeFactory/SimplifyingNodeFactory.cpp
@@ -614,15 +614,32 @@ ASTNode SimplifyingNodeFactory::CreateNode(Kind kind, const ASTVec& children)
         result = NodeFactory::CreateNode(kind, children[0][0]);
       break;
 
-    // x < x and x > x are false, NaN included.
+    // Mirror the less-thans onto the greater-thans, as the bit-vector
+    // comparisons are above (BVLT -> BVGT): fp.lt(a, b) is fp.gt(b, a)
+    // exactly, NaN included -- both mean "ordered and strictly ordered", so
+    // the swap is a pure mirror -- and likewise for the non-strict pair.
+    // Downstream simplification then meets only the greater-than forms and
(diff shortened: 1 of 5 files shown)


aytey (approved):
LGTM -- I like it!

PR #757, opened 2026-08-05
Bit-blast fp.gt/fp.lt over leaf operands natively

TrevorHansen:
fp.gt and fp.lt whose operands are leaves — symbols, interned constants, or to_fp's reinterpret form over constant bits — no longer expand to SymFPU circuits. FloatBlast leaves the source comparison in place and the bit-blaster encodes it directly over the packed IEEE bits (`BBcompareFP`).

The encoding: sign-magnitude maps onto an unsigned total order with a per-bit XOR against the sign — `key(f) = ~sign ++ (bits[w-2:0] ^ sign)` — so the whole comparison is the XORs, two field tests per operand (NaN, zero) and one existing `BBBVLE` comparator. The keys misorder exactly one pair, `fp.gt(+0, -0)`, whose keys are adjacent with nothing between; a both-zero conjunct repairs it. isNaN tests the exponent/significand fields, so arbitrary NaN payloads compare as NaN.

**Why it wins**: the expensive part of the SymFPU route for a leaf comparison is unpacking the operands (CLZ priority encoder + barrel shifter per operand for subnormal normalisation); the comparison over unpacked fields is nearly free. Comparing packed bits needs no unpacking at all. Measured on binary64 probes (same binary, flag A/B):

| query | config | SymFPU (vars/clauses) | native | ratio |
|---|---|---|---|---|
| `fp.gt x 1.3` | `--disable-simplifications` | 1557 / 4475 | 377 / 935 | 4.1x / 4.8x |
| `1.3 < x < 2.0` (to_fp literals) | default | 607 / 2291 | 119 / 295 | 5.1x / 7.8x |
| same | `--disable-simplifications` | 1589 / 4571 | 379 / 941 | 4.2x / 4.9x |

**The gate** (`FloatBlast::nativeComparisonSurvivor`) keeps this sound and non-regressing:

- Operands that are FP operations stay on the SymFPU path: their values are cached *unpacked*, and packing them costs more than the nearly-free unpacked comparison saves.
- A literal usually arrives as to_fp's three-child reinterpret form over constant bits, not as an interned constant (FP constant folding is deferred solver-wide). The gate resolves that form through the canonicalising `CreateFPConst` funnel — the same lookthrough `RemoveUnconstrained`'s comparison rule uses. Without it the common spelling of the motivating query never fires.
- A comparison of two constants must not survive: constant folding and model evaluation lower the node over its constant children and rely on the result collapsing to a constant (`ComputeFormulaUsingModel` asserts lowering changed the node).

Because the surviving node and its children are exactly the original well-sorted ones, nothing is rebuilt over packed carriers, so the format-stamping hazard documented in FloatBlast.h does not arise. And because FP_GT/FP_LT reached every downstream pass before lowering existed, they all have arms already — no pass outside FloatBlast and the bit-blaster changes.

The old behaviour stays selectable as `--bb.fp-native-cmp=false`; the flag is deliberately not cleared by `--disable-simplifications`, which now also exercises the native path.

**Verification**

- Exhaustive differential at format (3, 4): all 128x128 constant pairs for both kinds, native AIG blast vs SymFPU constant evaluation — 32768 checks, 0 mismatches (`FloatBlast_Test`).
- New lit tests pin: NaN with any payload on either side is never greater; no zero exceeds another zero (kills the both-zero conjunct if it regresses); native fp.gt agrees with SymFPU fp.leq on all ordered binary32 pairs (SAT-proved cross-encoding identity); the to_fp lookthrough fires end-to-end; and a sat model that `CheckCounterExample` validates through SymFPU's semantics. Each runs under default flags, `--disable-simplifications` and `--bb.fp-native-cmp=false`.
- Full suites green with `-DENABLE_FLOATING_POINT=ON` (101/101, asserts on) and `=OFF` (74/74; the new bit-blaster case compiles without SymFPU and is unreachable there).

Known non-win, accepted: when *other* SymFPU operations share the comparison's operands, the all-SymFPU form can be smaller (the unpack amortises across consumers). Gating on that would make emission depend on traversal order, so it is not attempted.

Follow-ups: fp.geq/fp.leq (the zero correction flips to an OR), float-array READ and ITE operands in the gate, and a pack-carrier node to lift the gate for arithmetic-result operands.

### DIFF
--- a/include/stp/STPManager/UserDefinedFlags.h
+++ b/include/stp/STPManager/UserDefinedFlags.h
@@ -126,6 +126,10 @@ public:
   bool bvplus_variant = true;
   bool conjoin_to_top = false;
 
+  // Bit-blast fp.gt/fp.lt over already-packed operands natively (sign-magnitude
+  // comparison of the IEEE bits) instead of via the SymFPU unpacking circuits.
+  bool fp_native_cmp = true;
+
   int64_t multiplication_variant = 1;
 
   // If the bit-blaster discovers new constants, should the term simplifier be
--- a/include/stp/ToSat/BitBlaster.h
+++ b/include/stp/ToSat/BitBlaster.h
@@ -193,6 +193,9 @@ class BitBlaster
   // Return bit-blasted form for BVLE, BVGE, BVGT, SBLE, etc.
   BBNode BBcompare(const ASTNode& form, BBNodeSet& support);
 
+  // bit blast a floating-point comparison (FP_GT/FP_LT) over packed operands
+  BBNode BBcompareFP(const ASTNode& form, BBNodeSet& support);
+
   // Return bit-blasted form for the overflow predicates BVUADDO, BVSADDO,
   // BVUMULO, BVSMULO, BVUSUBO, BVSSUBO.
   BBNode BBOverflow(const ASTNode& form, BBNodeSet& support);
--- a/lib/FloatBlaster/FloatBlast.cpp
+++ b/lib/FloatBlaster/FloatBlast.cpp
@@ -107,6 +107,56 @@ private:
                  "source sorts");
   }
 
+  // A comparison operand that is (or resolves to) a leaf the lowering
+  // leaves untouched: a symbol or an interned constant. FP constant folding
+  // is deferred solver-wide, so a literal usually arrives as to_fp's
+  // three-child reinterpret form over constant bits; resolve it through the
+  // canonicalising funnel (CreateFPConst), the same lookthrough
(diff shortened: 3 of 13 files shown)


aytey:
> fp.gt and fp.lt whose operands are leaves — symbols, interned constants, or to_fp's reinterpret form over constant bits — no longer expand to SymFPU circuits. FloatBlast leaves the source comparison in place and the bit-blaster encodes it directly over the packed IEEE bits (BBcompareFP).

I read the diff; AWESOME!

PR #758, opened 2026-08-05
Parse real constants under to_fp, exactly rounded at parse time

aytey:
# Parse real constants under `to_fp`, exactly rounded at parse time

STP was the odd one out: bitwuzla, cvc5 and z3 all accept SMT-LIB real constants under `to_fp`, and STP refused them with the curated "real literals are not supported" error. After this, every spelling the QF_FP-family logics admit parses and means the same bits it does everywhere else:

```
$ cat pi.smt2
(set-logic QF_FP)
(declare-const x (_ FloatingPoint 8 24))
(assert (= x ((_ to_fp 8 24) RNE (/ 355 113))))
(assert (fp.gt x ((_ to_fp 8 24) RTZ 3.14)))
(check-sat)
$ stp pi.smt2
sat
```

The fold happens at parse time, bitwuzla-style: the constant becomes the same interned float constant the packed-bits spellings produce — no Real sort, no new node kinds, nothing new for the rest of the solver to know about. Three commits: the dependency, the feature, the tests; each builds standalone.

## The dependency

The conversion is done by [LibBF](https://bellard.org/libbf/) (Fabrice Bellard, MIT), obtained the way minisat and CaDiCaL are — a `scripts/deps` setup script run before configuring — because upstream publishes release tarballs, not a git repository.

- **`scripts/deps/setup-libbf.sh`** downloads the pinned release (2025-06-03, sha256-checked; `LIBBF_TARBALL` names a pre-downloaded copy for offline builds), takes only the library proper out of the tarball, applies the MSVC portability patch, and builds `libbf.a` with `-fPIC`. `LIBBF_NO_BUILD` stops it after the portable fetch/verify/patch half, for platforms that compile differently.
- **cmake** gains `USE_LIBBF`/`LIBBF_DIR` in the CaDiCaL mould: off by default, fatal if requested but missing. The header is consumed through a SYSTEM include path — its `__int128` typedefs are rejected by gcc `-pedantic` in C++, the same reason ABC and SymFPU are `-isystem`.
- **The MSVC patch** (`scripts/deps/patches/0001-libbf-msvc-compat.patch`; a failed hunk on a version bump means it needs rebasing) gives `cutils.h` a cl.exe branch — `_BitScanReverse`/`_BitScanForward` for the clz/ctz builtins, `#pragma pack` for the packed structs, the printf-format attribute behind a macro — and keeps every Windows compiler on the 32-bit limb path: clang-cl defines `__SIZEOF_INT128__` but has no 128-bit division runtime when targeting MSVC (bellard/quickjs#14); cl.exe never defines it and lands there anyway.
- **Without `USE_LIBBF`** STP builds exactly as before and keeps refusing real constants, with the message now pointing at the setup script.

## What folds

- **Every spelling**: decimals (`0.1`), bare numerals (`1`), negated values (`(- 1.5)`), and the rational form (`(/ 1 3)`, with decimal components as in `(/ 1.2 3.25)`, negatable as `(- (/ p q))` or `(/ (- p) q)`). A small `an_real_constant` nonterminal covers the shapes; `-` and `/` reach the parser as plain symbols, so the operator and its arity are checked in the action, and the grammar stays conflict-free (`%expect 0`).
- **One rounding, ever.** A decimal is read exactly and rounded once by `bf_atof` — precision, IEEE exponent range, subnormals, per-mode overflow all in one call. A rational shifts each component's fractional places onto the other side as powers of ten, parses the two exact integers, and rounds once with `bf_div`. STP's side (`lib/FloatBlaster/DecimalLiteral.cpp`) only unpacks LibBF's mantissa and exponent into the packed fields — `bf_get_float64`'s layout, generalised to any width. No rounding logic lives in STP.
- **Symbolic rounding modes fold too**: the value is computed under every mode and becomes an if-then-else over the five constants — the shape bitwuzla builds — while a value every mode rounds identically (anything exactly representable) collapses to the one constant with no ite at all.

Three rules worth review attention:

- **The real zero has no sign.** `(- 0.0)` and `(/ 0 9)` convert to +zero, as SMT-LIB requires — the minus is stripped before LibBF can turn it into -zero — while an inexact underflow of a *negative* value still gives -zero, which is IEEE. There are tests on both sides of that line.
- **The ite's fall-through is sound.** The last mode's conversion is the innermost else, with no validity test, because a RoundingMode term is pinned to the five legal encodings when it is declared — no sixth case exists to fall into.
- **Numerals are capped at machine range**, as everywhere else in this parser (the lexer delivers a numeral's value, not its text), so a bare-integer constant beyond 2^64 would be misread; decimals and rationals carry full-precision digit strings and are unaffected.

## Refusals

Everything refused says what to write instead, with a line number: a zero denominator (`(/ 1 0)` names no real — bitwuzla, for contrast, feeds it to GMP unchecked and crashes), nested ratios, exponent width 2 (SMT-LIB allows it; LibBF's flag encoding starts at 3), exponent widths past `LIMB_BITS - 3`, and the rm-less form `((_ to_fp e s) 1.5)`, which now says a rounding mode is what is missing.

## Testing

**102/102 ctest and the full lit suite in both configurations** — with LibBF (18 folding query files + 16 gtest cases run) and without (the `REQUIRES: !libbf` refusal test runs, everything gated skips).

- **Differential, three independent implementations**: bitwuzla (MPFR), cvc5 `--fp-exp` (SymFPU) and z3, over nine formats from `(3,5)` to `(15,113)`, all five modes, ~90 constants across every spelling — exact values, halfway ties (2^24+1), the 2^-25 subnormal tie, underflow, overflow. 3735 of 3960 cases agree bit-for-bit; the 225 skips are bare numerals no oracle accepts under `to_fp`, whose values are covered via the same digits spelled with `.0`. The full corpus ran a second time with a free RoundingMode pinned by assertion, driving the ite chain.
- **Host oracles in the unit tests**: strtof/strtod for decimals under RNE, and hardware division for rationals — which taught its own lesson, pinned in a comment: for `1/16777217` the float divide rounds the operand first, and the engine is right to disagree, so that leg only runs when both components are float-exact.
- **CI enforces the feature everywhere floating point exists**: the main Linux matrix and both CaDiCaL jobs directly; `gcc (i386)` in its container, the one job exercising LibBF's 32-bit limb build (`LIMB_BITS = 32`, `BF_EXP_BITS_MAX` = 29), where the full suites run and pass; and Windows, which builds `bf.lib` with `cl /std:c11` (vswhere + Enter-VsDevShell, no third-party action) and answers two real-literal smoke queries through it. Only `gcc (no-fp)` stays without, by definition — which leaves the `REQUIRES: !libbf` test to local builds, a trade made knowingly: coverage of the feature everywhere over coverage of its absence.

All 14 checks are green on this series.


TrevorHansen:
> `((_ to_fp e s) RNE 1.5)` used to die with the curated "real literals are not supported" refusal,

Make sense now why we need a parser!

PR #772, opened 2026-08-06
Make the MiniSat backend optional, behind -DUSE_MINISAT (default OFF)

aytey:
## What

MiniSat becomes an ordinary optional backend behind `-DUSE_MINISAT` (default OFF), on the same pattern Riss and CaDiCaL already follow. Configure now requires at least one SAT backend (CaDiCaL, CryptoMiniSat, MiniSat, or Riss) and fails with a message naming the options when none is enabled.

## Why

MiniSat was the one mandatory backend: `find_package(minisat)` was a configure-time `FATAL_ERROR`, libminisat sat on the unconditional link line, and zlib was required purely because MiniSat's headers include it. Builds whose configured solver is CaDiCaL or CryptoMiniSat carried all of that for a backend they never run.

## How

The series is five commits, each of which builds and passes the full test suite on its own:

1. **`SATSolver` owns its literal type.** `SATSolver::vec_literals` was `Minisat::vec<Minisat::Lit>` and `mkLit` built a `Minisat::Lit`, so every consumer of the solver abstraction compiled against MiniSat's headers. It is replaced with an STP-owned `Lit` carrying the same `variable*2 + sign` encoding and a `vec_literals` over `std::vector`. `var`/`sign`/`toInt` become `SATSolver` statics, so the unqualified calls in the CryptoMiniSat and Riss wrappers resolve via the base class exactly where they used to resolve into namespace `Minisat` by ADL. The MiniSat-family wrappers now convert into `Minisat::vec` themselves, like every other backend already did.
2. **RunTimes stops using MiniSat's `System.h`.** Replaced with `getrusage` on POSIX and `GetProcessTimes`/`GetProcessMemoryInfo` on Windows. This fixes a real bug along the way: the old code divided a figure MiniSat already returns in megabytes by 1024², so "Peak Memory Used" has printed roughly zero for years (and on Windows MiniSat's `memUsed()` is a hardcoded 0 anyway).
3. **Backend construction moves into a factory.** `stp::createSATSolver(const UserDefinedFlags&)` is now the only place that turns `solver_to_use` into a backend; `ConstantBitP_MaxPrecision` stops hard-wiring `MinisatCore` and runs on whatever backend the build configured.
4. **The flip.** `USE_MINISAT` option (OFF by default), zlib gated with it, the at-least-one-backend check, `--minisat`/`--simplifying-minisat` compiled out of builds that cannot honour them, and `vc_supportsMinisat`/`vc_useMinisat`/`vc_isUsingMinisat` (plus the simplifying variants) answering for the build honestly instead of being hardcoded to true. Their symbols and the `ifaceflag_t` enum values are unchanged, so existing embedders still link and old integer flags still mean what they did. The test suite learns the same conditionality: a `minisat` lit feature gates the one query-file that drives `--simplifying-minisat`, the gtest suites that construct `MinisatCore` directly or select it through the C API are gated or skip, and the Python solver tests drop the assumption that MiniSat is always present. `rewrite_rule_gen` and `propagator_bench` now declare the backend each actually drives (MiniSat and CryptoMiniSat respectively).
5. **CI and README.** Every CI job that already built MiniSat passes `-DUSE_MINISAT=ON`, so each keeps exactly the coverage it had (including `--minisat` in the static release binaries). The cadical job stops setting MiniSat up and becomes the configuration that proves STP builds and passes its tests with no MiniSat anywhere -- which also retires its rpath-link/`LD_LIBRARY_PATH` install-check workarounds, since a static CaDiCaL leaves no solver shared library outside the install tree. The README presents the backends as a choice with CaDiCaL recommended.

## Verification

Two local configurations both pass the full 103-test ctest suite (gtest, lit query-files, Python bindings): `USE_MINISAT=ON` plus CaDiCaL, and a CaDiCaL-only build with no MiniSat available. In the latter, `--minisat` is rejected by the CLI, `ldd` shows no libminisat or libz, and the gated lit test reports as unsupported rather than failing. The no-backend configure fails with the expected message.

## Downstream note

Builds that today rely on MiniSat being picked up implicitly (distro packages, embedders that install only minisat) will need to add `-DUSE_MINISAT:BOOL=ON`. Worth a line in the release notes when this ships.

### DIFF
--- a/.github/actions/build-static-linux/action.yml
+++ b/.github/actions/build-static-linux/action.yml
@@ -70,12 +70,12 @@ runs:
   # builds CryptoMiniSat into a static-linkable prefix, and the only one that
   # needs neither gtest nor OutputCheck.
   #
-  # Only deps/install: both setup scripts install self-contained CMake
-  # packages there -- archives and configs, with no path pointing back at
-  # deps/cms or deps/minisat -- so the source and build trees are worth
-  # nothing after the install and several hundred megabytes to carry. On a
-  # hit the step below is skipped and they are not wanted; on a miss they are
-  # cloned again. This is what every other job caches.
+  # Only deps/install: the setup script installs a self-contained CMake
+  # package there -- archives and configs, with no path pointing back at
+  # deps/cms -- so the source and build trees are worth nothing after the
+  # install and several hundred megabytes to carry. On a hit the step below
+  # is skipped and they are not wanted; on a miss they are cloned again.
+  # This is what every other job caches.
   #
   # Shared between the build types below, which differ only in how STP itself
   # is compiled: the dependencies are identical either way. Two jobs racing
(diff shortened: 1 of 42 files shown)


aytey (commented):
~~Uses my old name; needs to be fixed~~

Done! Name corrected

PR #777, opened 2026-08-07
Add common sub-sum extraction and pair extraction, enabled by default with flattening

TrevorHansen:
Adds two new simplifications and turns them on by default, together with the existing `--flattening`, which they depend on.

## Common sub-sum extraction (`--common-subsum`)

A `BVPLUS` is hash-consed as a whole node, so two n-ary additions that share operands share no substructure: `(a + b + c)` and `(a + c + d)` each build their own adder chain even though both contain `(a + c)`. The new pass factors the shared part out so that adder is built once.

Addition is associative and commutative modulo 2^w, so the regrouping is value preserving and the pass needs no bit-level information at all. It greedily extracts the operand pair occurring in the most sums — the usual adder-network CSE heuristic — and repeats, so larger shared sub-sums are found by composition.

Nested additions hide these opportunities, because `(a + (b + c))` and `(b + (a + d))` share `{a, b}` but no node. Flattening is what makes the operand sets comparable. On a 517 file sample of hard QF_BV queries the extractable adder count rises from **261 (5 files) to 29,124 (130 files)** once flattening has run, and the pass fires on 4 files without flattening versus 114 with it. It is close to inert on its own, which is why `--flattening` is enabled alongside it.

## Pair extraction (`--pair-extract`)

An independent rule in `StrengthReduction`: if two addends' possibly-one bits are disjoint they cannot carry into each other, so their sum is their bitwise-or, and peeling such a pair out of an n-ary sum removes an adder stage. Unlike the existing whole-node rule it needs bit information for only the two addends being paired, so it still fires when the other operands are unknown — which is most of the time, since only about 30% of addends have known bits.

## Measurements

517 file stride sample of a hard QF_BV set, CNF clause counts over the 476 files that reach CNF in every configuration, relative to the previous defaults:

| configuration | clauses | vs previous default |
| --- | --- | --- |
| previous default | 490,392,262 | — |
| `--common-subsum` only | 490,328,070 | -0.01% |
| `--flattening` only | 486,672,000 | -0.76% |
| `--flattening --pair-extract` | 486,082,739 | -0.88% |
| `--flattening --common-subsum` | 485,004,262 | -1.10% |
| all three (this PR's default) | 484,415,189 | **-1.22%** |

Pair extraction contributes the same -0.121% whether added on top of flattening or on top of flattening plus common sub-sum extraction, so the two new passes are independent rather than competing for the same opportunities.

The two new passes are also orthogonal to `--cnf-generation-effort`: raising that to `very-high` is worth -10.99% on its own, and the three options still contribute -1.37% on top of it, for -12.21% combined. They act at different layers — the effort setting changes only the AIG to CNF mapping, while these change the circuit handed to it.

## Caveats

This is a high-variance change and the honest summary is that the wins and the losses are both large.

Individual wall clock results on hard instances (same binary, flags toggled, interleaved with the arm order rotated, `--cadical`): several instances drop from tens of seconds to hundredths — 45.38s to 0.02s, 23.34s to 0.03s, 23.95s to 0.02s — and one drops 34.88s to 3.35s. Those first three are flattening's doing rather than the new passes'; only the last needs common sub-sum extraction, where flattening alone gets 16.76s and the composition gets 3.35s.

There are real regressions too. One instance goes 4.49s to 19.75s while its CNF *shrinks* 30.5%, which is worth stating plainly: CNF size is a poor predictor of solve time on this corpus, and the -1.22% above should be read as evidence the transformations do what they claim, not as a speedup.

Of the files whose CNF grows under all three options, 98% already grew under flattening alone; the two new passes account for one file each. The single largest apparent regression attributable to common sub-sum extraction (+7.1%) turns out not to be a circuit regression at all — the AIG node count is identical in both arms, and the difference vanishes at every `--cnf-generation-effort` setting other than the default `medium`, so it is ABC cut-enumeration luck rather than a worse circuit.

## Testing

- Full `ctest` suite passes (102/102) with the new defaults.
- 600 randomly generated queries answer-compared against the previous defaults, no mismatches.
- 250 queries run under `-d` counterexample checking, no failures.
- All three options are cleared by `--disable-simplifications`.


TrevorHansen:
## Wall-clock benchmark: old defaults vs this PR's defaults

90 hard QF_BV instances, sampled by a fresh stride so the set is not the one the CNF measurements were selected on. Same binary with the flags toggled, so there is no build difference. Interleaved with the arm order rotated per file, `--cadical`, 90s cap.

| metric | result |
| --- | --- |
| soundness mismatches | **0** |
| timeouts (90s) | 30 old, 30 new — one file each way |
| commonly solved (59 files) | 519.7s → 524.8s (**+1.0%**) |
| dropping the single biggest mover each way | 445.5s → 447.5s (+0.5%) |
| per-file ratio, 50 files over 0.05s | geomean **1.035**, median **0.994** |
| files >5% slower / faster / within ±5% | 10 / 7 / 33 |

Largest movers:

| | old | new | |
| --- | --- | --- | --- |
| lfsr_008_143_080 | 17.03s | 24.68s | +7.65s |
| picorv32_mutAY_QF_BV_NONINCR | 4.55s | 10.13s | +5.58s |
| nextpoweroftwo512 | 8.42s | 10.93s | +2.51s |
| a3.3.2.asp | 57.24s | 52.56s | −4.68s |
| wire.14.x.14.b.8.a.40_unsat.data | 27.36s | 23.75s | −3.61s |
| edge-matching-width=8-height=8-colours=11 | 84.59s | 81.30s | −3.29s |

One file that previously timed out now solves (test_v5_r10_vr5_c1_s13679) and one that solved now times out (SRLI-SAFE-96-1024).

**Read:** on a representative hard sample this is a wash with a slight negative bias — 3.5% slower on the typical file by geometric mean, marginally faster at the median, and the totals move by about 1%. The large wins reported in the description are real but rare: they were found by selecting instances whose CNF moved most, and none of them appear in this unbiased sample.

This does not support enabling all three by default on aggregate evidence. The measurable benefit is concentrated in a small tail, and the regressions are of comparable size and frequency. Worth deciding explicitly whether the tail is worth the typical-case cost, or whether the passes should ship with the flags present but defaulted off.

One caveat about the corpus itself: a number of files in this "hard" set solve in 0.00–0.02s under default flags, because the classifier requires both the default and the `--disable-simplifications` configurations to be under 5s. The set is therefore not uniformly hard for the default configuration, which widens the noise on any aggregate over it.

TrevorHansen:
## Correctness bug found and fixed

A sweep of the full 4,661 file hard set (rather than the 517 file sample the description's numbers came from) turned up five queries whose CNF grew enormously — the worst by **+2640%**, from 663,170 to 18,172,995 clauses. That turned out to be a genuine correctness bug in the extraction, not a heuristic misfire.

**What went wrong.** An operand repeated k times in an addition produces the same candidate pair C(k,2) times, and each occurrence was being recorded separately. That inflated the apparent sharing, and worse, made the substitution run repeatedly on a single addition: after the first pass had consumed its two operands, later passes found nothing to remove but still appended the shared node. The addition then took on a different value.

On the worst query one addition had an operand repeated roughly 380 times. The pair was recorded 71,253 times for that single addition and the shared node was appended without a matching removal **142,904 times**. The bit-blasted circuit grew from 515,736 AIG nodes to 12,200,500.

**Why the tests missed it.** The node factory folds repeated addends away in most cases, so ordinary queries never hit the path. It takes a wide machine-generated addition to retain them — none of the 600 randomly generated queries, the counterexample checks, or the 102 suite tests produced one. The five affected queries all still returned the correct sat/unsat answer by luck, since a corrupted addition can easily remain satisfiable.

**The fix** records each distinct pair once per addition, and locates both operands before removing either so an addition that lacks one is left alone rather than rewritten. An extraction that ends up applying to fewer than two additions is abandoned, since it would add an adder without removing one anywhere.

Results on the five affected queries, old CNF versus new:

| query | before the fix | after the fix |
| --- | --- | --- |
| bench_14100 | +2640% | **-0.3%** |
| bench_12608 | +2212% | **-0.4%** |
| bench_5378 | +1301% | **-0.2%** |
| bench_14036 | +308% | **-3.0%** |
| bench_855 | +307% | **-2.9%** |

All five now show the small reduction the pass is supposed to produce. The pass also no longer spends 27 seconds on the worst query.

Re-verified: 102/102 suite tests pass, and 800 randomly generated queries answer-match the previous defaults.

The full-set CNF measurement and a wall-clock comparison restricted to the queries these options actually change are re-running now; the numbers in the description above came from the 517 file sample and will be replaced.

Issue #782, opened 2026-08-08
Float-to-float to_fp takes an unrounded fast path when the format's exponent range shrinks, giving wrong results and a sat/unsat flip


TrevorHansen:
Converting a float to a `(_ FloatingPoint 2 s)` target with `s >= 5` produces wrong results on master (ad58906b). The visible effects are a soundness-level sat/unsat flip and incorrect constant folding. The SymFPU conversion path is the one at fault; no standard format is affected, since the narrowest standard exponent is Float16's 5 bits.

## Reproducers

**1. The same formula answers both ways.** `x` is the smallest (3, 4) subnormal, `z` its RNE conversion into (2, 5):

```smt2
(set-logic QF_FP)
(declare-fun x () (_ FloatingPoint 3 4))
(declare-fun z () (_ FloatingPoint 2 5))
(assert (= x ((_ to_fp 3 4) #b0000001)))
(assert (= z ((_ to_fp 2 5) RNE x)))
(check-sat)
```

Default flags answer `sat` (the conversion constant-folds). `--disable-equality` and `--disable-simplifications` both answer `unsat`, because `x` stays symbolic and the conversion is built as a SymFPU circuit. The same file with a `(_ FloatingPoint 3 5)` target answers `sat` either way.

**2. Constant folding rounds the wrong way.** 3/32 is exactly halfway between the (2, 5) subnormals 1/16 and 2/16, so RNE must pick the even significand `0010`:

```smt2
(set-logic QF_FP)
(declare-fun z () (_ FloatingPoint 2 5))
(assert (= z ((_ to_fp 2 5) RNE ((_ to_fp 3 4) #b0000011))))
(check-sat)
```

`-p` reports `(fp #b0 #b00 #b0001)`, the odd neighbour.

## Scope

Every (3, 4) input converted under RNE into a range of targets, adjudicated against exact rational arithmetic (no floating point in the checker):

| target | wrong |
|---|---|
| (2, 5) | 38 / 112 |
| (2, 6) | 30 / 112 |
| (2, 4) | 0 / 112 |
| (3, 4) | 0 / 112 |
| (3, 5) | 0 / 112 |
| (4, 5) | 0 / 112 |

So it is not 2-bit exponents as such: (2, 4) is clean, and the failures need `s >= 5`.

Grouped by what the result should have been versus what came back:

| target | expected -> produced | cases |
|---|---|---|
| (2, 5) | infinity -> NaN | 14 |
[...]

PR #783, opened 2026-08-08
Patch the vendored SymFPU's float-to-float conversion

TrevorHansen:
Fixes the soundness bug reported in #782, by carrying a patch against the vendored SymFPU.

### The bug

`convertFloatToFloat` chooses between a rounding path and an unrounded "fast path strict promotion" on whether the exponent width increases — and reads the wrong width for that question:

```cpp
bool exponentIncreased = unpackedFloat<t>::exponentWidth(sourceFormat) <= unpackedFloat<t>::exponentWidth(targetFormat);
```

`significandWidth(format)` is the format's own `sb`, but `exponentWidth(format)` is the internal *unpacked* width, and the two come apart. A target whose format exponent range is smaller can still have an equal or larger unpacked width: source (3, 4) and target (2, 5) both unpack to 4. The narrowing is then taken for a promotion and returned unrounded and unclamped, so overflowing values never saturate to infinity — they come back as NaNs, subnormals or zeros — and ties never round, because `rounder()` is skipped outright.

It is a soundness bug, not just a wrong constant: a satisfiable formula answers `unsat` as soon as the conversion is built as a circuit rather than constant-folded.

Standard formats never reach it. Narrowing between Float16, Float32, Float64 and Float128 always shrinks the unpacked width too, so those take the rounding path; the bug needs a non-standard format on one side.

### The fix, and the trap in it

The fast path is decided on the format's exponent width — the condition it actually needs — while `expExtension` keeps the unpacked comparison. Those two have to stay separate, and the first version of this patch got it wrong by conflating them.

`expExtension` is computed before the branch and used by the rounding path as well. The unpacked width can shrink where the format's exponent does not: (2, 6) to (2, 4) goes 4 to 3. Widening that guard therefore underflows an unsigned subtraction and breaks conversions that were previously correct — 68 format pairs in the range checked. That regression was found by sweeping bit-widths, not by reading the code, which is why the new query file covers a significand-only narrowing alongside the cases that were broken.

### Verification

Expected results are computed with exact rational arithmetic, with no floating point in the checker. Two independently written encoders — one enumerating every representable value, one computing the exponent and significand directly — were cross-checked against each other on 3,513 conversions before either was trusted, and agree everywhere.

- **Wide formats**: Float16, Float32, Float64 and seven non-standard formats; all 90 ordered pairs, all five rounding modes, boundary values (zeros, infinities, NaN, minimum and maximum subnormal, minimum normal, maximum finite, and each of their bit-pattern neighbours in both signs) plus random values — **33,660 conversions, 0 wrong**.
- **Small formats, exhaustive**: every ordered pair over exponent widths 2 to 5 and significand widths 4 to 7, all five rounding modes, every source value. Running at the time of writing; I will post the result here.
- Both reproducers from #782 are fixed, and the full test suite passes (106/106).

### Machinery

Applying this needs the patch step to serve more than ABC, so the per-submodule parts move into a function and the patches move into `patches/<library>/`. ABC's behaviour is unchanged — apply once, skip what is already applied, hard error on a patch that neither applies nor reverse-applies, and do nothing when the submodule is not checked out. Each of those four paths was exercised.

The patch is meant to be temporary: it should be dropped once the fix is upstream, which is what the hard error on a bumped submodule is there to force.

Worth flagging for review: configure-time patching mutates the submodule working tree, so a configured checkout has a permanently dirty submodule — that is already true of ABC today, and this adds a second one. Vendoring a fork instead would avoid it, at the cost of the loud failure when upstream moves.


TrevorHansen:
Verification is complete, and it turned up a weakness in the test that is now fixed.

### The exhaustive sweep finished clean

Every ordered pair of formats over exponent widths 2 to 5 and significand widths 4 to 7 — 225 pairs — under all five rounding modes, for every source value: **710,400 conversions, 0 wrong**. Together with the wide-format run reported above (33,660 conversions across Float16, Float32, Float64 and seven non-standard formats), that is 744,060 conversions checked against exact rational arithmetic with no divergence.

### The regression test did not actually test the regression

I had verified the first version of the test passed with the patch, but not that it failed without it. It did not, meaningfully: of its nine probes only one failed on an unpatched build. Two mistakes — the values I had labelled as overflowing were inside the target's range (3 converted into (2, 5), whose maximum finite is 3.875), and one literal was written wider than the format it was written for. A test that passes either way documents nothing.

The cases are now derived rather than guessed: the unpatched build was run against the oracle over every input of each affected pair, and one representative kept per rounding mode, per expected result class (infinity, largest finite, subnormal) and per sign. The result is twelve cases that each demonstrably fail without the patch and pass with it, spanning all five rounding modes and both signs.

Measured both ways on the same source tree:

| build | divergence cases | controls |
|---|---|---|
| without the patch | 12 sat — every one detects the bug | 5 unsat |
| with the patch | 12 unsat | 5 unsat |

The five controls were correct before the fix and must stay correct after. They are not padding: in (2, 6) to (2, 4) the unpacked exponent width shrinks while the format's does not, which is exactly the case that the over-broad first version of this patch broke by underflowing the extension width. That mistake is now pinned by a test rather than by a comment.

Full suite remains 106/106.

TrevorHansen:
Rebased onto #785, which splits out the AIG single-operand gate fix that the branch above this one needs. The base is now `aig-single-child-node`; the contents of this PR are unchanged.

PR #784, opened 2026-08-08
Accept every floating-point format SMT-LIB allows

TrevorHansen:
**Stacked on #783** — please merge that first; the base is `symfpu-conversion-patch`, not `master`. This PR needs the per-submodule patch machinery that #783 introduces, and reviewing it against master will show #783's changes too.

STP refused two things the SMT-LIB theory permits. Neither refusal was about the formats — both worked around bugs in the vendored SymFPU. Now that patches can be carried against that submodule, this fixes them there and takes the refusals out.

### What was refused, and why

**Significands of three bits or fewer**, at the parser and the C interface. `unpackedFloat::exponentWidth` returns the format's own exponent width for those — correct for holding subnormals — but `unpack` extends the packed exponent into that width and asserts `INVARIANT(unpackedExWidth > exWidth)` strictly. It aborted on its own invariant, or with assertions off (which `CMAKE_BUILD_TYPE=Release` forces) underflowed into a segfault. `0002` gives those formats the headroom bit: it costs one bit and satisfies both, since 2^e ≥ 2^(e−1) + (s−3) for every s ≤ 3.

**`fp.roundToIntegral` wherever the unpacked exponent width equalled the significand width** — an unbounded family, (eb, eb+1) for every eb ≥ 3, so (4,5) and (8,9) and upwards. It resizes a rounding point of width `exponentWidth + 1` with `matchWidth`, which may only widen, on a guard that admitted equality. `0003` makes the guard strict so the extract arm runs there, which the collar above it already made safe.

**The two are not independent.** Reaching `(_ FloatingPoint 2 3)` at all needs the first; doing so gives it an unpacked exponent width equal to its significand width, which is exactly the family the second fixes. Neither alone leaves that format working — verified by building each subset.

### A latent STP bug this uncovered

Lifting the refusal immediately hit an assertion failure in `BBNodeManagerAIG::CreateNode`, which is STP's own code, not SymFPU's. It handles a one-element child vector for `AND` and `OR` but not for `NAND`, `NOR` or `XOR`, which fall through to `makeTower` — and `makeTower` needs at least two children. A two-bit significand stores exactly one bit, and the classification predicates build `NOR` over that field, so `(_ FloatingPoint 2 2)` with `fp.isInfinite` was enough to abort it. Fixed in its own commit; the results are the ones the two existing cases already imply.

### Verification

Expected results come from exact rational arithmetic, with no floating point in the checker.

- `fp.add` over **every operand pair** at (2,2), (2,3), (3,2), (3,3), (5,2), (4,3), (6,2), under all five rounding modes.
- `fp.roundToIntegral` over **every value** at (2,3), (3,3), (3,4), (4,5), (5,6), (2,5), (6,7), under all five rounding modes — the first four of those could not previously be tested at all.
- Full suite passes.

Numbers to follow in a comment; the run is finishing as I open this.

Being able to test *exhaustively* is the practical argument for these formats. A two-bit-exponent, two-bit-significand float has sixteen values, so every operation over it can be checked completely — which is not affordable at any standard format.

### Test changes

The three query files that asserted the refusals now assert that the formats compute correctly, and the API test that expected a rejection now builds constants across the narrow formats instead. The wide-exponent one is deleted: it pinned the top of the roundToIntegral family and was affordable only because the refusal happened before any circuit was built.

The `matchWidth` backstop stays. It is no longer expected to fire, so its message no longer names a predicate that decided in advance which formats would reach it.


TrevorHansen:
Verification numbers, re-run against the final build (an earlier run straddled a rebuild and was discarded).

Expected results from exact rational arithmetic, no floating point in the checker.

**`fp.add`, every operand pair, all five rounding modes**, at formats that were refused outright:

| format | conversions | wrong |
|---|---:|---:|
| (2, 2) | 650 | 0 |
| (2, 3) | 2,750 | 0 |
| (3, 2) | 3,770 | 0 |
| (3, 3) | 15,390 | 0 |
| (5, 2) | 76,250 | 0 |
| (4, 3) | 71,390 | 0 |
| (6, 2) | 316,250 | 0 |

**`fp.roundToIntegral`, every value, all five rounding modes**:

| format | cases | wrong |
|---|---:|---:|
| (2, 3) | 130 | 0 |
| (3, 3) | 290 | 0 |
| (3, 4) | 570 | 0 |
| (4, 5) | 2,410 | 0 |
| (5, 6) | 9,930 | 0 |
| (2, 5) | 490 | 0 |
| (6, 7) | 40,330 | 0 |

**540,600 operations checked, none wrong.** Of those formats, (2,2), (2,3), (3,2), (3,3), (5,2), (4,3) and (6,2) could not be parsed before this change, and roundToIntegral at (2,3), (3,3), (3,4), (4,5), (5,6) and (6,7) could not be built.

Full test suite passes (106/106).

Each patch was also shown to be individually necessary, by building the five subsets and probing four behaviours. Dropping 0002 makes any short-significand format dump core; dropping 0003 makes roundToIntegral at (4,5) hit the matchWidth backstop; and (2,3) fails without either, which is the interaction described above.

TrevorHansen:
The AIG single-operand gate fix has moved out of this PR into #785, where it stands on its own with a unit test. This branch is rebased on top of it — the stack is now #785, then #783, then this one — and its own contents are unchanged apart from that commit leaving.

The two -Werror failures are also fixed: `bitsToRepresent` was orphaned when the refusal predicates went, and the rewritten API test asked for a `Type` it no longer needs. Both were invisible locally because the build was not configured with `-DWERROR=ON -DBUILD_EXTRA_TOOLS=ON`; it is now.

aytey:
Can we please upstream the patch? As in, make a PR against symfpu with those changes?

TrevorHansen:
> Can we please upstream the patch? As in, make a PR against symfpu with those changes?

Yes, will do a PR on symfpu when I'm clearer on what's happening.

PR #789, opened 2026-08-09
Reject command lines whose options cancel each other out

TrevorHansen:
STP took several option combinations it cannot honour both halves of and resolved them silently, in favour of whichever the code happened to apply last:

- `stp --cadical --minisat` ran CaDiCaL, because `parse_options()` consults the solver flags in a fixed sequence rather than in the order given.
- `--disable-simplifications --flattening true` ran with flattening off, because `disableSimplifications()` clears the flag after the command line has been read.
- `--parse-only --output-CNF` wrote no CNF, because parse-only stops before one exists.

In each case the second request never reached the solver and nothing said so. That matters most for the generated command lines that option sweeps and build scripts produce, where a dropped flag reads as a measurement.

CLI11 has had `excludes()` for this the whole time and STP used none of it across its 42 options. The relationships are now declared in one block at the end of `create_options()`, looked up by name so the option definitions above are untouched — and `get_option()` throws if a name there stops matching one, so renaming an option cannot quietly drop its relationships.

## What is now rejected

| Combination | Why |
|---|---|
| two solver flags | only one solver runs |
| `--threads` + a solver that is not CryptoMiniSat | nothing else reads a thread count |
| `--disable-simplifications` + a simplification it clears | 15 options, listed against `disableSimplifications()` |
| `--size-reducing-only` + one it forces | 3 options, against `disableSizeIncreasingSimplifications()` |
| `--bb.simplify-during-bb` + `--disable-opt-inc` / `--disable-simplifications` | needs the rewriting simplifier; its help text already said so |
| `--parse-only` + `--output-CNF` / `--exit-after-CNF` | no CNF is ever built |
| `--interactive` + `--CVC` / `--SMTLIB1` | `interactive_read` is read only on the SMT-LIB2 path |

CLI11 tests whether an option was given, not what it was set to, so `--disable-simplifications --flattening false` is refused as well even though the two agree. That is the same mistake in a milder form, and treating it the same way keeps the rule to one sentence.

## What is deliberately left alone

- **`--cadical-factor`** — a request CaDiCaL may decline with a warning rather than a setting that must apply, and `tests/query-files/CMakeLists.txt` sweeps the whole corpus with it appended to every invocation. An exclusion would fail every test in that run that named a solver. Verified: with another solver selected it still warns and solves.
- **`--search-bias`** — documented as ignored by solvers that have no such setting.
- **The parser flags** — `parse_options()` already rejects those with a message of their own, which the existing test covers.

## Fixed along the way

`--riss` was declared with no variable to bind to, and `parse_options()` never looked at it. Under a build with another solver compiled in, asking for Riss selected nothing and said nothing. It is now bound and applied like the others. Untested here — there is no Riss checkout to build against — but it follows the shape of the four flags beside it.

## Testing

`tests/query-files/misc-tests/bad-cli-options.smt2` gains the new rejections plus a set of combinations that must still be accepted, so an over-broad exclusion fails the suite. The solver-flag cases are in a new `bad-cli-options-solvers.smt2` carrying `REQUIRES: minisat`, since which solvers exist varies by build and `REQUIRES` is per file.

Built and exercised in three configurations, all with `-DWERROR=ON`: minisat-only, `+CaDiCaL`, and `+CryptoMiniSat` — the last two being where the `#ifdef`-ed `--threads` and solver-list code differs. Full `query-files` suite: 427 passed, 18 unsupported, 0 failed. Also checked by hand that appending `--cadical-factor=on` to the new tests' command lines, as the corpus sweep does, leaves both their pass and fail outcomes unchanged.

### DIFF
--- a/tests/query-files/misc-tests/bad-cli-options-solvers.smt2
+++ b/tests/query-files/misc-tests/bad-cli-options-solvers.smt2
@@ -1,0 +1,28 @@
+; Two solver flags on one command line must be diagnosed, not resolved.
+;
+; parse_options() consults the solver flags in a fixed order rather than in
+; the order they were given, so 'stp --cadical --minisat' used to run CaDiCaL
+; without saying so. Only one can be meant.
+;
+; Separate from bad-cli-options.smt2 because this needs a pair of solver flags
+; that exist, and which solvers are compiled in varies. --minisat and
+; --simplifying-minisat come and go together, and neither is a default, so
+; they are the pair to test with.
+;
+; REQUIRES: minisat
+
+; RUN: not %solver --minisat --simplifying-minisat %s 2>&1 | %OutputCheck %s --check-prefix=TWOSOLVERS
+; RUN: not %solver --simplifying-minisat --minisat %s 2>&1 | %OutputCheck %s --check-prefix=TWOSOLVERS
+; TWOSOLVERS-NOT: terminate called
+; TWOSOLVERS: excludes
+
+; Either one on its own is how the flag is meant to be used.
+; RUN: %solver --minisat %s 2>&1 | %OutputCheck %s --check-prefix=SOLVE
+; RUN: %solver --simplifying-minisat %s 2>&1 | %OutputCheck %s --check-prefix=SOLVE
+; SOLVE: ^sat$
+
+(set-logic QF_BV)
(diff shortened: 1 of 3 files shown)


TrevorHansen:
Let's try this out for a while and see if it just annoys us or stops problems.

PR #792, opened 2026-08-09
Remove the optional bit-blast simplification

TrevorHansen:
> Stacked on #789, which must merge first. That PR registers `--bit-blast-simplification` in an exclusion list looked up by name, so if these two landed independently `get_option()` would throw at startup on the missing option. Base changes to `master` once #789 is in.

The pass bit-blasted the whole formula into a throwaway AIG part-way through preprocessing, read back the nodes ABC had proved constant or structurally equal, substituted those at the word level, and discarded the AIG. It was gated on `--bit-blast-simplification`, a difficulty threshold whose default of `0` makes the guard `initial_difficulty_score < 0` false — so only an explicit `-1` (or a positive bound) ever ran it, and nothing outside `main.cpp` writes the field. It was unreachable in every default run.

What the flag did reach was a crash. `BBTerm` has no arm for `READ`, and arrays are only rewritten away above by `numberOfReadsLessThan(10)` — 50 with `--ackermanize` — so any query with ten or more surviving reads walked the scratch bit-blast into a `READ` and hit `BBTerm: Illegal kind to BBTerm`. The pass already carried a guard of exactly this shape for array equality, whose substitutions bypass the solver map and could strip a symbol the extensionality procedure depends on; this was a second array assumption made without checking.

Rather than add the missing guard, drop the pass. It has been off by default for its whole life, so there is no measurement saying it repays the full bit-blast it costs, and the one configuration that switched it on was broken.

## What goes with it

**The AIG-size second opinion on difficulty reversion.** A later block re-blasted the final formula and marked the simplification "worse" if the AIG had grown, comparing against the node count this pass left behind — its comment notes the difficulty score "is sometimes completely wrong, the sage-app7 are the motivating examples". That baseline had no other source, so the check was equally unreachable by default and goes too. It is a discarded idea rather than a lost feature: reinstating it needs its own flag and its own baseline.

**Not removed:** `BitBlaster::getConsts`, which `tools/rewrite_rule_gen` still calls. That leaves the AIG-equivalence harvesting in the core library serving one extra tool; retiring that tool would let it go as well.

The difficulty snapshot retaken after floating-point lowering stays, but its comment now gives the reason that outlives the threshold: everything downstream works on the lowered form, so the value reversion later compares against must describe that form, and the recompute below it is skipped for array problems.

## Testing

The crash reproducer now solves. Full `ctest`: 106/106. Builds clean under `-DWERROR=ON` with CaDiCaL and with CryptoMiniSat.

Behaviour neutrality is argued above from the default flag value, and confirmed by measurement. Two statically linked binaries were built from this commit and its parent — static because a build-tree RUNPATH would make a copied binary load the rebuilt `libstp` and the comparison pass vacuously — and run over 1197 files (the whole `query-files` corpus plus 900 from a fuzz corpus), byte-comparing every generated CNF as well as the answers:

| | |
|---|---|
| identical CNF and answer | 1196 |
| CNF differs | 0 |
| answer differs | 0 |

The 1197th, a 2500-query file, first showed an answer difference that turned out to be the harness: it needs longer than the per-file cap, so the two runs truncated at different points. Uncapped, both binaries produce byte-identical 2500-line output. A sweep of the remaining ~1600 corpus files is running and I will report it here.

### DIFF
--- a/include/stp/STPManager/UserDefinedFlags.h
+++ b/include/stp/STPManager/UserDefinedFlags.h
@@ -72,7 +72,6 @@ public:
   bool enable_common_subsum = false;
 
   int64_t AIG_rewrites_iterations = 0; // Number of iterations of AIG rewrites.
-  int64_t bitblast_simplification = 0;
   int64_t size_reducing_fixed_point = 0;
   
 
@@ -246,7 +245,6 @@ public:
     enable_common_subsum = false;
     enable_ite_context = false;
 
-    bitblast_simplification = 0;
     simple_cnf=true;
   }
 
--- a/lib/STPManager/STP.cpp
+++ b/lib/STPManager/STP.cpp
@@ -51,7 +51,6 @@ namespace stp
 {
 
 const static string cb_message = "After Constant Bit Propagation. ";
-const static string bb_message = "After Bitblast simplification. ";
 const static string uc_message = "After Removing Unconstrained. ";
 const static string int_message = "After Unsigned Interval Analysis. ";
 const static string pl_message = "After Pure Literals. ";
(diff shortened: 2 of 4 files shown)


TrevorHansen:
The remaining sweep finished. Totals across the whole comparison:

| | |
|---|---|
| identical CNF and answer | **2797** |
| CNF differs | 0 |
| answer differs | 0 |

That is 297 files from `query-files` (the other 80 produce no answer to compare — print-back, parse-only and error-path tests) and 2500 from the fuzz corpus, every generated CNF byte-identical between the two static binaries. Together with the flag's default making the guard unreachable, the removal is neutral for any run that did not pass `--bit-blast-simplification`.

PR #797, opened 2026-08-10
Floating-point rewrite-rule generator, and the constant-operand rules it found

TrevorHansen:
The bitvector side has `tools/rewrite_rule_gen`; the floating-point kinds had nothing. This PR adds a floating-point rewrite-rule generator, and the rules it found missing from the `SimplifyingNodeFactory`.

## The generator (`tools/fp_rewrite_gen`)

A value-vector search rather than a syntactic enumerator: every depth-1 floating-point term and predicate over one float variable, the five rounding modes and a pool of special constants (NaN, ±oo, ±0, ±1) is blasted once through the hashing factory, and the resulting pure bitvector/Boolean circuit is folded to a constant under every assignment of a small format — the oracle `test_fprewrites` already uses. A candidate whose value vector equals that of a trivially cheaper form (a constant, `x`, `(fp.neg x)`, `true`/`false`, `(fp.isNaN x)`, …) is a sound rewrite at that format; hits are re-confirmed on a second format, and rebuilt through the `SimplifyingNodeFactory` to report whether it already knows the rule. `fp.min`/`fp.max` candidates must reach the same target under both pinned (+0, −0) choices. The whole space — 1730 candidate evaluations — runs in about 23 seconds, and found 140 sound rules of which 119 were missing.

## The rules

The 119 compress into six families, each a kind-plus-constant-operand check at construction, and each firing deletes an entire blasted symfpu circuit:

- **NaN absorbs through the arithmetic**: `fp.add`, `fp.mul`, `fp.div` and `fp.fma` with a NaN operand are that NaN. `fp.sub` and the mirrored comparisons inherit the rules through the existing normalisations.
- **Signed-zero addition identities**, guarded by a literal rounding mode: `x + (-0) = x` under every mode except round-toward-negative, and `x + (+0) = x` only there. The sum of the two opposite zeros is the one input that can see the zero operand — an unconditional version of either rule would be unsound, which the generator's per-mode grouping catches automatically.
- **`fp.fma` with an exact product is the `fp.add` of that product**: a ±1.0 multiplicand gives round(±y + z), and two constant multiplicands holding a zero give the xor-signed zero — or NaN for the invalid 0 · oo. The created add re-simplifies, so its own NaN and signed-zero rules finish the job.
- **`fp.rem`'s invalid operands**: a NaN or infinite dividend, or a NaN or zero divisor, make the remainder NaN whatever the other operand is. `fp.div` by −1.0 folds to `fp.neg`, alongside the existing 1.0 rule.
- **`fp.min`/`fp.max` ignore a NaN operand and absorb the matching extreme**: min(x, −oo) = −oo and max(x, +oo) = +oo for every x, NaN included. Sound for either choice of the totalising pass, which only matters on the two opposite zeros.
- **Comparisons against the extremes**: nothing is greater than +oo or less than −oo, nothing compares against NaN, and +oo ≥ x and x ≥ −oo hold exactly when x is not NaN.

One sound finding is deliberately **not** applied: `(= x NaN)` is exactly `(fp.isNaN x)`, but `PropagateEqualities` substitutes through a float `=` with a constant side (x := NaN everywhere), which is strictly stronger than holding the smaller predicate. A comment in the factory and a test pin the decision.

## Verification

- `test_fprewrites` grows from 60 to 135 checks: every new rule must produce the exact expected node, and the unsimplified original must agree with it on every float of the format, through the same lowering a real solve uses. The rm-guarded identities are checked under all five modes, including the modes where the rule must not fire.
- Re-running the generator against the patched factory reports every rule handled (the two `=`-against-NaN spellings excepted by design).
- Full test suite passes; on a 300-file sample of SMT-LIB QF_FP benchmarks the solver's answers agree with the benchmarks' `:status` annotations on all 294 files that carry one (10 s timeout, none hit).

### DIFF
--- a/include/stp/NodeFactory/SimplifyingNodeFactory.h
+++ b/include/stp/NodeFactory/SimplifyingNodeFactory.h
@@ -108,6 +108,11 @@ private:
   // sign bit, keeping the rest of the packed bits and the format.
   ASTNode foldFPSign(const ASTNode& fpConst, bool flip);
 
+  // The special constants of a format, for rules whose result is not one of
+  // the operands: interning canonicalises the NaN.
+  ASTNode makeFPNaN(unsigned eb, unsigned sb);
+  ASTNode makeFPZero(unsigned eb, unsigned sb, bool negative);
+
   ASTNode plusRules(const ASTVec& oldChildren);
 
 };
--- a/lib/NodeFactory/SimplifyingNodeFactory.cpp
+++ b/lib/NodeFactory/SimplifyingNodeFactory.cpp
@@ -26,6 +26,7 @@ THE SOFTWARE.
 #include "stp/AST/AST.h"
 #include "stp/AST/ASTKind.h"
 #include "stp/AbsRefineCounterExample/ArrayTransformer.h"
+#include "stp/FloatBlaster/rounding_modes.h"
 #include "stp/Simplifier/Simplifier.h"
 #include <cassert>
 #include <cmath>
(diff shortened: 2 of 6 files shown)


TrevorHansen:
Two additions since the PR opened: the generator now searches **depth 2**, and the high-value rules it found are implemented.

**The search** nests one inner term — from a pool chosen to constrain value class and sign: `(fp.abs x)`, `(fp.neg x)`, `(fp.sqrt rm x)`, `(fp.roundToIntegral rm x)`, `x*x`, `x+x`, `x-x` — into one operand slot of every outer operation, adds the classification predicates as outer operations, and adds "the inner term itself" to the target library. 8754 evaluations in ~3 minutes; 626 sound rules, 290 missing from the factory, all confirmed on the second format.

**Implemented** (the new `SimplifyingNodeFactory` commit):

- Never-below-zero terms (`fp.abs`, `fp.sqrt`, a self-product `t*t`) against sign-decided constants: never under a nonpositive constant, over a strictly negative one exactly when ordered, and a zero on the other side squeezes to `fp.isZero`. Plus `t > |t| → false` and `|t| ≥ t → ¬fp.isNaN t` with no constant involved.
- Classification predicates look through value-preserving shapes (extending the abs/neg peel): `isNaN` through `roundToIntegral`/`t+t`/`t*t`, `isZero` through `sqrt`/`t+t`, `isInfinite` through `roundToIntegral` (whose result is also never subnormal), and the sign predicates through abs/neg/`roundToIntegral`/`t+t`/`t*t`. These compose with the comparison rules, which emit `fp.isNaN`/`fp.isZero` terms for them to reduce.
- `(= t NaN) → (fp.isNaN t)` when `t` is **not** a bare symbol — the bare-symbol form still keeps its `=` for the equality-substitution pass.
- `fp.roundToIntegral` idempotence across *different* rounding modes (an already-integral value rounds exactly under every mode), and `abs(t*t) → t*t`.

**Deliberately skipped** (documented in the code): the `x−x` consequence cluster (syntactically rare), `fp.min`/`fp.max` against constants, `fp.eq`/`=` of compound terms against non-NaN constants, and the mode-guarded underflow/overflow squeezes the search surfaced (e.g. `fp.eq(x*x, ±0) → isZero(x)` is sound **only** under RTP — a tiny square underflows to +0 in every other mode).

Verification: `test_fprewrites` grows to 202 checks (including the cross-mode `roundToIntegral` case the generator could not produce); re-running the generator reports every implemented family handled; the full test suite passes (one printer-test fixture updated — its `fp.isNaN(x + x)` now simplifies at construction, so it uses two distinct floats); and a 300-file SMT-LIB QF_FP sample again answers 294/294 in agreement with the benchmarks' `:status` annotations.

Issue #483, opened 2024-04-26
Incremental STP Solving


sreeshmaheshwar:
Does STP support incremental SMT solving techniques (e.g. reusing learned lemmas) as other solvers, like Z3 and Boolector, do? (Can't find anything in the docs about how to enable an incremental solver and haven't observed any performance differences when using `push` and `pop` in an incremental way with STP)

If not, are there any plans for this?

TrevorHansen:
You're right that STP doesn't keep learned clauses between invocations to the SAT solver.

It has some simple analysis, for example if a problem is unsatisfiable, and you push more frames onto it, STP knows that those problems are unsatisfiable, too.

There's lots we could do to make incremental solving faster. If you provide us with some problems that are really slow with STP ,compared to other solvers, I'll see if there's anything we can do to speed up STP's solving of them.

sreeshmaheshwar:
Many apologies for the late reply. Thanks @TrevorHansen for your response and willingness.

Attached are some example files, using incremental querying with `push`/`pop`, for which Z3 outperforms STP, for me at least. They are KLEE generated, so sorry for the unreadability (I did try to condense them and remove redundancy but some remains) - hopefully it is not so bad that it isn't useful.

The time differences for me on my Ubuntu 20.04 machine are far more dramatic (1.5 minutes for STP vs 1.5 seconds for Z3) though I suppose the configuration is outdated. On my Ubuntu 22.04 machine, the differences are 15.1s (STP) vs 0.7s (Z3) and 8.7s (STP) vs 0.3s (Z3). I use mainline STP throughout, with Z3 4.8.12 (22.04) and 4.8.7 (20.04).

I can produce more files that show STP being outperformed in such an incremental setting, so do let me know if that would be helpful, though my current process of printing into SMT-LIB is a bit painful. If these differences are not substantial enough and you'd like for a small number of queries that produce larger differences, I can try for that as well but I've struggled to find this so far (through KLEE).

Please let me know if there's anything else that I can help out with on this topic! I'd be very happy to see STP supporting incremental solving.

SMT-LIB2 files: [dircolors.txt](https://github.com/stp/stp/files/15456941/dircolors.txt), [b64.txt](https://github.com/stp/stp/files/15456459/b64.txt)

sreeshmaheshwar:
@TrevorHansen just wanted to friendly ping on this as I have some free time now - are these examples sufficient?

sreeshmaheshwar:
[...]

PR #836, opened 2026-08-13
Incremental solving (5/5): the persistent driver

aytey:
# Incremental solving 5/5 — the persistent driver

**This PR builds on top of #835** (4/5 — the scope ledger and the CBP engine). Please review and merge it first; this branch contains its commits.

The last part of the series, and the one that answers #483. One SAT solver, one AIG and one CNF encoding now live across `(check-sat)` calls.

`docs/incremental-solving.rst` arrives here and is the thing to read first — it is the design document, not a summary of the diff, and it says what each measured policy is fitted to.

## How retraction works

Everything the driver encodes is a **conservative extension**: fresh Tseitin variables and definitional clauses. It therefore stays valid for the whole encoding epoch, and what changes between checks is only which root literals are asserted.

A base-level (level 0) conjunct becomes a permanent unit clause. A conjunct at a pushed level has its root literal *assumed* per solve, so a pop retracts it by simply not assuming it any more. **Learned clauses survive both check-sats and pops by construction** — nothing is deleted, so nothing has to be justified as safe to delete. That is the reuse #483 asked about.

Base-level units being permanent is sound because `reset` and `reset-assertions` destroy the driver object; a fresh one is built on demand.

## Push and pop need no hooks

The parser's assertion stack stays canonical and is supplied as a complete snapshot on every call. The driver reconciles it into the scope ledger from 4/5, which owns stability, promotion, active preprocessing and model state. The assumption set is recomputed from the snapshot against permanent content-addressed encodings, so no SAT clause deletion and no frontend push/pop callback is required, and a re-collapsed level conjunction cannot go stale.

## Coverage

The whole input language: plain bit-vectors, arrays by lazy refinement or `--ackermanize`, floating point, and `--array-equality`. `canHandle()` is the seam should a future construct need excluding; today it answers yes for everything both frontends can produce, does no work and caches nothing.

## When a session engages

A session becomes incremental at its first `(push)`, or from the start with `--incremental` (`vc_setFlags 'i'` for the C API). Engagement is then *delayed* rather than immediate, by the policy from 4/5: the first solves carry the largest all-new formulas, where the batch pipeline's whole-formula simplification earns its keep, and a session too short to repay a persistent encoding never builds one. `--incremental-auto-engage-at` overrides the ordinal, and `INCREMENTAL_AUTO_ENGAGE_AT` reaches the same policy from the C API.

## Memory is bounded, not monotone

The encoding grows monotonically **within an epoch**. When the solver's variable count, or the graph the semantic caches pin, has grown mostly out of *popped* content, the complete semantic/AIG/SAT epoch is rotated and reconstructed from the live stack. Rebuilds that exist only to replace the SAT backend keep the semantic epoch — re-simplifying the whole base for a rebuild whose entire purpose is to set one CaDiCaL option was measured at 18ms per rebuild on a three-thousand-conjunct base, scaling with the base.

The liveness test before a rotation is an exact walk of the live AIG cone, not an estimate. Three of the tests here are negative cases for it: a monotonically growing live extensionality stack, an ordinary root built mostly from AIG cones first introduced by popped formulas, and a shared cone. Neither live shape may be mistaken for dead churn.

## Everything beyond the core, and the switch that turns it off

`--incremental-core-only` runs the persistent assumption/refinement core alone. It disables cross-level constant-bit propagation, promotion of long-stable levels to permanent units, per-level substitution and simplification, the first-solve shortcuts, the retirement of inprocessing that costs more than it earns on a many-solve session, trail reuse across prefix-stable assumption sequences, and phase hints away from retracted content.

**It deliberately keeps every theory-correctness mechanism and the memory-relief rotation.** That is what makes it useful during review: if a query disagrees with the batch engine, `--incremental-core-only` says whether the fitted work is implicated or the core is.

## The worker thread

`checkSat` runs its body on a worker thread with a large explicit stack. Several passes walk formulas by recursion, and parse-time inlining of chained `define-fun`s builds nodes tens of thousands of levels deep from flat input, which exhausts a default-sized stack. The node uid counter travels with the worker, so nodes it creates continue the spawning thread's numbering — a worker starting from its own zero would give fresh nodes uids that collide with live main-thread nodes, and every cache keyed on node numbers could then cross-bind.

## Note for the reviewer

Three things that are easier to check than to spot:

- **`RemoveUnconstrained` gains an `alsoUntouchable` set.** A symbol constrained at another assertion level, or by an already-encoded conjunct, is mentioned once in the formula being simplified and is not unconstrained. The parameter defaults to `NULL` and the batch path passes nothing.
- **`SimplifyingNodeFactory`'s reflexive-`fp.eq` fold is not gated on the solver mode any more,** and `test_fprewrites` asserts that it is not. Node construction is context-free; an earlier version of this branch suppressed the fold while incremental solving was on, which meant a pushing session got a different word-level DAG even when the driver never engaged, and the batch-versus-incremental differential was comparing two engines handed different graphs.
- **Models are constructed lazily.** A `sat` answer defers construction unless something reads the model at solve time; the readers call `materializePendingModel()` first. `Cpp_interface` asks `hasIncrementalSolver()` rather than `getIncrementalSolver()` before doing so, because the latter constructs a driver on demand — asking it whether a driver exists would build one, and a SAT backend with it, in every batch session that printed a model.

## Walks keep their state on the heap

`splitConjuncts`, which flattens a level's conjunction into the conjunct list the level is encoded and assumed in, walks nested ANDs iteratively. A chain of `define-fun`s inlined at parse time nests them as deep as the input says, so the depth is the writer's choice and not something a fixed stack can accommodate. Children are pushed in reverse, so they come back off in the order the recursion visited them -- a level's content must not depend on how the walk is organised -- and, as before, shared subtrees are not deduplicated.

It is the one converted walk in the series with no `DeepDag_Test` case: it lives in an anonymous namespace and nothing outside the file can name it. Its coverage would come from driver-level cases in that harness.

`ArrayTransformer::TransformDriver` reaches the touched-read recording through aliases beside `arrayToIndexToRead` and `ack_pair`, matching how the driver already reaches the rest of the transformer's state.

## Note on 32-bit

Two configured limits are narrowed to `size_t` through one `clampToSize` helper. The clamp is vacuous where `size_t` is 64 bits and load-bearing where it is 32, so it goes through `std::min` rather than an explicit comparison -- written as a comparison, clang proves it always false on a 64-bit target and `-Werror` rejects it, while removing it would break the i386 leg.

## Verification

Every axis with `-DWERROR:BOOL=ON`:

| axis | result |
| --- | --- |
| FP + assertions + tests | **125/125 ctest**, 551 lit (533 passed, 18 unsupported, 0 failures), 103 deep-DAG cases |
| no-fp + tests | **96/96 ctest**, 90 deep-DAG cases |
| g++-13 Release | clean |
| clang-21 Release (`--target stp`) | clean, zero warnings from STP's own sources |

The 18 unsupported lit tests are the `REQUIRES: libbf` ones, which this configuration does not enable.

## End of series

With this merged, all nine commits in the chain — the four prerequisites plus the five parts above — reproduce the `incremental-solving` development branch exactly: same tree hash, and `git diff master` over the merged chain matches `git diff master incremental-solving` byte for byte.


aytey:
## Benchmark scatter plots: virtual timeout 24s

Comparing `2c71c1fd` (master) vs `23df5e57` (incremental).

### All theories
scatter_all_24s

### QF_ABV
scatter_QF_ABV_24s

### QF_ABVFP
scatter_QF_ABVFP_24s

### QF_BV
scatter_QF_BV_24s

### QF_BVFP
scatter_QF_BVFP_24s

### QF_FP
scatter_QF_FP_24s

aytey:
## Benchmark scatter plots: virtual timeout 2m

Comparing `2c71c1fd` (master) vs `23df5e57` (incremental).

### All theories
scatter_all_2m

### QF_ABV
scatter_QF_ABV_2m

### QF_ABVFP
scatter_QF_ABVFP_2m

### QF_BV
scatter_QF_BV_2m

### QF_BVFP
scatter_QF_BVFP_2m

### QF_FP
scatter_QF_FP_2m

Issue #444, opened 2022-08-11
Performance issue on QF_BV query (missing rewrite?)


aytey:
For this file:

```smt2
(set-logic QF_BV)
(declare-const __T1 (_ BitVec 5))
(declare-const __ (_ BitVec 6))
(declare-const _T1 (_ BitVec 3))
(declare-const __T (_ BitVec 6))
(declare-const T1 (_ BitVec 3))
(declare-const T (_ BitVec 7))
(declare-const _T (_ BitVec 1))
(assert
  (let
    ((?x25 ((_ zero_extend 24) ((_ zero_extend 1) ((_ zero_extend 1) __T)))))
    (bvslt
      (_ bv111 32)
      (bvadd
        ((_ zero_extend 24) ((_ zero_extend 4) ((_ zero_extend 1) T1)))
        (bvadd
          ((_ zero_extend 24) ((_ zero_extend 1) T))
          ?x25
          (
            (_ zero_extend 24)
            ((_ zero_extend 1) ((_ zero_extend 1) ((_ zero_extend 1) __T1))))
          ((_ zero_extend 24) ((_ zero_extend 4) ((_ zero_extend 1) _T1)))
          ((_ zero_extend 24) ((_ zero_extend 1) ((_ zero_extend 1) __))))))))
(assert
  (let
    (
      (?x1088
        (bvadd
          ((_ zero_extend 24) ((_ zero_extend 7) _T))
          ((_ zero_extend 24) ((_ zero_extend 1) ((_ zero_extend 1) __))))))
    (bvsge
      (_ bv111 32)
      (bvadd
        (bvadd
          ((_ zero_extend 24) ((_ zero_extend 4) ((_ zero_extend 1) T1)))
          (
            (_ zero_extend 24)
            ((_ zero_extend 1) ((_ zero_extend 1) ((_ zero_extend 1) __T1)))))
        ?x1088
        ((_ zero_extend 24) ((_ zero_extend 4) ((_ zero_extend 1) _T1)))
        ((_ zero_extend 24) ((_ zero_extend 1) ((_ zero_extend 1) __T)))
        (bvadd
          ((_ zero_extend 24) ((_ zero_extend 7) _T))
          ((_ zero_extend 24) ((_ zero_extend 1) T)))))))
(check-sat)
```

STP as submitted to SMTCOMP 2022 times-out:

```
/usr/bin/time -vvv ./stp reduced.smt2
Timed Out.
        Command being timed: "./stp reduced.smt2"
[...]

PR #858, opened 2026-08-13
Stop two adder-sharing opportunities being thrown away before CommonSubSum sees them

aytey:
Two additions that share operands should build one adder between them. #777 added `CommonSubSum` to arrange exactly that, and it works — but two rules sitting on either side of it hand it inputs it cannot use, and both failures are silent: the pass reports `Adders saved:0` and nothing indicates anything was missed.

Found while checking whether #444 is still open. It isn't — the instance in it answers `unsat` in 0.02s on master, because #777 also turned `--flattening` on by default, which is what @TrevorHansen said would fix it when it was filed. These are the two things that turned up on the way to establishing that.

## The width-narrowing rule reads the operand count

`StrengthReduction` narrows an addition whose operands all have leading zeros: if the sum cannot reach the node's width, rebuild it at a width that can carry it, extracting each operand down. The width it picked was the widest operand plus `ceil(log2(degree))` bits of carry.

Sound, but it reads the operand *count*. Two additions over almost the same operands land on different widths as soon as their counts straddle a power of two — and then their operands are extracts of different widths, so they are different nodes, and the two additions share nothing at all. Not for `CommonSubSum`, which is hunting for precisely that sharing, and not for the bit-blaster's own structural hashing either.

One 1-bit addend is enough. Two additions of n 24-bit addends inside 64 bits, the second carrying one extra 1-bit addend:

| n | 15 | **16** | 17 | 31 | **32** | 33 | 63 | **64** | **65** |
| --- | --- | --- | --- | --- | --- | --- | --- | --- | --- |
| `Adders saved`, master | 14 | **0** | 16 | 30 | **0** | 32 | 62 | **0** | **0** |
| `Adders saved`, this PR | 14 | **15** | 16 | 30 | **31** | 32 | 62 | **63** | **63** |

n=64 goes 0.11s → 0.05s and n=65 0.22s → 0.05s. The cliff is invisible from the outside; the query just costs twice the adders it needs to.

Bounding the sum by the sum of the operands' bounds removes it. An operand with `e` possibly-one bits is at most `2^e - 1`, so the sum is at most `sum(2^e_i - 1)`, and the width that carries it is that bound's bit length. Two things follow:

- **It is never wider than the old choice.** `sum(2^e_i - 1) < 2^(maxEffective + ceil(log2(m)))` for every input, so this only ever narrows further — the existing narrowing tests are unchanged.
- **It is stable.** A bound dominated by the large operands does not move far enough for a couple of tiny addends to push it past the next power of two, which is the property #444's shape needs.

The bound is accumulated in limbs rather than a machine word, because the node's width is whatever the input asked for. The headroom is `width+65` bits against a bound below `2^(width+32)`.

## CommonSubSum stalls one step short of a sub-multiset

The greedy pulls out the operand pair held by the most additions. Additions of fewer than three operands were skipped entirely, on the grounds that such an addition already *is* its own pair and extracting from it would leave a one-child `BVPLUS`. True of rewriting it — but it also meant a two-operand addition never counted as holding its operands, so a pair it already builds was invisible to every wider addition holding the same two.

The case that costs is an addition whose operands are a sub-multiset of another's, which is what #444 reduces to: a sum compared against itself plus two more addends. The greedy walks the smaller addition down to two operands and stops, because from there it votes for nothing — leaving the last adder the two have in common built twice. On the instance from that issue the pass reports four adders saved where five are available.

So count a two-operand addition as holding its operands, and tally its pair as one already realized. Extracting such a pair is not merely worthwhile, it is **free**: the node the extraction builds hash-conses to the addition that was already there, so no adder is spent setting it up, and one wider addition holding the pair is enough to be worth rewriting rather than the two the general rule needs. An addition the greedy has just narrowed to two operands becomes realized the same way, which is what lets the walk take its last step instead of stalling.

## Checking it

- **129/129 ctest**, including two new unit tests. Both fail without their change: `CommonSubSum_Test.sub_multiset_sum_becomes_the_shared_node` gets a 3-operand sum back instead of the shared node, and the new `StrengthReduction_Test` case sees the two additions at widths 7 and 6 instead of both at 6.
- **1500-query differential fuzz** built around the narrowing rule — random sums of zero-extended narrow variables and constants against random bounds — against a binary built from this branch's parent: **0 disagreements**, 938 unsat / 562 sat.
- **300 real QF_BV files** ([Zenodo 15493096](https://zenodo.org/records/15493096)) against the same baseline: **0 answer mismatches**, 186.5s vs 187.6s total. Neutral, which is what I expect — SVCOMP-style queries have few wide n-ary sums, and neither change can fire without them. I have not run a large-corpus performance campaign; if that matters before merging, say so and I will.

The query file added under `simplification-tests` is the instance from #444 itself. It answers `unsat` either way today, so it pins the issue rather than this PR — the unit tests are what fail without these changes.

## What I looked at and left alone

`Flatten` is deliberately sharing-aware: it splices a child into its parent only when the child has one reference. That means two AC-equal additions whose intermediate sums are referenced elsewhere keep their own nesting and stay distinct nodes. It reads like a third gap, but I could not build an instance where it actually costs anything, so I am not claiming it as one.

### DIFF
--- a/include/stp/Simplifier/CommonSubSum.h
+++ b/include/stp/Simplifier/CommonSubSum.h
@@ -124,8 +124,15 @@ class CommonSubSum
   // additions share nothing is all of them.
   ankerl::unordered_dense::set<uint64_t> shareable;
 
+  // Pairs an addition of exactly two operands already builds. Extracting
+  // such a pair out of a wider addition costs nothing -- the node it makes
+  // hash-conses to the two-operand addition, whose adder is being built
+  // anyway -- so one wider addition holding it is enough to be worth doing.
+  ankerl::unordered_dense::map<NodePair, uint32_t, PairHash> realized;
+
   void collect(const ASTNode& n, ASTNodeSet& seen, ASTVec& plusNodes);
   void markShareable();
+  void markRealized(const ASTVec& v);
   void eligibleOf(const ASTVec& v, std::vector<uint64_t>& out) const;
   bool bump(uint64_t a, uint64_t b);
   void drop(uint64_t a, uint64_t b);
--- a/lib/Simplifier/CommonSubSum.cpp
+++ b/lib/Simplifier/CommonSubSum.cpp
@@ -78,6 +78,13 @@ void CommonSubSum::markShareable()
     // A two-operand addition already *is* its own pair; extracting from it
     // would leave a one-child BVPLUS that rebuilds to itself. It is not a
     // holder, because pairs are never enumerated over it.
+    //
+    // Counting it as one would widen this set to operands that can only ever
+    // pair *within* a single wide addition, and a pair internal to one
+    // addition is never shared. The wide addition would then pay a C(k,2)
+    // enumeration for nothing: on one addition over 1500 variables alongside
+    // two-operand additions of the same variables, that is 5.5s in a pass
+    // that otherwise does no work at all.
     if (v.size() < 3)
       continue;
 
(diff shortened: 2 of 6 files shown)


TrevorHansen:
Great fix. I've got some other commits to CommonSubSum, and want this in beforehand.

PR #868, opened 2026-08-14
Run only the array rules in the pre-lowering unconstrained pass

TrevorHansen:
Follow-up to #867, which flagged this regression as known and unexplained: `dwp_formulas/try5_small_noof_functions_flanagansaxe_{shuf,shred}` about 15x slower (0.15 s → 2.85 s), and the `try3` dwp pair about 3x.

**It is nothing to do with arrays.** The pass added before `lowerArrayEqualities` runs all of `RemoveUnconstrained`, not just the array rules that needed to be there — and the ordinary bit-vector rules had never run at that point before. Disabling each array rule in turn does not fix it; disabling *all* of them does not fix it; removing the call site does:

| file | master | as merged | array rules off, window on | array rules on, window off |
|:---|---:|---:|---:|---:|
| flanagansaxe_shuf | 0.15 | 2.83 | 2.80 | 0.15 |
| flanagansaxe_shred | 0.15 | 2.81 | 2.98 | 0.15 |
| unconstrained10 | 21.78 | 0.04 | 21.88 | 22.07 |
| unconstrained07 | 7.04 | 0.05 | 6.97 | 7.01 |

The cost follows the call site, not the rules — and the window is still needed, since without it neither `unconstrained` file is fixed.

**Nor is it the encoding.** Under `--print-functionstat` every arm hands the bit-blaster the same node count (2969) and the identical difficulty score (69750 / 182566). Only the pre-lowering pass's own output differs (5238 nodes with the scalar rules, 5778 without), and that difference is gone by the end. What changes is which fresh variables stand in for what, so a different counterexample comes back, and the read-axiom refinement loop then needs 22 rounds where it had needed 2:

| arm | SAT solving calls | counterexample generations |
|:---|---:|---:|
| master | 2 | 2 |
| as merged | 22 | 22 |
| this change | 2 | 2 |

A model-guidance effect rather than a structural one, which fits its hitting ~17 of 15148 files rather than a whole class.

### The change

Only the array rules have a reason to run before the lowering, so the pre-lowering pass runs only those. The scalar rules and the ground-path collapse stay in `sizeReducing`, where they always ran.

| file | master | as merged | this change |
|:---|---:|---:|---:|
| flanagansaxe_shuf | 0.15 | 2.84 | **0.15** |
| flanagansaxe_shred | 0.15 | 2.93 | **0.15** |
| try3_true_functions_dwp_shuf | 1.87 | 5.84 | **1.85** |
| try3_true_functions_dwp_shred | 1.92 | 6.02 | **1.85** |
| unconstrained10 | 21.46 | 0.02 | **0.05** |
| unconstrained07 | 7.03 | 0.04 | **0.05** |
| sharing-is-caring/bench_30_96 | 3.06 | 0.01 | **0.03** |

Every regression gone, every win kept.

### Testing

The full 15148-benchmark QF_ABV sweep again, with `--check-sanity` on the candidate arm so every sat answer's model is constructed and checked: **no answer changed** (10193 sat, 4533 unsat) and **no counterexample was rejected**. The files whose outcome moved were re-measured with both arms plain at 60 s:

| | count |
|:---|---:|
| faster by >25% | 22 |
| **slower by >25%** | **0** |
| newly unsolvable | 0 |

Before this change that middle row was 18.

`ctest` is 129/129. One array-equality test's model moves again, because the rewrite order does; it stays self-consistent, which is what the test is for, so the pinned values are updated and the comment now names the property being pinned rather than an arithmetic coincidence between the values.

### DIFF
--- a/include/stp/Simplifier/RemoveUnconstrained.h
+++ b/include/stp/Simplifier/RemoveUnconstrained.h
@@ -77,6 +77,15 @@ class RemoveUnconstrained
   // something to do; see TopLevelSTPAux.
   bool arrayRules;
 
+  // Whether the ordinary (bit-vector) rules may fire in this pass.
+  // The pre-lowering call leaves them off: they run later in
+  // sizeReducing anyway, and eliminating scalars that early costs the
+  // dwp_formulas flanagansaxe pair ~15x -- the abstraction refinement
+  // loop then needs 22 rounds where it had needed 2. Only the array
+  // rules have to run before the array-equality lowering, so only they
+  // do.
+  bool scalarRules;
+
 public:
   RemoveUnconstrained(STPMgr& bm);
 	
@@ -88,8 +97,11 @@ public:
   // level, an already-encoded conjunct) even though this formula alone
   // mentions them once. Merged with the extensionality frozen set when
   // both apply.
+  // `arrayRulesOnly` restricts the pass to the array rules; see
+  // scalarRules.
   ASTNode topLevel(const ASTNode& n, Simplifier* s,
-                   const std::set<ASTNode>* alsoUntouchable = NULL);
+                   const std::set<ASTNode>* alsoUntouchable = NULL,
+                   bool arrayRulesOnly = false);
 };
 }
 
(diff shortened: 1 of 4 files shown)


TrevorHansen:
I can't think of any reason this is better. Seems it's randomly better on some benchmarks.

PR #893, opened 2026-08-15
Bump ABC to the last revision before it bundled its own SAT solvers

TrevorHansen:
Follows #892, which pointed the submodule at the fork.

The pin was abc `953930643`, from 2024-09-22. Upstream added kissat on 2025-03-04 and CaDiCaL a day later, and both sit in the module list in ABC's Makefile with no guard of the kind CUDD and readline have, so taking either would compile a third copy of CaDiCaL's symbols into `libabc` beside the ones STP and CryptoMiniSat already provide. That is what stops us following upstream today, and unpicking it needs the module trim, which is still to come.

There are 137 commits before that point which cost nothing to take, so this takes them. The new base is `c25bf7346`, 2025-03-03, the last commit with neither solver in the tree. Our four fixes rebase onto it without conflict — nothing upstream has touched the files they change — and the delta they apply is byte for byte the one they applied before.

## Why it is neutral, and how that was checked

Of the 137 commits, 23 touch the four ABC packages STP uses, and all 23 are in `aig/gia`, which STP barely enters — one include of `gia.h`, one of `giaAig.h`. None touch `aig/aig`, `opt/dar` or `sat/cnf`:

```
src/aig/aig      0 commits
src/opt/dar      0 commits
src/sat/cnf      0 commits
src/aig/gia     23 commits
```

Checked rather than assumed:

- 40 CNFs written from a fuzz corpus are **byte-identical** either side of the bump.
- The 40 query files carrying `(set-info :status ...)` still solve to the declared status.
- Full Release build is clean.

The previous pin is preserved as the tag `stp-at-953930643` in the fork, so the revision #892 pinned stays reachable regardless of what happens to the branch.


TrevorHansen:
Not as easy as I thought. The revision I selected builds with CUDD.

Issue #199, opened 2015-09-09
Copyright information in LICENSE


mdimjasevic:
In the `LICENSE` file only Vijay Ganesh is listed as the copyright holder. However, this is in contradiction with e.g. the copyright header in `bindings/python/stp/stp.py` where (C) 2014 Jurriaan Bremer is also listed. Also, the `AUTHORS` file lists quite a few authors, but it is not clear from the file if they're copyright holders.

What I would like to ask is to make it clear if an author is also a copyright holder (this doesn't have to be the case in general). Please clarify this in the `LICENSE` file and elsewhere if appropriate.

Making copyright information clear is useful when packaging STP, e.g. for Debian, which I've been working on per https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=789055.

hellovijay:
Dear Marko,

Currently only Vijay Ganesh is the copyright holder. (In the future, some
of the leader maintainers may become copyright holders as well.)

However, we may have different copyright holders for bindings.

One way you could package STP is to package it without the bindings, and
redirect users to the STP repo for bindings.

Cheers,
Vijay Ganesh.

mdimjasevic:
Hi Vijay,

There are other places where other authors are listed as well, e.g. `lib/AST/ASTmisc.cpp`. What is confusing about a file like that is that it lists a license, but it doesn't say who is the copyright holder. As a matter of fact, this file in particular with the AUTHORS tag suggests that both you and Trevor Hansen are copyright holders (though the copyright is not really stated), but then you said you are the only copyright holder of STP.

msoos:
Well.. I think Trevor Hansen should be contacted for this issue, to be on the safe side of the legal fence as he contributed a measurable chunk of the code. I wouldn't be surprised if he was fine with Vijay being the only copyright holder, but I think an email would do good.

As for others:  my contributions were relatively minor and I'm happy to give up copyright for them. I believe Dan, Ryan and Wi, would all do the same, considering that their contributions too, were relatively minor.

I know this is somewhat politicised issue (as all copyright is), so maybe you can give some guidance on how this works and whether we really need to contact some/all these people.

In good faith and with hopes that all this is easy to clear up,

Mate

[...]

PR #894, opened 2026-08-15
Record the licensing position: CONTRIBUTING.md, LICENSE and LICENSE_COMPONENTS

TrevorHansen:
STP has never recorded the terms under which contributions are accepted, and the copyright notice in `LICENSE` has not kept up with who has written the code. #199 asked for the copyright position to be made explicit and was never resolved. This is an attempt at the uncontentious parts of that.

### CONTRIBUTING.md

A short new file saying that contributions are accepted under the MIT Licence, that contributors confirm they have the right to submit the code on those terms, and that they keep the copyright in their work. It documents existing practice rather than changing it — no copyright assignment and no contributor licence agreement are required. GitHub links this file from the new issue and new pull request forms, so contributors see the terms at the point of submitting.

### LICENSE

**Copyright holders.** The block named seven people and omitted several substantial contributors; Andrew V. Teylu alone has over three hundred commits. Everyone with a real body of work in the revision history is now listed, and the list ends with a catch-all — with no contributor licence agreement every contributor keeps copyright in their own work, so no enumeration can be complete and it should not imply otherwise. Names were taken from `git shortlog`, and nobody has been removed.

**Subversion-era contributors.** `git shortlog` cannot see work predating the migration to git. The Subversion repository is still live on SourceForge — 1673 revisions, February 2008 to February 2013 — and the git history starts part way through it, in October 2008, having dropped roughly a fifth of the revisions on the way. Philippe Suter is invisible to git entirely, since all eight of his revisions predate its first commit, and Raphael Michel's ten Subversion revisions arrive as one. Peter Collingbourne, Xi Wang and Philip Guo are thin or absent likewise. They are now listed, except Raphael Michel: his Subversion work was string and regular expression support on a branch that never reached trunk, and his single git commit adds his own name to AUTHORS and nothing else, so no code of his is in STP. Names come from the SourceForge profiles behind the committer handles; year ranges from the Subversion log, which agrees with git wherever both hold a revision. One committer, `ar2rd`, is left out — a single path fix in a test file, and the profile behind the handle is gone, so there is no name to record.

**Year ranges** now match each holder's first and last commits, and the block is ordered by contribution rather than by the sequence names happened to be added in.

Two ranges are narrowed rather than widened, both belonging to the founders. Each read 2005 - 2015, and neither end date meant what it looked like: the 2015 came from Vijay Ganesh's last commit, `f02f1216`, titled "added David Dill to COPYRIGHT", so it recorded the year the entry was written rather than a year either of them worked on STP. Vijay Ganesh becomes 2005 - 2010, 2015, his code commits having stopped on 2010-09-03 with only that copyright edit after. David L. Dill ends in 2008.

Neither Dill nor Tim King has a commit in either repository — their work predates the Subversion history, which begins in 2008, and the CVS history it was converted from did not survive. Dill's 2008 end date rests on the project's own knowledge rather than on anything recoverable from the repositories.

**Moral rights.** A short assertion. These are personal to each author and cannot be transferred by a licence; the MIT text does not address them.

**Third-party components.** The inline list is replaced by a pointer to `LICENSE_COMPONENTS`. It had drifted — five components listed against eleven in `LICENSE_COMPONENTS` — and keeping two copies invites that again. The new wording also states what the old list left to inference: that third-party code is not covered by the grant above.

### LICENSE_COMPONENTS

The file was organised around code STP imports into its own tree, so the solvers it links fell outside it and went unrecorded. That is the wrong boundary for a notice file: the released asset is one static binary, and `release.yml` ships `LICENSE` and `LICENSE_COMPONENTS` beside it precisely because the MIT and BSD notices have to accompany a binary distribution. CryptoMiniSat, and the CaDiCaL and CadiBack that CryptoMiniSat 5.14 bundles into its library, are inside every released binary and were named nowhere.

There is now a section for the linked solvers, noting that which are present depends on the build configuration, with the licences of CryptoMiniSat, CaDiCaL, CadiBack and MiniSat. The CaDiCaL, CadiBack and MiniSat texts are reproduced verbatim from the sources the build uses. For CryptoMiniSat only the MIT part is reproduced, which is the part STP links, with a note pointing at the GPLv2 remainder.

Riss gets a note rather than a licence, since STP neither downloads nor distributes it — `USE_RISS` needs a checkout supplied by whoever is building — but it is not MIT licensed like the others, so a binary built with it does not redistribute on the same terms.

I also checked the existing entries against the versions in the tree: mimalloc, CLI11, ankerl::unordered_dense and SymFPU all still match, so nothing has drifted.

The `llvm-lit` entry was stale and is corrected. It reproduced the University of Illinois/NCSA licence, which LLVM moved away from in 2019; lit is now under the Apache License v2.0 with LLVM Exceptions, confirmed against the `LICENSE.TXT` and package metadata of an installed lit 16.0.6. The exceptions are recorded verbatim, with a pointer to the Apache text rather than two hundred lines reproduced for a tool that is installed separately and never distributed with STP. The old text was not wrong so much as superseded — LLVM still carries it under the heading "Legacy LLVM License".

### Worth reviewing

- The two narrowed ranges are the sharpest edit here, and one of them is Vijay Ganesh's own. Nothing else in the block has been shortened.
- The **copyright holder list** is the part that needs other eyes. Being named is a claim about ownership, and it sits against the statement in #199 that only Vijay Ganesh is a copyright holder. If that is still the intended position then that commit is wrong and should be dropped.
- The threshold for inclusion is arbitrary — copyright subsists in short contributions too. That is why the catch-all matters more than the enumeration. The bar is now four commits or revisions, applied across both eras, which names 24 people. Below it the tail is another 37 contributors, covered by the catch-all clause. `Robbepop` is recorded as the handle the contributor used — guessing at the person behind it would be worse than recording what the revision history says.
- GitHub currently reports STP's licence as `NOASSERTION` / "Other" rather than MIT, because the detector matches `LICENSE` against the canonical text and the extra material defeats it. Removing the component list helps, but the moral rights paragraph may keep it failing. Detection is computed from the default branch, so this can only be checked after merge. If the MIT label matters more, both paragraphs could move to a `NOTICE` file.

### Not included

No `Signed-off-by` / DCO requirement. As written the confirmation in CONTRIBUTING.md is passive. A DCO would put the assertion in the commit trailer, which is better evidence, but it needs a CI check and blocks pull requests, so it should be agreed separately.

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -15,20 +15,6 @@ Copyright (c)
     2018 - 2019       Norbert Manthey
     2008 - 2010       Stephen McCamant
     2013 - 2014       Florian Merz
-    2008              Philippe Suter
-    2013              Matt Arsenault
-    2013 - 2014       Edward J. Schwartz
-    2015 - 2023       Jiri Slaby
-    2016 - 2025       Martin Nowack
-    2020              Brian Foley
-    2017              Robbepop
-    2014              Stephan Falke
-    2021 - 2022       Hennadii Chernyshchyk
-    2013              Alexander Jesner
-    2014 - 2022       Jerry James
-    2009 - 2010       Xi Wang
-    2009 - 2012       Peter Collingbourne
-    2009              Philip Guo
     2007              Tim King
     and the other contributors recorded in the revision history of this
     project
--- a/LICENSE
+++ b/LICENSE
@@ -15,6 +15,11 @@ Copyright (c)
     2008 - 2010 Stephen McCamant
     2013 - 2014 Florian Merz
     2008 - 2010 Michael Katelman
+    2008 - 2008 Philippe Suter
+    2009 - 2009 Philip Guo
+    2009 - 2010 Xi Wang
+    2010 - 2012 Peter Collingbourne
+    2011 - 2011 Raphael Michel
     and the other contributors recorded in the AUTHORS file and in the
     revision history of this project
 
--- a/LICENSE
+++ b/LICENSE
@@ -4,7 +4,7 @@ Copyright (c)
     2005 - 2010, 2015 Vijay Ganesh
     2009 - 2024, 2026 Trevor Hansen
     2009 - 2022       Mate Soos
-    2018 - 2026       Andrew V. Teylu
+    2018 - 2026       Andrew Teylu
     2013 - 2017       Dan Liew
     2013 - 2025       Ryan Govostes
     2005 - 2008       David L. Dill
--- a/LICENSE
+++ b/LICENSE
@@ -1,35 +1,35 @@
 The MIT License
 
 Copyright (c)
-    2005 - 2015       Vijay Ganesh
+    2005 - 2010, 2015 Vijay Ganesh
     2009 - 2024, 2026 Trevor Hansen
-    2013 - 2025       Ryan Govostes
-    2013 - 2017       Dan Liew
-    2011 - 2013       Khoo Yit Phang
-    2005 - 2015       David L. Dill
     2009 - 2022       Mate Soos
-    2007              Tim King
     2018 - 2026       Andrew V. Teylu
+    2013 - 2017       Dan Liew
+    2013 - 2025       Ryan Govostes
+    2005 - 2008       David L. Dill
     2017              Felix Kutzner
+    2011 - 2013       Khoo Yit Phang
     2014              Jurriaan Bremer
+    2008 - 2010       Michael Katelman
     2018 - 2019       Norbert Manthey
     2008 - 2010       Stephen McCamant
     2013 - 2014       Florian Merz
-    2008 - 2010       Michael Katelman
     2008              Philippe Suter
-    2009              Philip Guo
-    2009 - 2010       Xi Wang
-    2010 - 2012       Peter Collingbourne
-    2013              Alexander Jesner
     2013              Matt Arsenault
     2013 - 2014       Edward J. Schwartz
-    2014              Stephan Falke
-    2014 - 2022       Jerry James
     2015 - 2023       Jiri Slaby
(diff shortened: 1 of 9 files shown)


LICENSE
```diff
@@ -1,34 +1,35 @@
...
-    2021 - 2022 Hennadii Chernyshchyk
+    2005 - 2015       Vijay Ganesh. Main codebase, project lead
+    2009 - 2024, 2026 Trevor Hansen
+    2013 - 2025       Ryan Govostes
+    2013 - 2017       Dan Liew
+    2011 - 2013       Khoo Yit Phang
+    2005 - 2015       David L. Dill
+    2009 - 2022       Mate Soos
+    2007              Tim King
+    2018 - 2026       Andrew V. Teylu
```
aytey: Can we drop the 'V.'? 

If stuff is inconsistent, that's my fault, but since I moved from Jones, I don't need the 'V' to make me unique :D

aytey: Thanks

TrevorHansen:
@stp/stp-team let me know if you have preferences.

aytey:
@TrevorHansen: what's your expectation as to when someone adds themselves to 'LICENSE'? Whenever they commit or only if they make, say, > 10 commits?

This all seems great to me, and my comment is a negative one, I just don't know if we want to define a "threshold" for when they get added or not.

TrevorHansen:
> @TrevorHansen: what's your expectation as to when someone adds themselves to 'LICENSE'? Whenever they commit or only if they make, say, > 10 commits?
> 
> This all seems great to me, and my comment is a negative one, I just don't know if we want to define a "threshold" for when they get added or not.

Sure, >=10 commits sounds reasonable. Let's do that.

PR #896, opened 2026-08-17
Raise FlattenKind's output budget above what real formulas reach

TrevorHansen:
CI has not completed on any branch since #891 landed. Every job that runs the test suite, master's own push runs included, sits until the six-hour ceiling and is cancelled. The build is fine; the hang is in `query-files`, on one file.

### What happens

The output budget added in #891 fires on `tests/query-files/generated-tests/form_32.var_64.bits_32.cvc`. Its widest flatten is **1054 operands**, against a cap of 1000.

Over budget the pass hands the operands back exactly as they arrived, so the `BVPLUS` never widens and the CNF that comes out is one the SAT solver does not finish:

| budget | result |
|---|---|
| 1000 | no answer after 5 minutes |
| 5000 | `Invalid.` in 9.4s |
| 20000 | `Invalid.` in 2.0s |
| 100000 | `Invalid.` in 1.8s |

A 5% shortfall in the cap is the difference between a two-second solve and one that does not return. The 600 QF_BV benchmarks the number was sampled against did not include the repository's own test files, which is how a cap that an in-tree test reaches looked like a cap nothing reaches.

### The value

A budget an ordinary formula can reach is worse than not having the pass at all, so it belongs far above what a formula plausibly needs rather than just above it. Stopping the explosion is all it has to do, and raising it costs the file it was added for almost nothing — on `QF_BV/Sage2/bench_16265.smt2`, the walk that grew past 30GB unbounded:

| budget | time | peak RSS |
|---|---|---|
| 1000 | 0.53s | 53MB |
| 100000 | 0.68s | 53MB |
| 1000000 | 14.2s | 178MB |

100000 is a hundred times the widest flatten seen in the suite and about 800KB of `ASTVec`, and the pathological file does not notice it. A million does, so the margin stops there.

### Verification

The `.cvc` file is already in the suite, so it is the regression test: at 1000 the run does not terminate.

- `query-files`: was hanging → **591 passed, 2 unsupported, 0 failed**, 9.8s
- full `ctest`: **130/130 passed**, 45s

### Not fixed here

The cliff is still a cliff. When the budget does fire, the result is not "flattened a bit less" but "not flattened at all", and this file shows that can be catastrophic rather than the under-2% the original measurement suggested. Bounding the walk without the discontinuity — expanding each distinct same-kind child once instead of once per path, which is linear in the DAG and needs no budget — looks like the better shape, but it is a redesign of the traversal and wants its own change and its own review. This one only moves the number so CI can run.

### DIFF
--- a/lib/AST/ASTmisc.cpp
+++ b/lib/AST/ASTmisc.cpp
@@ -512,7 +512,17 @@ void FlattenKindNoDuplicates(const Kind k, const ASTChildren& children,
 //
 // Returns false if the budget was passed, in which case flat_children holds a
 // prefix of the operands and must be discarded: a truncated sum is not the sum.
-const size_t FLATTEN_OUTPUT_BUDGET = 1000;
+//
+// The budget only has to stop the explosion, so it is set far above what a
+// formula plausibly needs rather than close to it. Declining to flatten is not
+// a small loss: generated-tests/form_32.var_64.bits_32.cvc wants 1054 operands
+// at its widest, and capping it below that leaves BVPLUS nested, which turns a
+// two-second solve into one that has not finished after five minutes. A cap
+// tight enough to be reached by an ordinary formula is worse than no pass at
+// all. Against that, raising it costs the pathological file almost nothing:
+// on QF_BV/Sage2/bench_16265.smt2 the walk that used to reach 30GB is abandoned
+// at 0.68s and 53MB here, against 0.53s and 53MB at a thousand.
+const size_t FLATTEN_OUTPUT_BUDGET = 100000;
 
 bool FlattenKind(const Kind k, const ASTChildren& children, ASTVec& flat_children, int depth)
 {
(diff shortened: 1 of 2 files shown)


TrevorHansen:
Corpus evidence for the value, since the original number came from a sample and so does this one.

400 QF_BV files — 150 from Sage2, where the pathological file lives, plus 250 across the whole logic — each run at budget 1000 and at 100000, with a third pass recording the widest `flat_children.size()` actually reached. 368 produced a width.

| widest flatten | files |
|---|---|
| median | 7 |
| p90 | 78 |
| p99 | 206 |
| max | 1494 |
| over 1000 (old cap fires) | 1 of 368 |
| over 100000 (new cap fires) | 0 of 368 |

The distribution is nowhere near either cap. The single file that crossed the old one, `Sage2/bench_11118.smt2`, paid nothing for it: 2.66s at 1000 against 2.65s at 100000.

That is the point worth drawing out. The old cap really did fire on ~0.3% of files, and really did cost almost nothing when it fired here — the same result the original sampling got. What the rate does not tell you is the worst case, because the cost of declining to flatten is not bounded by how often it happens. One file crossed the cap by 49% and paid 0.4%; the `.cvc` file in the test suite crossed it by 5% and did not finish.

Timing across the sample shows no signal either way: 404.6s total at 1000 against 398.8s at 100000 over the 367 files where both arms completed. The largest individual swings are on files whose widest flatten is 0 or 2 — which neither budget can affect — so ±3s there is the noise floor of a loaded box, not an effect.

CI on this branch is 18/18 green. `gcc` completed in 11m22s and `gcc (minisat only)` in 9m7s, against the six-hour timeout they had been hitting; `CNF determinism` ran for the first time since #891, having previously been skipped because the jobs it depends on never finished.

PR #907, opened 2026-08-19
Compress the unconstrained-divisor comments

TrevorHansen:
Comment-only follow-up to #905: the derivations behind the FP_DIV arm are compressed to the constraints they establish (the full arguments live in #905's description and the exhaustive tests). No code changes — the diff has no non-comment lines.


TrevorHansen:
Should be part of a larger sweep to reduce verbosity.

PR #916, opened 2026-08-20
Order a distinct whose operands occur nowhere else

aytey:
A `(distinct x1 ... xn)` over variables that appear in no other assertion is invariant under every permutation of those variables, so of the `n!` assignments that differ only by which variable got which value, `n!-1` are copies. The solver searches them all.

Requiring `x1 < x2 < ... < xn` keeps one copy of every answer and replaces `n(n-1)/2` disequalities with `n-1` comparisons — and, more to the point, states an ordering the bit-blaster would otherwise have to find. Three hundred unconstrained 32-bit variables under one distinct go from **172s to 0.2s** (RelWithDebInfo, CaDiCaL, this machine). That is the only measurement quoted anywhere in the branch: the header comment, the fixture comment and the commit message all give the same pair of numbers and the same build.

The parser records each group before it dissolves it, because a clique of disequalities does not say which variables were interchangeable and nothing downstream can recover it. The record is manager state, so the two-field `DistinctGroup` is declared next to the member that holds it rather than in the Simplifier header that reads it — a core header does not need to include a Simplifier one to name a struct. `~STPMgr` releases the vector in its body, alongside every other member that holds nodes and for the same reason: releasing a node calls back into the manager to unindex the symbol's name, and the implicit member-destruction phase runs after that index has already been destroyed.

Deciding is left to a pass over the assembled formula, since only the whole formula shows whether the operands really do occur nowhere else. That pass runs per solve in the batch pipeline, so an assertion added after one `check-sat` simply stops the rewrite applying at the next — an ordering is never inherited by a formula that did not earn it. It is batch-only, and deliberately so: the persistent driver's clauses survive the solve that produced them, so a chain encoded into a block would still be in force once a later assertion destroyed the symmetry behind it. Under `--incremental=on` nothing is ever ordered.

The guard has three cases to get right, and two of the three are wrong answers if it fails. An operand used outside the distinct is not interchangeable, and ordering it would report `unsat` where the answer is `sat`. Two groups sharing an operand are each symmetric alone and not symmetric together, and ordering both would again turn a satisfiable query unsatisfiable. The third is a negated distinct, where the chain becomes the weaker claim: still equisatisfiable, but the model would not be a model of the input, so the rewrite declines rather than publish one it cannot stand behind — and that block answers `sat` either way, so it is pinned by the absence of the stats line rather than by its verdict. The comments in the fixtures say which is which.

The models are checked, not just asserted. Four two-bit variables under one distinct have exactly one increasing assignment, so the last block of `distinct-ordering.smt2` reads the model back and pins all four values: they are a model of the unrewritten distinct, all four differing, and the chain leaves the solver no other choice, so the check is stable rather than a snapshot of what CaDiCaL happened to pick. It is also where the user-visible consequence is recorded: a published model is now canonical where it used to be any of the 24 permutations.

## Verification

`distinct-ordering.smt2`, `distinct-ordering-boundaries.smt2` and `incremental-tests/distinct-ordering-batch-only.smt2` all **fail on master**, and the model block alone fails on master independently of the stats line. The new `UserDefinedFlags_Test` case does not compile against master's headers at all.

**135/135** `ctest` in RelWithDebInfo and in Release with `-DWERROR=ON`; the latter with a `ccache`-disabled rebuild of all 135 translation units, which produced no warning from any STP source.

`valgrind` is clean over all three fixtures. The destructor ordering above does not show up there on its own, because the manager's other tables are emptied in the destructor body and that makes the late release benign in practice; removing one of those body clears in a probe build turns it into 49 error contexts, every one of them arriving through `~vector<DistinctGroup>` and none of them present when the query holds no `distinct`, and the same probe build with this commit's one-line release reports none.

3000 randomised differential cases, 680 of which exercised the rewrite, agree with `--distinct-ordering false`, and all 2096 reported models re-check as models of the unrewritten assertions.

## The default

The flag `--distinct-ordering` defaults to **on**, so this changes what every user of the batch pipeline gets, and it reroutes the existing `sample-smt-tests/arity1.smt2` — still `unsat`, now by a chain over five two-bit variables rather than by pigeonhole. **Worth a maintainer's opinion on the default.** If it stays on: it is registered under "Simplifications" rather than with the refinement loops, `--distinct-ordering false` switches it off, and `--disable-simplifications` clears it along with the rest of that stack (asking for both at once is refused, the same way the other fourteen are).

### DIFF
--- a/include/stp/STPManager/STPManager.h
+++ b/include/stp/STPManager/STPManager.h
@@ -56,6 +56,16 @@ enum class FPSpecial
   MinusZero,
 };
 
+// One (distinct t1 ... tn) as the parser saw it, before it was dissolved into
+// pairwise disequalities: the operands in source order, and the node the
+// dissolution produced. Declared here rather than beside the pass that reads
+// it, so that a core header does not have to include a Simplifier one.
+struct DLL_PUBLIC DistinctGroup
+{
+  ASTVec operands;
+  ASTNode emitted;
+};
+
 /*
  * STP Node Manager. Tools for managing AST nodes.
  */
@@ -111,6 +121,12 @@ private:
 public:
   HashingNodeFactory* hashingNodeFactory;
   NodeFactory* defaultNodeFactory;
+
+  // (distinct ...) groups as the parser saw them, before it dissolved each
+  // into pairwise disequalities. Kept so a later pass can ask whether the
+  // operands are symmetric and, if they are, state an order instead. Cleared
+  // whenever the assertion stack is.
+  std::vector<DistinctGroup> distinctGroups;
 
   // State of the array-equality (extensional arrays) decision procedure:
   // solve-local equality records, the complete per-solve array graph, and
--- a/include/stp/STPManager/UserDefinedFlags.h
+++ b/include/stp/STPManager/UserDefinedFlags.h
@@ -181,6 +181,16 @@ public:
   // ARRAY_EQ, which is lowered only after the complete query is assembled.
   bool enable_array_equality = false;
 
+  // Replace a (distinct x1 ... xn) whose operands are variables occurring
+  // nowhere else with the strict chain x1 < x2 < ... < xn. Every permutation
+  // of such operands maps the formula to itself, so fixing one of the n!
+  // orderings loses no answer, and n-1 comparisons replace n(n-1)/2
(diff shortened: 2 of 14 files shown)


TrevorHansen:
I never thought of this. Nicely done. 

Taking it further and implementing a first-class `distinct` predicate is appealing to me?

PR #959, opened 2026-08-23
Say why the SymFPU submodule is nested a level down

TrevorHansen:
`lib/extlib-symfpu/symfpu` is the one submodule that does not sit directly at its `extlib-` directory, which reads as an inconsistency next to `lib/extlib-libbf` until you know why: SymFPU's headers include each other as `symfpu/core/...`, so the include root has to be a directory *containing* one named `symfpu`. `extlib-symfpu` is that root, and is what `SYMFPU_INCLUDE_DIRS` names.

LibBF needs no such wrapper because its header is included as `<libbf.h>`, so the submodule directory is itself the include root.

This was part of #958, but that merged into #957's branch before I pushed it, so it did not reach master with the rest. Documentation only.


TrevorHansen:
Trivia.

PR #967, opened 2026-08-23
Fetch and build the dependencies rather than requiring them beforehand

aytey:
Every dependency STP does not contain was something you had to arrange beforehand: initialise four submodules, then run some subset of eight shell scripts under `scripts/deps/`, then configure — and configuration failed, sometimes obscurely, if you had guessed wrong about which. This makes the build fetch and build them itself.

From a clean checkout, in full:

```
git clone https://github.com/stp/stp
cd stp
./configure.sh --auto-download
cmake --build build -j$(nproc)
```

There are no submodules left, and `.gitmodules` is gone.

### The ladder

Each dependency gets a `cmake/FindX.cmake` that works down the same rungs, described once in `cmake/deps-helper.cmake`:

0. `<X>_DIR` names a copy explicitly — use it, or fail. Never fall through: silently building something other than what the caller named is worse than stopping.
1. The system, which includes anything on `CMAKE_PREFIX_PATH` — so an installed copy, or one another build directory left in `STP_DEP_DIR`, or the `deps/install` the remaining setup script writes.
2. An ExternalProject this build directory already downloaded.
3. `ENABLE_AUTO_DOWNLOAD` — fetch and build it.
4. Fail, naming both `--auto-download` and `-D<X>_DIR=`.

Nothing reaches the network unless it is asked to. `--auto-download` is off by default and the error that mentions it is half the point.

Dependencies install into `STP_DEP_DIR`, which also goes on `CMAKE_PREFIX_PATH`. That is the whole of the sharing mechanism: point several build directories at one path and only the first builds anything, because the rest find the artefacts at rung 1 and create no ExternalProject at all — so they do not need `--auto-download` either. ExternalProject's own scratch and stamp files stay per build directory, because those are mutable state two builds sharing would corrupt. A `deps` target builds the dependencies and nothing else, for filling such a directory ahead of time.

### Two tools, and which gets used

**ExternalProject for what STP links; FetchContent for what STP compiles.**

Most of these are a header set and an archive — LibBF, CaDiCaL, MiniSat, Riss, SymFPU, CLI11, ABC — and nothing of STP's build reaches inside them. They are ExternalProjects, built at build time with STP's own compiler, flags, launcher, sysroot and `-fPIC`, forwarded from one place.

mimalloc, GoogleTest and `ankerl::unordered_dense` are different: STP configures mimalloc's build, compiles GoogleTest with its own flags, and reaches `unordered_dense` through a project-wide include from `stp/AST/AST.h`. Each needs its source to exist during *this* configure, which is what FetchContent does and ExternalProject cannot.

CryptoMiniSat is the exception with no build rung at all, and `cmake/FindCryptoMiniSat.cmake` says why at length: it reaches STP as a *CMake package*, and an ExternalProject writes that package at build time — after the configure that has to read it. Installing it by hand instead would leave a copy no later build could find the same way, and would make STP's own `STPConfig.cmake` name a package that was not there. It gets rungs 0, 1 and an error that names `setup-cms.sh`. It is also the dependency where this matters least: distributions package it.

### What changes for an existing build

**The SAT backends are chosen by flag, not by what is installed.** CryptoMiniSat used to be linked in whenever it happened to be present, so two people running the same command on different machines got solvers with different backends in them and neither could tell from the command. `USE_CRYPTOMINISAT` now defaults to `OFF`; `AUTO` restores the old behaviour by name; `ON` makes a missing one a configuration error. `NOCRYPTOMINISAT` and `FORCE_CMS` still work, warn once, and translate — `NOCRYPTOMINISAT=OFF` meant "do look for it", so it becomes `AUTO`.

Something then has to be on by default or a plain `cmake ..` cannot solve anything, and `USE_CADICAL` is it: no system library, the backend the documentation already recommends for hard bitvector problems, and one the build can now produce.

**`-DENABLE_ASSERTIONS=ON -DCMAKE_BUILD_TYPE=Release` no longer silently produces a build without assertions.** A plain `option()` cannot tell "the user asked for OFF" from "nobody said anything", so the Release branch overwrote the request. `ENABLE_ASSERTIONS`, `USE_CRYPTOMINISAT` and the new `BUILD_MANPAGE` are three-valued.

**A minimum CryptoMiniSat.** 5.11, which is the oldest release anything in this repository builds and in which every method `lib/Sat/CryptoMinisat5.cpp` calls exists. Older ones are refused for being untested here, not for being known broken, and the error says so.

**`stp/Sat/Cadical.h` includes `<cadical/cadical.hpp>`** rather than `"src/cadical.hpp"`, a path only a *checkout* has — which is why `CADICAL_DIR` had to name one, and why STP could not be built against an installed CaDiCaL at all. Which in turn lets the version probe be fixed: it read a `VERSION` file from the checkout root, so anything found any other way read as `unknown` and quietly lost `--cadical-factor`. An installed CaDiCaL is now asked directly, by compiling and running `CaDiCaL::Solver::version()`.

**`find_package(ZLIB REQUIRED)`** under `USE_MINISAT`. It was unqualified and its result unchecked, so a missing zlib surfaced minutes later inside a MiniSat header.

**`lit` is installed into a virtual environment** under the build directory when there is none, so `ENABLE_TESTING` on an unprepared machine no longer stops with "install it by e.g. `pip install lit`". Nothing reaches the system interpreter. OutputCheck's `not` helper also gets a `python` to find — its shebang is `#!/usr/bin/env python`, which resolves on no system that ships only `python3`, and every error-path test is a `not %solver ...`.

**An installed *static* STP is self-contained.** A static libstp does not absorb the archives it links, so CMake propagates them into `STPTargets.cmake` — naming the path each was linked from, which meant `find_package(STP)` worked only on the machine that produced the package and only until its build tree was removed. They are installed beside libstp now and named relative to the install prefix.

### Fallout

Six of the eight scripts under `scripts/deps/` are gone. `setup-cms.sh` stays, because CryptoMiniSat is still the one dependency the build does not produce, and `cache-key.sh` stays but no longer resolves a revision over the network — it only did that because OutputCheck was cloned without a pin, and OutputCheck is pinned now.

CI jobs name one cache path instead of four and configure with `--auto-download` instead of running a list of scripts whose contents had to agree with the cmake flags further down. Three get more than tidying:

* **Both Windows jobs stop building dependencies by hand.** LibBF was fetched with the script's clone half and compiled separately with `cl`, and again with `gcc` for MinGW, because the script's compile half is `cc` and `ar`. `cmake/FindLibBF.cmake` supplies LibBF with a `CMakeLists.txt` of its own, so whichever compiler the job is already using builds it. The MinGW job's hand-written CaDiCaL build goes the same way — it existed to build `libcadical.a` alone, since CaDiCaL's default target also builds `mobical` and `mobical` does not compile on MinGW, which is what the Find module does anyway.
* **The 32-bit job** keeps its point and loses its scaffolding: LibBF's 32-bit limb build is exercised there and nowhere else, and it now gets that by being compiled with the job's own toolchain rather than with whatever `cc` a script found.
* **The leg that checks `CADICAL_DIR` outranks CryptoMiniSat's bundled CaDiCaL** clones one inline. Its whole point is that the second CaDiCaL is somewhere the build did not put it.

Both Dockerfiles are repaired rather than updated. Neither named a SAT backend, which stopped working when CaDiCaL became the default and had to be *found*, and neither ever fetched LibBF, which has been required since well before this — they could only build from a context that happened to carry a hand-built `deps/libbf`.

`patches/symfpu/` and the sixty lines of `stp_apply_vendored_patches()` are gone. All of that existed because the patches were applied to a submodule working tree that persisted between configures, so each had to be tested for having already been applied. An ExternalProject unpacks a fresh tree and patches it once.

### Deliberately not done

ABC and mimalloc could have stayed submodules and I argued they should; they are fetched at the maintainer's direction, and the fork workflow survives — `-DABC_DIR` at a build of a clone, or `-DFETCHCONTENT_SOURCE_DIR_MIMALLOC` at a checkout, uses it in place and fetches nothing.

Bison, flex and python3 stay toolchain. zlib and GMP stay system-found: both are transitive, of MiniSat and CryptoMiniSat respectively, and STP includes neither.

### What has not been exercised

Everything here was verified on Linux with gcc — from a tree containing nothing but STP's own sources, `./configure.sh --auto-download` gives 155/155 ctest, and each dependency's rungs were checked individually, as were static, `WERROR`, and each of the four backend configurations.

**Nothing has been run under MSVC or MinGW**, and those two jobs are where the most code was removed. That is the part of this that CI is being asked about.


aytey:
@TrevorHansen : FYI, this was the big restructure of the deps

PR #993, opened 2026-08-25
Remove the unused surface of extlib-constbv

TrevorHansen:
This library came with a much wider API than STP uses. A linker garbage-collection pass (static build, `-ffunction-sections -Wl,--gc-sections`) plus whole-repo reference checks over tools, tests, bindings and examples found 48 exported functions with no caller at all — including a matrix-multiplication family in a bitvector library. Removed along with their header declarations and internal-only helpers that the removals orphan (the closure was computed, e.g. `BitVector_Interval_Empty`'s only callers were the also-dead `Insert`/`Delete`).

Also gone: `BitVector_BYTENORM`, a 1KB `const` table defined in the header — so every translation unit that includes `constantbv.h` carried its own copy — whose single reader was the dead `Set_Norm`.

Deliberately untouched:

- `BitVector_Block_Store` and everything reachable from the live API (`shift_left/right`, `Word_Insert/Delete`, `BitVector_compute`, `BIT_VECTOR_reverse`, …) — the closure stops exactly where a live caller appears.
- The boot machinery. The load-time self-boot (`boot_at_load`) runs once on the loading thread, but the tables it fills are thread-local, so the explicit `BitVector_Boot()` calls around the tree are still required and none were removed.

~1500 lines. Whole-tree build (which links every remaining reference) and the full test suite pass (166/166).


TrevorHansen:
I don't think it makes sense to chop the parts of constbv that we aren't using.

PR #997, opened 2026-08-25
Remove dead code from FloatBlaster, Extensionality, UF and Incremental

TrevorHansen:
Continues #991–#996, same method. ~120 lines, mostly write-only state:

- `ExtEvent` carried three fields (`seq`, `indexValue`, `accessValue`) that every event write filled in and nothing — including the extensionality unit tests, which read the other five fields — ever read. The `event()` helper loses two parameters, and the `!recordEvents` fast path stops maintaining a sequence counter nothing sees.
- `UFContext::allById_` was a per-declaration `std::map` that was inserted into and cleared but never probed (`byIdentity_` and `activeByName_` are the lookups actually used).
- `UFTheoryAdapter::candidateChecks()` was a pure virtual forcing both adapters to carry an override, plus a counter incremented per candidate check — with no caller through the base or on either adapter.
- `FloatBlaster::operandFormat` (both overloads) survived only in two comments after `deriveFPFormat` replaced the mechanism; the comments now name the replacement.
- One provably-unreachable `requireSameFormat` call in `FloatBlast`'s ITE arm: the line directly above returns when the sorts differ, which is the exact condition the check re-tests.

One near-miss worth recording: `UFConcreteValue::scalar` looked dead under a qualified-name search but is called unqualified by the other factory functions in the same class — it stays.

No reachable code changes. Full test suite passes (166/166).

### DIFF
--- a/include/stp/Extensionality/ExtChecker.h
+++ b/include/stp/Extensionality/ExtChecker.h
@@ -337,14 +337,11 @@ struct ExtEvent
     CONFLICT,
     WITNESS_CHECK
   };
-  int seq;
   Kind kind;
   const char* rule;
   ASTNode source; // null for seeds / witness checks
   ASTNode destination;
   size_t access;
-  ASTNode indexValue;  // null on SKIP_SEEN
-  ASTNode accessValue; // null on SKIP_SEEN
 };
 
 struct ExtCheckResult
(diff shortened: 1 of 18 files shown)


aytey:
@TrevorHansen is there a way to turn this into a CI check?

As in, to detect this dead code at commit time

TrevorHansen:
> @TrevorHansen is there a way to turn this into a CI check?
> 
> As in, to detect this dead code at commit time

I imagine such things exist. Feel free to install. Otherwise, I'm happy to occasionally set my AI to run a cleanup.

PR #998, opened 2026-08-25
Choose the CNF effort from the size of the AIG

aytey:
The CNF effort scale trades generation time for a smaller CNF, and the default makes that trade on every query. It is not always worth making: cut enumeration and technology mapping are superlinear in the AIG, while the extra clauses `very-low` leaves behind cost a modern SAT solver very little. On a large AIG the minimising is most of the work.

A floating-point query with three `fp.div` and two `fp.sqrt` — 2,715 bytes of SMT-LIB — blasts to 1.3M AIG nodes and 1.7M clauses. STP spends **1857ms generating that CNF** and MiniSat then spends **137ms solving it**.

This adds an `auto` level that reads `Aig_ManNodeNum()` at the point of conversion and drops to `very-low` at or above `cnf_auto_threshold` nodes. It is the new default; every fixed level stays reachable and behaves exactly as before.

### The threshold is 200k, and high on purpose

Timed net of the ~9.5ms a process spends starting before it solves anything — which on small queries is most of the wall clock — `very-low` is the better choice at almost every size on the corpus measured. That does not make `very-low` right everywhere; it makes the crossover **a property of the workload rather than a constant**.

So the heuristic overrides the effort only where the evidence is unambiguous. At 200k it switches the queries whose circuits are large enough for minimising to be most of their cost — six of the measured set, every one a gain, geometric mean 0.45. A lower threshold measured better on totals and worse on regressions: at 32k, fifteen gains against three losses.

Callers who have measured their own workload can move it, with `--cnf-auto-threshold` or through the `CNF_AUTO_THRESHOLD` interface flag. That flag is **appended after the published prefix** of `ifaceflag_t` rather than filed beside `CNF_GENERATION_EFFORT` where it belongs by subject — these are integers callers compile into their binaries before loading a newer libstp, and `tests/api/C/counter-enum-abi.cpp` keeps the same rule for `stp_counter_t`.

### Exact refinement is deliberately excluded

`auto` was calibrated on whole-query conversion, where the CNF is built once and discarded, so trading clauses for generation time is free. Refinement encodes into a live solver where the clauses stay. A 64-bit `BVMULT` clears the threshold, and letting `auto` decide there left about 20% more clauses in the solver for the rest of the search — 127,987 against 106,753, which `BVExactEncoderTest.TheMappedEncodingCostsFewerClausesThanWrittenOutGates` measures directly.

An explicitly chosen level still reaches both paths, as `TheSelectedCNFStrategyReachesExactEncoding` requires.

### What it is worth

On the queries KLEE sends STP for a driver where STP loses to Bitwuzla, the deficit **halves: 5.61x to 2.88x**.

End to end across a whole floating-point benchmark corpus — 255 drivers, 129 comparable — the effect is small: **0.975** solver time, faster on 46 drivers and slower on 52, with the mean carried by the size of the gains rather than their number. An earlier revision of this description claimed 0.914 from a fourteen-driver sample; that sample was biased toward drivers chosen because they were affected. **The SMT-level result is the one to read.**

### Tests

`tests/query-files/misc-tests/cnf-generation-effort-auto.smt2` covers the option, both branches of the decision, the fixed levels staying silent, and the unknown-level error. `cnf_effort_flag` gains `TheDefaultIsAuto` and `TheAutoThresholdIsReachableThroughTheCAPI`, the latter checking that zero is honoured (every AIG is at or above it) and that a negative value is refused rather than wrapping to a threshold nothing could reach.

160 tests pass locally. CI could not run this suite until #999 landed.

### DIFF
--- a/include/stp/STPManager/UserDefinedFlags.h
+++ b/include/stp/STPManager/UserDefinedFlags.h
@@ -625,16 +625,43 @@ public:
   // LOW, HIGH and VERY_HIGH use ABC's newer Mf_ManGenerateCnf() at LUT sizes
   // 3, 6 and 8; its cost grows steeply with LUT size for little further gain
   // past 6.
+  //
+  // AUTO picks between VERY_LOW and MEDIUM from the size of the AIG, because
+  // the trade the scale offers is only worth making when the SAT solver is
+  // what costs. Minimising the CNF is itself work, and on a large AIG it is
+  // most of the work: a floating-point query with three fp.div and two
+  // fp.sqrt blasts to 1.7M clauses, of which MEDIUM spends 1857ms generating
+  // and MiniSat then spends 137ms solving. VERY_LOW generates the same query
+  // in a fraction of that, and the larger CNF it leaves behind costs the
+  // solver almost nothing.
   enum CNFEffort
   {
     CNF_EFFORT_VERY_LOW = 0,
     CNF_EFFORT_LOW,
     CNF_EFFORT_MEDIUM,
     CNF_EFFORT_HIGH,
-    CNF_EFFORT_VERY_HIGH
+    CNF_EFFORT_VERY_HIGH,
+    CNF_EFFORT_AUTO
   };
 
-  enum CNFEffort cnf_effort = CNF_EFFORT_MEDIUM;
+  enum CNFEffort cnf_effort = CNF_EFFORT_AUTO;
+
+  // AIG AND-node count at or above which AUTO drops to VERY_LOW.
+  //
+  // High on purpose. Over a floating-point corpus, timed net of the ~9.5ms a
+  // process spends starting before it solves anything, VERY_LOW is the better
+  // choice at almost every size -- so the honest reading is that the crossover
+  // is a property of the workload rather than a constant, and that this
+  // heuristic should only override the effort where the evidence is
+  // unambiguous. At 200k it switches the queries whose circuits are large
+  // enough for minimising to be most of their cost -- six of the measured set,
+  // every one a gain, geometric mean 0.45 -- and leaves everything else alone.
+  //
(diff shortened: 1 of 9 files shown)


aytey:
## Correction to the KLEE figure above

The PR body cites solver time **0.914** against stock STP, measured over fourteen drivers. A full sweep of the same corpus — 255 drivers, 129 comparable under both configurations and a Bitwuzla reference — puts it at **0.975**, and the distribution is worth stating plainly:

| | |
| --- | --- |
| geometric mean per driver | 0.975 |
| drivers faster | 46 |
| drivers slower | 52 |

More drivers regress than improve; the mean falls below 1 because the gains are larger than the losses (`atan2f` 0.42, `N_VWL2Norm` 0.43, `gsl_sf_multiply_e` 0.44, against `cs_di_lsolve` 2.00 and `lroundf` 1.74). Per library: sundials 0.864, gmp 0.952, gsl 0.972, blis 0.984, cxsparse 1.009, openlibm 1.022, fftw 1.036.

Against Bitwuzla the standing improves slightly, 0.858 to 0.836.

The fourteen-driver figure was a biased sample — those drivers were picked partly because they were affected. The SMT-level results in the body (2.26x on large-AIG queries, 5.61x to 2.88x on the queries KLEE sends STP for one driver) were measured on their own terms and are unchanged.

**This leaves a real open question about the threshold.** 32k sits inside the band where my two calibration datasets disagree: `very-low` won on every query above 28k on one corpus, `medium` won between 10k and 40k on another. The regressions above are consistent with the threshold being too low, and the large gains all come from queries far above it — 257k and 1.3M nodes. A higher threshold would likely keep the wins and drop the marginal cases, but I have not measured that yet, so I am not proposing a number.

Happy to hold this until that is settled.

PR #1034, opened 2026-08-28
Add a build mode that uses nothing from outside the build directory

TrevorHansen:
`-DSTP_DEPS_LOCAL_ONLY=ON`, or `./configure.sh --local-deps`, for a build that links nothing it did not produce itself.

## What it does

It removes rung 1 of the ladder in `cmake/deps-helper.cmake` — the plain `find_path`/`find_library`/`find_package(CONFIG)` that answers with an installed copy — from all eight Find modules, and takes the source tree's `deps/install` off `CMAKE_PREFIX_PATH`. ABC, CaDiCaL, CLI11, LibBF, MiniSat, Riss and SymFPU are then built into `STP_DEP_DIR`, which is inside the build directory unless it was pointed elsewhere.

Rung 0 is deliberately untouched: `-DABC_DIR=` and friends name a copy, which is an answer rather than a search, and a caller who names one has said which to use. `--local-deps` implies `--auto-download` unless `--no-auto-download` is also given, since local-only against an empty build directory can only fail.

## Why

On a machine whose `/usr/local` holds the CaDiCaL that CryptoMiniSat >= 5.14 installs beside itself, a plain configure says

```
-- CaDiCaL 2.1.3 predates 3.0.0: --cadical-factor will be unavailable
-- Found CaDiCaL 2.1.3: /usr/local/lib/libcadical.a
```

and the build quietly has no bounded variable addition. `docs/building.rst` already described that shadowing; `CADICAL_DIR` pins one lookup past it and does nothing about the rest. This is the switch that turns the rung off wholesale.

## The two dependencies that need more than a guard

**CryptoMiniSat** reaches STP through the CMake package it installs, which an ExternalProject could not write until after the configure that has to read it — so its rungs 0 and 1 are one `find_package` call with nothing behind them. Under the option it is used only if `cryptominisat5_DIR` names a copy; `-DUSE_CRYPTOMINISAT=ON` without one is a configuration error saying exactly that, and left at AUTO it simply goes off.

**LibBF**'s `LIBBF_DIR` defaulted to the source tree's `deps/libbf`, which made `-DLIBBF_DIR=` naming that same path indistinguishable from saying nothing — and the two are different things, since naming a copy is an answer and a directory that happens to hold one is a place to look. The cache entry now defaults to empty, with `deps/libbf` applied as a separate fallback that the option skips. Behaviour with the option off is unchanged, including in build directories configured before this, whose cached value simply persists.

## Not a CMake global

`CMAKE_FIND_USE_CMAKE_SYSTEM_PATH=OFF` and `CMAKE_IGNORE_PREFIX_PATH` are the obvious alternative and do not work: the same search also resolves flex, bison, zlib, GMP via pkg-config and python3, none of which STP can build for itself, so configuration fails rather than being confined. Hence a per-dependency skip. (`CMAKE_FIND_USE_PACKAGE_REGISTRY=OFF` is safe on its own and worth knowing about, but it only covers the config-package channel.)

## Testing

Configure-only, on a host whose `/usr/local` holds CaDiCaL 2.1.3, CryptoMiniSat 5.14.7 and MiniSat:

| configuration | ABC | CaDiCaL | CryptoMiniSat | MiniSat |
| --- | --- | --- | --- | --- |
| default, `--auto-download` | built | found 2.1.3, installed | found 5.14.7 | found, installed |
| `--local-deps` | built | built 3.0.1 | off | built |
| `--local-deps`, `USE_CRYPTOMINISAT=ON` | — | — | refused, message names `cryptominisat5_DIR` | — |
| `--local-deps`, `-Dcryptominisat5_DIR=…`, `USE_CADICAL=OFF` | built | off | found 5.14.7 | — |
| `--local-deps`, `-DCADICAL_DIR=…` | built | found at that checkout | off | — |

LibBF's rung 0, which now has a fallback rather than a default:

| `deps/libbf` present | `LIBBF_DIR` | `--local-deps` | outcome |
| --- | --- | --- | --- |
| yes | unset | off | found at `deps/libbf`, unchanged |
| yes | unset | on | ignored, built |
| yes | set to that same path | on | found — an explicit answer still wins |
| no | unset | off | built |

End to end: a cold `./configure.sh release --local-deps --ninja` followed by a build produced a working `stp` whose only libraries from outside the build tree are `libstdc++`, `libgcc_s`, `libc` and `libm`, and whose `--version` reports `cadical 3.0.1` where the default build on the same host reports `2.1.3`. Both rung-4 error messages were checked, for a required dependency and for an optional one.


TrevorHansen:
@aytey Often I like the build to be self-contained so that it's obvious what's being linked in. Happy to handle this differently if t here's a better way.

aytey:
> @aytey Often I like the build to be self-contained so that it's obvious what's being linked in. Happy to handle this differently if t here's a better way.

Works for me, if it works for you!

Does it work with ccache when you work like this? I just got sick of continually recompiling the deps for every worktree 🤣

PR #1043, opened 2026-08-29
Add an in-house AIG package

TrevorHansen:
An and-inverter graph of our own: a flat array of two-input AND nodes, hash-consed on creation, never mutated and never deleted. **Nothing calls it yet.** It lands with its tests so that the port of the folding rules — the part that can be subtly wrong — can be reviewed on its own, before anything depends on it.

## Eight bytes a node

ABC's `Aig_Obj_t` is 48 bytes on LP64. Every field it carries beyond the two fanins pays for a capability STP does not use:

| field | what it buys ABC | needed here |
|---|---|---|
| `nRefs`, and the slab allocator behind it | deletion, so rewriting can remove nodes | no — nothing is deleted |
| `Level`, `fPhase` | depth-aware rewriting in place | no — nothing rewrites in place |
| `TravId`, `fMarkA`, `fMarkB` | traversal marks | no — a side bitmap is cheaper per node and can be discarded |
| `Id` | an identifier, because identity is the pointer | no — the id *is* the array index |
| `pNext` | the structural-hash chain, stored inside the node | no — the table sits beside the nodes |

That last one is worth more than the node size. Because the chain is not woven through the nodes, `freeStrash()` can hand the table back once the graph is built — and measured on four million gates the table is 64 MB against 30 MB of nodes, so it is the larger allocation. ABC cannot do this at all.

## Two invariants the rest of the design rests on

**Fanins always have strictly smaller ids than their node,** because `And()` only ever receives literals of existing nodes and appends. Creation order is therefore topological order, which is what will let reachability be a descending sweep rather than a depth-first search: no recursion, no stack, sequential access. A million-deep chain is a test here rather than a hazard.

**Node identity is a creation counter, not an address.** Nothing can come to depend on a pointer, an allocator, or the order two objects landed in memory. `Xor` and `Mux` build their intermediates in *separate statements* rather than inside one argument list, so the order they are created in is fixed by the language rather than left unspecified — which is exactly why the ABC path needs hand-written replacements for its own `Exor` and `Mux`.

## The structural-hash table

Open-addressed with Robin Hood insertion, and **the key is not stored** — it is the node's own fanins, so a 24-bit fingerprint beside the index is enough to reject a colliding probe without touching the node array. Distance sits in the high bits of the slot so one unsigned comparison does both the Robin Hood test and the fingerprint compare.

Robin Hood earns its place because a *miss* — which is every gate that turns out to be new — otherwise probes all the way to an empty slot. Measured at load 0.7 over 4M slots: misses drop from 6.02 probes to 2.51, and the worst case from 133 to 18.

## The folding rules

The one-level folds and the two-level Brummayer–Biere rules are ported from ABC's `Aig_And`, which STP already runs with `fAddStrash` enabled — so this is parity, not a new optimisation. ABC's four recursive re-entries become assignments and a loop.

ABC's guard requiring at least one operand to be an AND node is kept, and commented, because it is not an optimisation: with two childless operands every comparison in the block is between sentinels, and without the guard the rules return an answer built from one.

## Tests

In order of what they would catch:

- **The same random gate sequence driven through this package and through ABC**, requiring every literal to denote the same function. Comparison is by truth table, never by node id — the two number nodes differently by design. 200 seeds.
- Every AND over every pair of literals drawn from four inputs and one level of gates above them, checked against the conjunction.
- Hash-consing under twenty thousand gates and repeated table growth.
- A million-deep chain, which no recursive implementation survives.
- Pre-sizing and `freeStrash()` leaving the graph identical.

**One thing the tests deliberately do not assert:** that `And(a,b)` and `And(b,a)` return the same node. They need not. The two-level rules test `p0`'s children against `p1` before the reverse, so where two rules could both fire the argument order decides which wins. ABC behaves the same way — measured at 122 differing results in 15000 ordered pairs — so this is parity, but it does mean the AIG depends on the order the blaster presents operands in, which is worth knowing. The test asserts the two agree as *functions*, and a separate test asserts the asymmetry still exists, so that losing it would register as drift rather than as an improvement.

## Measured against ABC

Four million gates, identical AND counts from both:

| | ours | ABC |
|---|---|---|
| construction | 883 ms | 1864 ms |
| peak while building | 144 MB | 282 MB |
| live at CNF time | **50 MB** | **282 MB** |

The last row is the one that matters, since that is when the CNF arena gets allocated, and it is where `freeStrash()` earns its place. ABC's 282 MB cannot be reduced — its hash chain is a field of every node.

## A negative result, recorded rather than re-derived

The second commit records an alternative that was measured and rejected: a 16-byte slot carrying both fanin literals beside the index, so that a rehash reads only the table and a probe never touches the node array. On four million gates it is a wash on time — both variants land inside a 953–1340 ms run-to-run spread over five interleaved rounds — and costs 45% more peak memory pre-sized, 75% more when the table grows by doubling.

It sounds like it should win, and does not, because the fingerprint already avoids the node read on a *failed* probe, and the read on a successful one is of a node the caller is about to use anyway. Against that, a 16-byte slot halves the slots per cache line and makes every ordinary probe worse.

### DIFF
--- a/include/stp/AIG/Literal.h
+++ b/include/stp/AIG/Literal.h
@@ -1,0 +1,74 @@
+/********************************************************************
+ * AUTHORS: Trevor Hansen
+ *
+ * BEGIN DATE: August, 2026
+ *
+Permission is hereby granted, free of charge, to any person obtaining a copy
+of this software and associated documentation files (the "Software"), to deal
+in the Software without restriction, including without limitation the rights
+to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+copies of the Software, and to permit persons to whom the Software is
+furnished to do so, subject to the following conditions:
+
+The above copyright notice and this permission notice shall be included in
+all copies or substantial portions of the Software.
+
+THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+THE SOFTWARE.
+********************************************************************/
+
+#ifndef STP_AIG_LITERAL_H
+#define STP_AIG_LITERAL_H
+
+#include <cstdint>
+
+namespace stp
+{
+namespace aig
+{
+
+// A node is an index into the manager's array; a literal is that index with
+// the inverting bit in the low position, which is AIGER's convention:
+// literal 0 is constant false, 1 is constant true, and negation is `^ 1`.
+//
+// Nothing here is a pointer. Node identity is a creation counter, so nothing
+// downstream can come to depend on an address, an allocator, or the order two
+// objects happened to land in memory -- which is the whole reason the AIG STP
+// builds today has to be built through ordered wrappers.
+using Node = uint32_t;
+using Lit = uint32_t;
+
+constexpr Lit LIT_FALSE = 0;
+constexpr Lit LIT_TRUE = 1;
+
+// Both slots of a CI or of the constant node hold this. It is deliberately
+// outside the literal space *and closed under negation there*: the two-level
+// rules below compare grandchildren without first asking whether the operand
+// has any, so a sentinel that could equal a real literal -- or whose negation
+// could -- would make a rule fire on a node that has no children.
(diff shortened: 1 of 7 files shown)


TrevorHansen:
Pushed a follow-up: `And()` is now commutative.

The PR body above says it is not, and explains why ABC's is not either — the two-level rules test `p0`'s children against `p1` before the reverse, so where two rules could both fire the argument order decides which wins. That was measured at **148 differing results in 16000 ordered pairs**, and this package reproduced it exactly, which was good evidence the rule port was faithful.

It turns out to be cheap to remove. ABC canonicalises the operands at the *hashing* step; sorting them before the *rules* run costs two lines and eliminates the asymmetry.

It is not a trade:

| | asymmetric | total AND nodes |
|---|---|---|
| ABC (`fAddStrash`) | 148 / 16000 | 14576 |
| ours, before this change | 148 / 16000 | 14576 |
| ours, after | **0 / 16000** | **14422** |

Fewer gates, because equivalent requests now take the same path through the rules and land on the same node. On a four-million-gate workload the count is identical to ABC's, and aggregated over the 200 random circuits the lockstep test drives, ours is never larger — which the test now asserts.

One test expectation changed with it. Per-circuit gate-count equality against ABC is gone: the two can now reach different-but-equivalent graphs on any single circuit. The test compares **functions** per circuit, by truth table as before, and gate counts only in aggregate. That is the honest form of the claim — equality per circuit was pinning an implementation detail rather than a property.

PR #1049, opened 2026-08-30
A third bit-blasting backend, over ABC's Gia

TrevorHansen:
`Mf_ManGenerateCnf` is the generator behind `--cnf-generation-effort low`, `high` and `very-high`, and it takes a `Gia_Man_t`. `ToCNFAIG` reaches it by building an ABC Aig and calling `Gia_ManFromAig`, so a graph at 48 bytes an object and a graph at 12 are both live when the mapper runs.

This blasts into the Gia directly. No Aig, no conversion, and the structural-hash table is handed back before the mapper allocates. Three new levels — `gia-low`, `gia-high`, `gia-very-high` — mirror the LUT sizes of `low`, `high` and `very-high`, so comparing the two backends is one word on the command line.

## The pieces

The seam is the one #1046 and #1047 established, so this is a third instantiation of it rather than new machinery:

| | in-house (#1047) | this |
|---|---|---|
| handle | `BBNodeLit` | `BBNodeGia` — a 4-byte Gia literal |
| manager | `BBNodeManagerLit` | `BBNodeManagerGia` |
| lowering | `ToCNFTseitin` | `ToCNFGia` → `Mf_ManGenerateCnf` |

A Gia handle has to be a literal rather than a pointer because `Gia_Man_t::pObjs` reallocs as it grows.

Two things are easier here than on ABC's Aig. `ciOrdinal` is O(1), since `Gia_ManAppendCi` stores the ordinal in the object — the handle carries no `symbol_index` and there is no search over the input list. And no ordered-gate wrappers are needed: `orderedAigExor` and `orderedAigMux` exist because `Aig_Exor` and `Aig_Mux2` build two nodes inside one argument list, where the evaluation order is unspecified and the ids it decides reach the CNF, whereas `Gia_ManHashXor` and `Gia_ManHashMux` assign their intermediates in separate statements.

## Two things worth reading the comments for

**The inputs are not objects 1..nCi.** On the converted path they are, because `Gia_ManFromAig` appends every CI before any AND, and `derive_cnf_mf` says so. This manager cannot: `CreateSymbol` mints an input the first time a symbol bit is touched, so inputs land interleaved with gates — on 375 of 388 fuzz corpus files, with an input moved by Mf's coarsening on 186 of 193.

That turns out not to matter, and the reason is recorded because it is not obvious. Mf coarsens the graph before mapping and derives the CNF over the copy, but `Mf_ManDeriveCnf` notices the managers differ and rebuilds `pVarNums` against the original, bridging by input and output **ordinal** rather than by id. I built a hoisting pass to restore the 1..nCi invariant before finding this, then deleted it.

**No cleanup before generation.** The Aig path runs `Aig_ManCleanup` here. A node outside the cone of the output is one Mf never maps and never gives a variable to; it costs only its share of cut enumeration — 2.1% of the AND nodes on the largest query in the hard set — and removing them means copying the graph, which costs more than the 2.1% is worth.

`ToCNFAIG::adopt()` moves to a header as a free function, unchanged, so both lowerings share the one copy.

## Measurements

Largest query in the hard set, `--exit-after-CNF`, each level against the one it mirrors:

| level | clauses | variables | peak RSS | wall |
|---|---|---|---|---|
| low | 9,414,737 | 2,121,486 | 1331 MB | 9.5 s |
| **gia-low** | 9,414,733 | 2,121,654 | **906 MB** | 9.2 s |
| high | 8,054,120 | 1,151,136 | 1711 MB | 17.9 s |
| **gia-high** | 8,054,125 | 1,151,672 | **1257 MB** | 16.0 s |
| very-high | 7,997,033 | 1,128,315 | 1777 MB | 29.3 s |
| **gia-very-high** | 7,997,021 | 1,150,007 | **1288 MB** | 27.3 s |

−32%, −27% and −28% peak. Clause counts are within 5 of each other; the graphs differ slightly because Gia's structural hash is commutative and `Aig_And` is not (stp/abc#8).

## Verification

2501 fuzz corpus queries under `--check-sanity`, so every counterexample was constructed and checked against the formula — the direct test of the input-to-variable projection, which is what could plausibly break here. `gia-low` and `gia-high` agree with `low` and `high` on all 2501.

The check was itself validated: shifting the projection by one input made it fail on 41 of 200 files, so a green sweep means something. 175/175 ctest, with the effort-level query file extended to cover the new rungs.

## One open question

On four array queries of 250 (`--array-equality` with both abstractions on), `gia-low` is 3–18x slower than `low`. Every answer is still correct and CNF generation is *faster* in each case — the whole loss is SAT search, e.g. 1311 ms → 25378 ms on one, with an extra refinement round. A slightly different formula sending the solver down a worse path. Too few cases to call it a regression, too many to call it noise; it wants a proper A/B on a larger array corpus before these levels are recommended for anything but measurement.

---

*Rebased onto #1048. Three conflicts, all from the `new_low`/`new_medium` rungs it added: the `CNFEffort` ordinals (the `gia-*` ones now follow the `new_*` ones, since the ordinals are the C interface's contract), the CLI parse chain and its level list, and the effort-level query file, which now checks all twelve. The backend dispatch merged without conflict. Every number above and the 2501-query sweep were re-taken on the merged tree; 175/175 ctest.*

### DIFF
--- a/include/stp/STPManager/UserDefinedFlags.h
+++ b/include/stp/STPManager/UserDefinedFlags.h
@@ -894,8 +894,27 @@ public:
     // little effort it spends. AUTO never picks it -- AUTO chooses from the
     // AIG's node count, which means blasting through ABC first, so reaching
     // this rung has to be an explicit request.
-    CNF_EFFORT_NEW_VERY_LOW
+    CNF_EFFORT_NEW_VERY_LOW,
+
+    // Mf_ManGenerateCnf again -- the same generator low, high and very-high
+    // reach -- but over a Gia the blaster built itself rather than one
+    // converted from an ABC Aig. Same LUT sizes, 3, 6 and 8, so gia-low
+    // against low is a comparison of the two backends and nothing else.
+    //
+    // AUTO never picks these, for the reason it never picks the rung above:
+    // it decides from ABC's Aig node count, which means blasting through ABC
+    // first, and the point of these is that no ABC Aig is built at all.
+    CNF_EFFORT_GIA_LOW,
+    CNF_EFFORT_GIA_HIGH,
+    CNF_EFFORT_GIA_VERY_HIGH
   };
+
+  // Whether a level blasts through the Gia backend rather than ABC's Aig.
+  static bool isGiaEffort(enum CNFEffort e)
+  {
+    return e == CNF_EFFORT_GIA_LOW || e == CNF_EFFORT_GIA_HIGH ||
+           e == CNF_EFFORT_GIA_VERY_HIGH;
+  }
 
   enum CNFEffort cnf_effort = CNF_EFFORT_AUTO;
 
--- a/include/stp/ToSat/AbcCnfAdopt.h
+++ b/include/stp/ToSat/AbcCnfAdopt.h
@@ -1,0 +1,69 @@
+/********************************************************************
+ * AUTHORS: Trevor Hansen
+ *
+ * BEGIN DATE: August, 2026
+ *
+Permission is hereby granted, free of charge, to any person obtaining a copy
+of this software and associated documentation files (the "Software"), to deal
+in the Software without restriction, including without limitation the rights
+to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+copies of the Software, and to permit persons to whom the Software is
+furnished to do so, subject to the following conditions:
+
+The above copyright notice and this permission notice shall be included in
+all copies or substantial portions of the Software.
+
+THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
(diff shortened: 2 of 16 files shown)


aytey:
@TrevorHansen 

> On four array queries of 250 (--array-equality with both abstractions on), gia-low is 3–18x slower than low. Every answer is still correct and CNF generation is faster in each case — the whole loss is SAT search, e.g. 1311 ms → 25378 ms on one, with an extra refinement round. A slightly different formula sending the solver down a worse path. Too few cases to call it a regression, too many to call it noise; it wants a proper A/B on a larger array corpus before these levels are recommended for anything but measurement.

Did we capture any of these/do we know if they’re reproducible?

TrevorHansen:
> @TrevorHansen
> 
> > On four array queries of 250 (--array-equality with both abstractions on), gia-low is 3–18x slower than low. Every answer is still correct and CNF generation is faster in each case — the whole loss is SAT search, e.g. 1311 ms → 25378 ms on one, with an extra refinement round. A slightly different formula sending the solver down a worse path. Too few cases to call it a regression, too many to call it noise; it wants a proper A/B on a larger array corpus before these levels are recommended for anything but measurement.
> 
> Did we capture any of these/do we know if they’re reproducible?

I didn't save them.  I'll do comprehensive benchmarking later on. Problems should be clearer then.

PR #1062, opened 2026-09-01
Don't crash when an unusable CryptoMiniSat defines the target

TrevorHansen:
An installed CryptoMiniSat can be broken in a way its own CMake package half-survives: when the cadical/cadiback it was built against are uninstalled, its Targets file still runs far enough to define the cryptominisat5 imported target before noticing the missing targets and reporting NOT FOUND. cmake/FindCryptoMiniSat.cmake then falls to its build-here rung, whose add_library() aborts the configure with:

```
CMake Error at cmake/FindCryptoMiniSat.cmake:200 (add_library):
  add_library cannot create imported target "cryptominisat5" because another
  target with the same name already exists.
```

— an error that names neither the package at fault nor what is wrong with it. A second path reaches the same abort: a package that works but is rejected by the version check (older than 5.11) has defined the target too, so the fallback dies identically.

Imported targets cannot be redefined, so when the name is taken nothing can be built here; the fix is to say so instead of colliding. The fallback rung now checks for an existing cryptominisat5 target first:

- Under `-DUSE_CRYPTOMINISAT=ON` the configure stops with an error naming the package directory, the package's own reason when it gave one (or its version when that was the objection), and the ways out: repair or remove the installation, point `-Dcryptominisat5_DIR` at a usable copy, or configure with `-DUSE_CRYPTOMINISAT=OFF`.

```
CMake Error at cmake/FindCryptoMiniSat.cmake:160 (message):
  CryptoMiniSat was asked for, but the cryptominisat5 imported target is
  already defined, by the unusable package in
  '/usr/local/lib/cmake/cryptominisat5', so a CryptoMiniSat cannot be built
  here.  Its config file reports: The following imported targets are
  referenced, but are missing: cadiback cadical.  Repair or remove that
  installation, point -Dcryptominisat5_DIR at a usable copy, or configure
  with -DUSE_CRYPTOMINISAT=OFF.
```

- Under AUTO the build continues without CryptoMiniSat, as it does when none is installed, with a STATUS line carrying the same information.

Verified against a machine with exactly this broken package: AUTO now configures and generates cleanly with CryptoMiniSat in the disabled-features list, ON fails with the message above, a usable package is still found and used unchanged, and a synthetic 5.8 package reproduces the version-path crash on master and the clean skip with this change.

### DIFF
--- a/cmake/FindCryptoMiniSat.cmake
+++ b/cmake/FindCryptoMiniSat.cmake
@@ -131,6 +131,46 @@ if(NOT CryptoMiniSat_FOUND_SYSTEM)
     # check_auto_download() is fatal by construction, and is reached only when
     # this was asked for by name. Otherwise the answer is the one AUTO asks
     # for: say nothing and build without it.
+    # A package consulted above can define the cryptominisat5 imported target
+    # and still be unusable: a config whose Targets file references imported
+    # targets that no longer exist -- a system CryptoMiniSat built against a
+    # since-removed CaDiCaL -- defines the target and then reports NOT FOUND,
+    # and a package rejected for its version defined it too. An imported
+    # target cannot be redefined, so no copy can be built here under that
+    # name; without this check the add_library() below aborts naming the
+    # collision but not the package that caused it.
+    if(TARGET cryptominisat5)
+        if(cryptominisat5_DIR)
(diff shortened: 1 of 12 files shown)


TrevorHansen:
Obscure fix.

PR #1072, opened 2026-09-01
Keep comparator cells propagation-complete when their equality is shared

TrevorHansen:
Second of the stack, on top of #1071; the diff shows only this step.

Sharing used to break the matching: a cell whose equality something else also reads (an eq over the same operands, two comparisons sharing xnors) fell back to the per-gate encoding, whose windows join on two variables and lose unit refutation. Now the match never declines. A private equality dies into the majority block as before; a shared one keeps its variable and its own four clauses for the other readers, and the cell emits the minimum linking set that makes the window over (operand, operand, equality, arm, out) propagation-complete with the equality inside it: ten clauses for the full cell (either arm orientation), five for the bottom conjunction.

The minimums are exact (computed over the joint relations' prime implicates), and chains of linked cells verify propagation-complete exhaustively at widths 3–4 over every variable with every equality shared. The obvious repair — emitting the majority block beside the live equality — is no better than the fallback (both still miss conflicts and ~6% of implied literals in that sweep): a block that never reads the live equality cannot couple to it, which is what the link clauses are for. Both joint matchers verify their normalization by evaluating the cone over every operand assignment, so a polarity surprise declines to the default encoding instead of miscoding.

Cost is +4 clauses per shared full cell and +2 per shared bottom against the old fallback, and nothing anywhere else: on a 40-file corpus the totals move by two hundredths of a percent, and the 208-file differential stays answer-identical. Because the link blocks can pass plain Tseitin's clause count from the third cell sharing one equality, the test that asserted matching never costs a clause now asserts the bounds that still hold: never a variable, and aggregate savings.

Unit tests cover both link templates, the shared bottom, and the agreeing-arm orientation, each exact over every input.

### DIFF
--- a/include/stp/AIG/Tseitin.h
+++ b/include/stp/AIG/Tseitin.h
@@ -70,6 +70,17 @@ bool matchMajority(const Manager& m, Node n, Lit& x, Lit& y, Lit& z);
 // over the operands: n = g & h. Structural only, as matchMajority is.
 bool matchXorAnd(const Manager& m, Node n, Lit& g, Lit& h);
 
+// The same two cells for a condition that stays live because something else
+// reads it. matchJointCell normalizes an ITE cell to the study templates
+// over (a, b, e, other, out) with e == (a == b): tSide false means
+// out = e ? other : b (the borrow-chain cell), true means out = e ? a :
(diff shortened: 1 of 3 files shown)


TrevorHansen:
I need to evaluate more carefully whether this is common enough to justify the extra complexity.

PR #1073, opened 2026-09-01
Complete the recovered ITE with its two condition-free prime implicates

TrevorHansen:
Last of the stack, on top of #1072; the diff shows only this step. This is the one with a real trade in it, split out so it can be judged on wall-clock evidence separately from the strict wins beneath it.

The recovered ITE emitted only the four clauses that mention the condition, so agreeing arms could not decide the output while the condition was unset: ite graded 94.7–97.6% of implied literals derived at widths 1–4, and every multiplexer and barrel shifter inherited the gap. With the two condition-free prime implicates added the block is the relation's full prime set: ite grades unit-refutation and arc-consistency clean at widths 1–4 and propagation-complete exhaustively at widths 1–2, the shifters become refutation-clean, and probe-style workloads on mux cones solve by pure propagation. An exclusive-or's arms share a node, making both extra clauses tautologies, so xor/iff/bvxor/bvadd/eq are byte-identical.

The price is clause growth where multiplexers dominate: measured alone on a 40-file corpus it added 10.0% clauses, up to 40% on register-mux unrollings and nothing where no recovered ITEs occur; combined with the comparator recoveries beneath it the stack lands at +6.35% clauses and −4.09% variables overall. Denser clause sets tax search-heavy proofs even as they win propagation-bound ones, so whether this pays overall wants a solve-time round on the hard set before it becomes the default; the encoding-strength case for it is complete either way.

Verification: the mux unit test pins the six-clause block, the 208-file differential at new-medium is answer-identical, and the graded sweep is unchanged outside ITE cones.

### DIFF
--- a/include/stp/AIG/Tseitin.h
+++ b/include/stp/AIG/Tseitin.h
@@ -132,8 +132,10 @@ public:
   // In the cone, so it gets a variable and its defining clauses.
   bool live(Node n) const { return (live_[n >> 6] >> (n & 63)) & 1u; }
 
-  // Encoded as one four-clause ITE over its grandchildren rather than as
-  // three ANDs. Its two fanin nodes are then not live at all.
+  // Encoded as one ITE block over its grandchildren rather than as three
+  // ANDs: the six prime implicates of the relation (four when the arms
+  // share a node and it is an exclusive-or). Its two fanin nodes are then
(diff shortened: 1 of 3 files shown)


TrevorHansen:
This need to be measured carefully.

PR #1075, opened 2026-09-01
Make new-medium the default CNF conversion

TrevorHansen:
The default `--cnf-generation-effort` was `auto`, which resolved to `gia-low` or `new-medium` from the recorded blast estimate and fell back to the older `very-low`/`medium` choice under array refinement, on SAT backends other than CaDiCaL, or with no estimate recorded. The default is now `new-medium` everywhere — the in-house Tseitin writer over the in-house AIG, with XOR, if-then-else and n-ary AND/OR recovery — for the command line and for embedders alike (both the `UserDefinedFlags` initializer and the CLI's mapped default string move, since the CLI always writes its string over the flags).

`auto` stays selectable and resolves exactly as before. The C interface documentation now names ordinal 8 as the default, and the paragraph claiming auto never selects the backend rungs — stale since auto started resolving to `gia-low`/`new-medium` — is corrected in passing.

Tests: the three query-file tests that exercised auto's resolution relied on it being the default; they now pass `--cnf-generation-effort auto` explicitly, the effort-menu test checks that a bare run prints the `new-medium` line, and the C API default test expects `CNF_EFFORT_NEW_MEDIUM`. Full ctest suite passes (179/179).

### DIFF
--- a/docs/bv-abstraction.rst
+++ b/docs/bv-abstraction.rst
@@ -242,7 +242,7 @@ Which CNF generator
 The abstraction's search is many-solve: every refinement round is another
 call on a solver that keeps the whole CNF, and which CNF it keeps decides
 how that search goes far more than it decides one solve. With
-``--cnf-generation-effort`` at its default ``auto``, turning
+``--cnf-generation-effort`` at ``auto``, turning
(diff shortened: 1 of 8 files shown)


TrevorHansen:
Premature. The new AIG/CNF isn't as nice as I'd like yet.

PR #1077, opened 2026-09-02
Add propagator_bench division-variant flags and a UP-vs-cbitp duel mode

TrevorHansen:
## What

Three instrumentation additions used by the 2026-09-03 division-encoding study, mirroring the multiplication study's `--bb.mult-v` precedent.

- `--bb.div-v1` .. `--bb.div-v4` on propagator_bench set `UserDefinedFlags::division_variant_1..4`, so the graded `--consistency` checks cover every divider circuit variant.
- `--duel W` runs an exhaustive per-state head-to-head at width W: for every ternary partial assignment of the operation's input/output bits it compares what unit propagation on the bit-blasted CNF derives against what the constant-bit transfer function derives (iterated to a fixed point), refereed by the exact solution table. Tallies are split on whether the divisor-position operand can still be zero, which is exactly where the division transfers bail out, and `--duel-dump FILE` writes the asymmetric cases for offline classification. Both engines came out sound on every arm measured; the mode exists to locate complementary strengths, e.g. the bvurem transfer refutes 45,060 w=4 conflicts the gia-low CNF misses, while the CNF refutes 44,264 divisor-can-be-zero conflicts the transfer abstains on.
- `STP_DIVMAP=<path>` makes ToSATAIG write each term-abstraction record's kind, width, and operand/result DIMACS variables (numbered as `--output-CNF` numbers them) next to the CNF, for external-propagator experiments that need to know which SAT variable is which divider bit. Off unless the environment variable is set.

## Why

The duel answers a question the existing precision and consistency phases cannot: not "how strong is each side against the ideal" but "which states does one engine decide that the other cannot", which is the sizing input for both eager repair blocks and lazy propagation oracles.


TrevorHansen:
Second commit adds the payoff the duel priced: `--bb.div-lemmas` (off by default) asserts each divider's order laws — b≠0 ⇒ r<b, r≤a, b≠0 ⇒ q≤a, the b=0 clamps and the b=1 identities — as side constraints next to the blast. Exhaustively graded at w=2..4: bvurem goes from 1,742 missed conflicts to 29 at w=3 and from 72,376 to 4,839 at w=4, for +2.7–7.4% clauses at 64 bits; no unsoundness in the sweeps and a 400-file differential over the fuzz corpus' division files agrees with the flag off on every answer. Three lit tests included. The propagator_bench flag mirrors it so both states stay measurable.

TrevorHansen:
Third commit closes the other hole the duel priced: the unsigned division transfers now case-split on a zero divisor instead of bailing out (the old TODO). The zero branch is exact (b=0, q all ones, r=a), the nonzero branch reruns the existing engines seeded past zero, and a bit propagates when both branches force it — a dead branch decides the divisor. Exhaustively at w=3/4: bvudiv misses 26 instead of 2,081 conflicts (87% instead of 52% of deducible bits), bvurem 210 instead of 1,446 (83% vs 61%); zero unsound derivations, corpus answers unchanged, and a 400-file differential against --disable-cbitp agrees everywhere. With this the constant-bit propagator out-derives every CNF rung on both operations across the whole state space, not just the nonzero-divisor region.

TrevorHansen:
The division-transfer case split has moved to its own pull request, #1079, cherry-picked unchanged (including the empty-interval CONFLICT follow-up). It is reverted here rather than rewritten out, so this branch now carries only the duel/variant instrumentation and the off-by-default `--bb.div-lemmas` flag; the net diff no longer touches the constant-bit transfers. Review of the pipeline-affecting change can proceed on #1079 independently.

PR #1080, opened 2026-09-03
Settle the zero-divisor case splits in one call, aliased operands included

TrevorHansen:
Stacked on #1079, fixing the failure its exhaustive aliased tests hit once the assertion repair lets them run to completion.

Two repairs to the division and remainder case splits. Copies of aliased operands now stay aliased -- one variable read twice becomes one copy -- so a branch can no longer fix two views of the same variable apart and hand the join an inconsistent state. And each split now reruns its branch and join until nothing changes: a round's fixings can kill the zero branch or sharpen the next round's copies, and a transfer call is required to reach its own fixed point in one invocation, which the aliased exhaustive tests check.

With this, the full ConstantBitP_TransferFunctions suite passes in a Debug+assertions build (aliasedDivisionsWidth3 included), and the exhaustive width-4 precision sweeps and the UP-vs-cbitp duel reproduce bit-identically, so the measured transfer numbers are unchanged.

### DIFF
--- a/lib/Simplifier/constantBitP/ConstantBitP_Division.cpp
+++ b/lib/Simplifier/constantBitP/ConstantBitP_Division.cpp
@@ -771,67 +771,80 @@ Result bvUnsignedModulusBothWays(vector<FixedBits*>& children,
     FixedBits& b = *children[1];
     const unsigned width = output.getWidth();
 
-    bool zeroBranchOk = true;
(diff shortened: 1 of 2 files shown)


TrevorHansen:
Superseded: both repairs were folded into the base branch in its own spelling (the settle iteration plus conflict-detecting adoption for aliased operands, and the cbitp pin in the paired-relation test), and its CI is green there.

PR #1088, opened 2026-09-08
Keep the SAT trail between UF congruence refinement rounds

TrevorHansen:
A UF congruence refinement round adds the lemmas the last candidate earned and asks the solver again. CaDiCaL backtracks to the root whenever a clause arrives after a solve, so every round unassigned the whole trail and re-decided every variable before it could see the new clauses. On the largest 20241113-Certora QF_UFBV file (2246.smt2: one predicate with 2,651 applications over 55 distinct argument terms, a 2.3M-clause encoding) that was 66k decisions and 437k propagations on a round that produced no conflict at all, 915 rounds, 287 s. The 349 ms per round was entirely this re-descent plus the repair of the eight new lemmas; STP's own checker and counterexample construction were 11 ms of it.

The batch driver now asks the backend for incremental lazy backtracking in its clause-scope mode (CaDiCaL `ilb=2`) when the query has an active UF view. `SATSolver::enableTrailReuse` gained a scope for this: the incremental driver keeps the assumption-prefix mode it measured, the refinement loop takes the mode that also keeps the trail across clauses added between calls, backtracking only as far as an added clause is falsified. `--uf-trail-reuse` (default on) and the C API flag `UF_TRAIL_REUSE` turn it off; other backends decline the hint as before. A unit test exercises the refinement-shaped loop (plain solves, model-blocking clauses over fresh variables added between them) under the new mode.

Measured over the local QF_UFBV corpus at a 60 s timeout, arms interleaved per file in one parallel run. Certora 2023/2024, calc2, btfnt (1,136 files): 566 solved against 538, +46/−18, geomean time ratio 0.835 on the 497 files over a second, 190 files faster than 0.8x and 38 slower than 1.25x. Goel-hwbench, Bouvier, Wolf (1,403 files): 1,361 solved either way, geomean 0.955, no file lost. No verdict disagreed in either sweep. On 2246.smt2 alone, unloaded: 915 rounds and 287 s to 338 rounds and 73 s at the default lemma cap; the round count falls as well as the per-round cost, because a candidate that moved only where the lemmas forced it exposes fewer new collisions than one rebuilt from saved phases through a restarted search.

Two things were tried on the same file and not shipped. Uncapping `--uf-lemmas-per-round` (already available) is the fastest single setting for this file at 59 s but a wash on the family (+17/−19 solved). A cap that doubled after each refuted candidate was within noise of the fixed cap in opposite directions on the two sweeps and no better here once the trail is kept. Disabling CaDiCaL's lucky phases, the obvious suspect for a per-call fixed cost, changed nothing. Array read refinement and bit-vector abstraction refinement add clauses between solves the same way and pay the same re-descent; the change is gated to UF until they are measured.

### DIFF
--- a/include/stp/STPManager/UserDefinedFlags.h
+++ b/include/stp/STPManager/UserDefinedFlags.h
@@ -414,6 +414,15 @@ public:
   // apart can only change how quickly an answer is found, never which answer.
   bool uf_phase_hints = false;
 
+  // Whether the batch solver keeps its trail between congruence refinement
+  // rounds. A round differs from the last only by the clauses the refuted
+  // candidate earned, so the search can resume on the kept trail,
+  // backtracking only as far as one of those clauses is falsified. Without
+  // this every round re-decides and re-propagates every variable from the
+  // root before it reaches the new clauses, a cost that grows with the
+  // encoding rather than with the lemma.
+  bool uf_trail_reuse = true;
+
   // The carrier width given to a sort introduced by (declare-sort S 0).
   //
   // An uninterpreted sort has no operations but equality, so a query
--- a/include/stp/Sat/Cadical.h
+++ b/include/stp/Sat/Cadical.h
@@ -124,7 +124,7 @@ public:
   bool disableEliminationAndShrinkingInternal() override;
   bool disableLuckyPhasesInternal() override;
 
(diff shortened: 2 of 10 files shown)


TrevorHansen:
Superceeded.

PR #1091, opened 2026-09-09
Correct roundToIntegral and the FMA invariant in the vendored SymFPU, and round an integer past the format to an infinity

aytey:
Three floating-point defects found while exercising the tiny formats that the test suite and the fuzzers use, where the exponent field is two bits wide or the exponent range is narrower than the significand. None is reachable in an IEEE format, but all three are reachable in formats the input language accepts, and two of them answer a query wrongly rather than failing loudly.

**SymFPU roundToIntegral in two-bit-exponent formats.** The rounding collar widens the significand by one bit and shifts by the exponent; with `eb = 2` and a power-of-two significand the width computed for that shift wrapped, and `1.0` rounded to something that is not `1.0`. Both back-ends took the same path, so the constant folder agreed with the symbolic encoding and a true equation came back `unsat`. Three query files cover the literal fold and the symbolic encoding at `(2, 4)` and `(2, 8)`.

**The SymFPU FMA invariant that aborted the literal folder.** An FMA whose product overflows the format tripped a postcondition in the vendored copy rather than returning the value the standard asks for. Three query files: the overflowing product, the overflowing sum, and the same over symbols so that the symbolic path is covered beside the fold.

**Rounding an integer beyond the format.** Where `2^(eb-1) < sb` the top binade holds non-integers, and rounding one up is an integer the format has no finite for: `3.5` in `(2, 3)` is `4` under RNE, past the largest finite `3`. SymFPU left an unpacked float one exponent past the largest normal, which is no valid float — the symbolic encoding packed it, by the accident of its bits, to an infinity that its own flags did not consider infinite, so `fp.isInfinite` of it was false while equality with `+oo` was true. The node factory made it worse by folding `fp.isInfinite` over a rounding to the argument's answer outright. The value is now the infinity of the argument's sign, which the postcondition accepts and both back-ends agree on, and the folding rule is kept only for the formats where a rounding never overflows, which is `2^(eb-1) >= sb`. Three query files: the symbolic query, the same toward zero where no overflow is possible, and the folds that aborted.

Please merge as separate commits rather than squashing: each carries its own evidence and trailers.

### DIFF
--- a/cmake/FindSymFPU.cmake
+++ b/cmake/FindSymFPU.cmake
@@ -58,13 +58,13 @@ if(NOT SymFPU_FOUND_SYSTEM)
     endif()
 
     # stp/symfpu is a fork laid out like the ABC one: main tracks upstream and
-    # the `stp` branch, which this pins a commit of, carries STP's four
+    # the `stp` branch, which this pins a commit of, carries STP's seven
     # correctness fixes as commits. They used to be patch files applied at
     # configure time; as commits each one names what it fixes, and a bump is a
     # rebase in that repository where a conflict says which change upstream has
     # met.
-    set(SymFPU_COMMIT "d358a6defeace0cd44695e7d922fc62c2f8b8ee8")
-    set(SymFPU_CHECKSUM "2557cc598ebde7e6d673cbb7f48a6f0928a18e38bbd3a5b00d126c0f45ba79f1")
+    set(SymFPU_COMMIT "e457a07735f0d1f56cb543a3b3ff8f28551c02db")
+    set(SymFPU_CHECKSUM "f147b4b16cdab398bd3588d75e0fe524f7af0cf675997a62e7e1aff65e868ee4")
 
     ExternalProject_Add(
         SymFPU-EP
--- a/lib/NodeFactory/SimplifyingNodeFactory.cpp
+++ b/lib/NodeFactory/SimplifyingNodeFactory.cpp
@@ -273,6 +273,18 @@ static bool fpIsRoundToIntegral(const stp::ASTNode& n)
   return n.GetKind() == stp::FP_ROUNDTOINTEGRAL && n.Degree() == 2;
 }
 
+// Whether every value of the format's top binade is an integer, which is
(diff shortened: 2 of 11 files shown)


aytey:
Restructuring this: the two SymFPU fixes belong in stp/symfpu rather than as whole-file overlays here, and they are now stp/symfpu#2 (three commits, the resulting `core/convert.h` and `core/fma.h` byte-identical to the copies this PR carries, so nothing measured changes).

Once that merges I will rewrite this branch to two commits: one that moves `SymFPU_COMMIT` and its checksum past the fixes and deletes `cmake/deps-utils/symfpu/core/{convert,fma}.h` along with the `PATCH_COMMAND` that overlays them, keeping the six query files that cover the two defects through STP; and one for the part that is genuinely STP's, the `fp.isInfinite` over `fp.roundToIntegral` folding in `SimplifyingNodeFactory` and its three query files.

Marking it a draft until then, since merging it as it stands would take the overlay.

aytey:
stp/symfpu#2 is merged, so this is now what it should have been: a pin move and the one change that is genuinely STP's.

**Move the SymFPU pin past the roundToIntegral and FMA fixes.** `SymFPU_COMMIT` goes to e457a07 on the `stp` branch with its checksum, and the comment above it counts seven fixes rather than four. The whole-file overlay under `cmake/deps-utils/symfpu/` and the `PATCH_COMMAND` that applied it are gone; the headers the pin now fetches are byte-identical to what the overlay carried, so nothing measured changes. Six query files come with it, three for each of the two defects observable from here on their own.

**Fold `fp.isInfinite` over a rounding only where a rounding cannot overflow.** The part that is STP's: with the pin moved, SymFPU rounds an integer beyond the format to an infinity, and the node factory's rewrite that answered `fp.isInfinite` over an `fp.roundToIntegral` as false outright is right only where every value of the top binade is an integer, `2^(eb-1) >= sb`. Three query files.

Built against the merged pin with a fresh dependency tree, so the new tarball is actually fetched rather than a cached one reused: 183/183.

PR #1095, opened 2026-09-09
Drop references to files that exist only where they were written

TrevorHansen:
Seven comments pointed a reader at something outside the repository:

- `.github/workflows/ci.yml` and `lib/Simplifier/DifficultyScore.cpp` each named a dated report under a private benchmark tree.
- `include/stp/AIG/Tseitin.h` and `tests/unit-tests/RemoveUnconstrained_Exhaustive_Test.cpp` named that tree without naming a file in it.
- `tests/api/C/getbv.cpp`, `tests/api/C/simplify.cpp` and `tests/api/C/stp-counterex.cpp` each carried a `g++` line against `/home/vganesh/stp/...`, written when the C interface lived in `c_interface/` and the library in `lib/`. Neither directory exists, the files are gtest cases built by CMake, and the command has not described how to build them for a long time.

None of them can be followed by anyone reading the repository, and the two that mattered were standing in for a finding rather than adding to one.

Where the reference stood in for a finding, the finding is now stated. Where a tool in the repository regenerates the numbers, that tool is named: `DifficultyScore.cpp` now points at `tools/difficulty_bench`, which runs the sweep the constants were fitted to and prints the estimate against the measured AND-node count. The stale `g++` lines are deleted rather than corrected.

Comments only -- no code changes, and nothing generated differs.

### DIFF
--- a/tests/api/C/stp-counterex.cpp
+++ b/tests/api/C/stp-counterex.cpp
@@ -22,9 +22,6 @@ OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
 THE SOFTWARE.
 **********************/
 
-// g++ <this-filename> -I/home/vganesh/stp/c_interface
-// -L/home/vganesh/stp/lib -lstp -o cc
-
 #include "stp/c_interface.h"
 #include <gtest/gtest.h>
 #include <stdio.h>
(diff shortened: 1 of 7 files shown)


tests/api/C/stp-counterex.cpp
```diff
@@ -22,9 +22,6 @@ OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
 THE SOFTWARE.
 **********************/
```
ravisingh810392gmailcom-lang: https://github.com/stp/stp/pull/1095/changes/f7e4eb3d0e4e5c636c3942abbe7fd8534e01ce08#diff-2dcfebb9a6bc6ab9a7656fea7639c84c85914fada18aada6d8891b3fd293c583L24

PR #1097, opened 2026-09-11
REVERTED: Manage native Python handles through solver lifetime

aytey:
Track native expression and type handles in their owning solver and release them before destroying the checker. Context exit and explicit close retire the solver deterministically; retained expressions reject access after close. Release temporary floating-point carriers after model reads.

Validation: Focused checks passed for explicit close, context exit, retained expressions, owner collection, repeated close, arrays, function applications, and repeated 128-bit floating-point model reads.


aytey:

The change itself was not rejected. It is reopened for review as #1104, where the tree is byte-identical to what was merged here. The commit was rebuilt rather than reused, because the original was unsigned, carried retired `@vector.com` identities as both author and committer, and had no attribution trailer.

Nothing is lost: the original commit stays permanently reachable at `refs/pull/1097/head`.

PR #1100, opened 2026-09-11
REVERTED: Document ownership and lifetime of raw C API handles

aytey:
State when validity-checker children become invalid, who may delete them, and which operations require the owner to remain live. Clarify the same contract beside counterexample and destruction entry points.

Validation: Documentation-only change; verified the documented entry points against the existing C API implementation.

### DIFF
--- a/include/stp/c_interface.h
+++ b/include/stp/c_interface.h
@@ -65,10 +65,9 @@ extern "C" {
 // fail at link time. A shared build gets dllexport while the library is being
 // compiled and dllimport for everyone else.
 //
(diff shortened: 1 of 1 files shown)


aytey:

The change itself was not rejected. It is reopened for review as #1107, where the tree is byte-identical to what was merged here. The commit was rebuilt rather than reused, because the original was unsigned, carried retired `@vector.com` identities as both author and committer, and had no attribution trailer.

Nothing is lost: the original commit stays permanently reachable at `refs/pull/1100/head`.

PR #1102, opened 2026-09-11
REVERTED: Copy deep AIGs into GIA without recursive stack growth

aytey:
Replace recursive AIG-to-GIA traversal with an explicit work stack so deep circuits do not exhaust the process stack. Apply the patch to the pinned ABC source during automatic dependency builds.

Validation: Built the patched pinned ABC dependency from source. A 200,000-level AIG converted successfully with a 512 KiB stack limit; the same fixture using the original recursive implementation failed with SIGSEGV.


aytey:

The change itself was not rejected. It is reopened for review as #1109, where the tree is byte-identical to what was merged here. The commit was rebuilt rather than reused, because the original was unsigned, carried retired `@vector.com` identities as both author and committer, and had no attribution trailer.

Nothing is lost: the original commit stays permanently reachable at `refs/pull/1102/head`.

PR #1103, opened 2026-09-11
REVERTED: Use the last value of repeated Boolean command-line options

aytey:
Allow repeated value-taking Boolean options and use the last occurrence. Appending an override now selects that value instead of failing command-line parsing.

Validation: Focused command-line checks passed with repeated --flattening values in both orders (0 then 1, and 1 then 0).


aytey:

The change itself was not rejected. It is reopened for review as #1110, where the tree is byte-identical to what was merged here. The commit was rebuilt rather than reused, because the original was unsigned, carried retired `@vector.com` identities as both author and committer, and had no attribution trailer.

Nothing is lost: the original commit stays permanently reachable at `refs/pull/1103/head`.

